PR #3, opened 2015-02-25
pmlogger: Ensure that pmlogger starts after pmcd.

mvollmer:
When pmlogger and pmcd are started at the same time by systemd (as
happens during system boot), we need to make sure that pmlogger is not
activated before pmcd is ready to accept connections from clients.


kmcdonell:
We have a similar issue for the pmie service script.  I'll fix that through the regular PCP source tree and it will flow to github in due course.

PR #31, opened 2015-07-16
[python-pcp] Fix a bug in pmsubsys which causes timestamp to reset.

marcliangma8:
Problem:
pmFreeResult is used to free memory allocated by metric_result. However, it will corrupt memory and reset self._timestamp for diff calculation. So, all subsequent collections will have huge aggregated values instead of rate between intervals.

Fix: 
Use copy.deepcopy to assign values for _timestamp which will survive from pmFreeResult call.

### DIFF
--- a/src/python/pcp/pmsubsys.py
+++ b/src/python/pcp/pmsubsys.py
@@ -173,7 +173,7 @@ class Subsystem(object):
 
         try:
             metric_result = pcp.pmFetch(self.metric_pmids)
-            self._timestamp = metric_result.contents.timestamp
+            self._timestamp = copy.deepcopy(metric_result.contents.timestamp)
         except pmErr as e:
             self._timestamp = timeval(0, 0)
             raise e


pcpemail:
Hi Marc,

marcliangma8 <notifications@github.com> writes:
> Problem:
> pmFreeResult is used to free memory allocated by metric_result.
> However, it will corrupt memory and reset self._timestamp for diff
> calculation. So, all subsequent collections will have huge aggregated
> values instead of rate between intervals.
>
> Fix: 
> Use copy.deepcopy to assign values for _timestamp which will survive
> from pmFreeResult call. 

Thanks for posting this.  The patch looks sound to me, and fixes the issue
on the boxes I've tested it on.  I'll try and add some qa later this
week.

Cheers,

Lukas

natoscott:
This fix was merged awhile back, not sure why its still showing as open.

PR #33, opened 2015-08-04
missing global_config

prasincs:
Seems like a typo to me


ryandoyle:
Yep seems to be a typo, should be on the `global_config` scope.

PR #34, opened 2015-08-07
Fixes SIGSEGV, 64-bit format specifiers, dlsym bug and IPC on Windows

dgiagio:
This patch makes it actually possible to run PCP on Windows platform by cross-compiling from Fedora with mingw64.

1 - The format specifier for printing 64-bit signed and unsigned numbers is incorrect for Windows platform. It should be %I64d and %I64u as opposed to %lld and %llu.

2 - The MinGW on Windows already correctly defines the macros S_IRGRP, S_IWGRP, S_IROTH, S_IWOTH, S_IRWXG, and S_IRWXO but the current code incorrectly redefines all of them to 0 (zero).

3 - The current code tries to use Unix Domain Sockets for IPC communication with plugins. This is not supported on Windows platform.

4 - Fixed a SIGSEGV which happened when an error returned from setting up signal handlers. Instead of comparing the return to NULL it compared to < 0, which is a classic arithmetic to pointer comparison error.

5 - Fixed a bug of error handling when dynamic loading a shared library (DLL in case of Windows).

### DIFF
--- a/configure
+++ b/configure
@@ -11317,13 +11317,21 @@ $2 == "bozo"	{ t = $1
 	_fmt='"ld"'
 	;;
     'long long'|'long long int')
-	_fmt='"lld"'
+        if test $target_os = mingw; then
+            _fmt='"I64d"'
+        else
+	    _fmt='"lld"'
+        fi
 	;;
     'unsigned long'|'unsigned long int'|'long unsigned int')
 	_fmt='"lu"'
 	;;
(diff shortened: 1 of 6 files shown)


lberk:
Seems to fix the build warnings mentioned on my fedora-mingw64 buildbot, as well as passes a quick sanity testsuite run.

lberk:
Merged upstream

PR #50, opened 2015-11-09
libpcp: unlock context before returning

ryandoyle:
If called in a non-archive context, the context ctxp->lock c_lock is
never unlocked.

### DIFF
--- a/src/libpcp/src/logutil.c
+++ b/src/libpcp/src/logutil.c
@@ -2268,8 +2268,12 @@ pmGetArchiveLabel(pmLogLabel *lp)
 {
     __pmContext		*ctxp;
     ctxp = __pmHandleToPtr(pmWhichContext());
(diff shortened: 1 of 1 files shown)


kmcdonell:
Looks good Ryan, thanks.  I've pulled your commit into my tree and it will flow back to github via pcp.io in due course.

This is a pretty unlikely code path (calling pmGetArchiveLabel() when the current context is valid, but not an archive) and I would expect none of our code can actually be persuaded to do this, so I am curious if you actually observed the problem that this patch fixes, or noticed by code inspection, or have some devious test harness and lock audit tool that exposed the logic error.

Anyway, thanks for improving the correctness and robustness of the code.

ryandoyle:
Hi Ken

Its true that its an unlikely code path. I discovered it while in the process of writing Ruby bindings (still a work in progress!). I have test cases for local and archive contexts which test both success and error conditions.

If you're interested, I have a WIP branch called `ruby-bindings` at https://github.com/ryandoyle/pcp

PR #66, opened 2016-02-09
pcp2graphite improvement: send all metrics at once in pickled mode

test-account-0:
Before this change "pickled_input" variable was appended and metrics send over and over again.

Time to send metrics to graphite dropped for me from 6 minutes to 2 seconds.

### DIFF
--- a/src/pcp2graphite/pcp2graphite.py
+++ b/src/pcp2graphite/pcp2graphite.py
@@ -211,12 +211,15 @@ network.""")
                 self.socket = socket.create_connection((self.graphite_host,
                                                         self.graphite_port))
             if self.pickle:
-                import cPickle
+                try:
+                    import cPickle as pickle
+                except:
+                    import pickle
                 import struct
                 pickled_input = []
                 for (metric, value) in miv_tuples:
                     pickled_input.append((metric, (timestamp, value)))
-                pickled_output = cPickle.dumps(pickled_input, protocol=self.pickle_protocol)
+                pickled_output = pickle.dumps(pickled_input, protocol=self.pickle_protocol)
                 header = struct.pack("!L", len(pickled_output))
                 msg = header + pickled_output
                 if self.context.pmDebug(c_api.PM_DEBUG_APPL0):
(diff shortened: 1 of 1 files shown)


test-account-0:
About cPickle. I think it was supported around Python 2.3 - https://www.python.org/dev/peps/pep-0307/

But in Python 3 - "The pickle module has an transparent optimizer (_pickle) written in C. It is used whenever available. Otherwise the pure Python implementation is used."

test-account-0:
I've added workaround for Python 3.

test-account-0:
Also the number of the metrics (self.metrics) printed at te beginning of execution is not really true. I think it should be len(miv_tuples) in the function execute.

fche:
Yeah, but miv_tuples is a changing value within the main loop; self.metrics is a one-time value suitable for printing at program startup.

dbrolley:
Would it be possible to create a qa test for this?

natoscott:
@dbrolley qa/667 gives pretty good regression coverage for these changes, but maybe the newly added -r option would be worth adding a check for?  (in that same test script I guess).

dbrolley:
@natoscott, I went ahead and merged this. I took a look at qa/667, but couldn't figure out how to test if the -r option is working properly. Same output as the final test in the script?

natoscott:
> I went ahead and merged this

@dbrolley heh, there's no rush, take yer time.  Yep, I think same output makes sense from a cursory glance.  And just realised we also need a man page update here for the new option.

natoscott:
@dbrolley I'll add the test case
@test-account-0 could you update the man page for the new option?  Also, can you send through author details (name and email address) for these commits please?  There's some concern about tracking who/where code is coming from, and we like to give commit log recognition to contributors too.

Thanks!

dbrolley:
@test-account-0, Thanks again for contributing, but we really do need your name before can continue to merge your work, for copyright considerations and also to recognize your contribution.

test-account-0:
I think I'm not ready for it. Is it really such a big problem?

PR #87, opened 2016-04-29
Change the blank length of pmiostat's output

wuliming:


lberk:
Hi, is there a particular reason for this output change?

The change also breaks qa 842 and 1099 (easy fix as it's simply whitespace), but it would be great to have a commit with those changes as well if there's a reason to update the blank length.

goodwinos:
the timestamp printed with the -x t  option is 24 characters. So how come you want to change it to 28?

wuliming:
Hi ：

I think the blank is too little(just one blank as follow) between  first column and second column.
# Timestamp              Device        rrqm/s  wrqm/s    r/s    w/s    rkB/s    wkB/s avgrq-sz avgqu-sz   await r_await w_await %util
Thu May  5 01:33:38 2016 sda              0.0     0.0    0.0    2.0      0.0      4.0     2.00     0.00     2.5     0.0     2.5   0.5

the timestamp printed with the -x t option is 24 characters. So how come you want to change it to 28?

goodwinos:
On 05/05/2016 03:35 PM, wulm wrote:
> Hi ：
>
> I think the blank is too little(just one blank as follow) between first column and second column.
> # Timestamp Device rrqm/s wrqm/s r/s w/s rkB/s wkB/s avgrq-sz avgqu-sz await r_await w_await %util
> Thu May 5 01:33:38 2016 sda 0.0 0.0 0.0 2.0 0.0 4.0 2.00 0.00 2.5 0.0 2.5 0.5

well, the timestamp lines up with the timestamp column header, and the device name (sda in your example)
lines up with the 'Device' column header .. and space is rather precious in this report.

lberk:
Closing due to inactivity, perhaps the spacing could be an argument to be specified/passed in the event there are alternate personal preferences.

Issue #90, opened 2016-05-13
Problems using mmv with Python


taintedkernel:
Registering this issue after discussing the problem in #pcp without any resolution.

After discovering the ability to instrument Python code with PCP, I attempted to do so with example code provided in the module - **src/python/pcp/mmv.py**.  With a few modifications to get it to run, I had problems retrieving the data with pminfo.  I also compiled the acme.c example, which worked just fine - **pminfo -f mmv** saw it without problems (and **/var/lib/pcp/tmp/mmv/acme** had data).  Strace returned nothing for **/var/lib/pcp/tmp/mmv** on the Python code, so something appears to be broken on the binding.

This is on a RHEL6 box with latest packages (3.11.3 as of writing) built from scratch via Makepkgs and Python2.6.  I can provide the full strace or example Python code if necessary.

fche:
At least one bug evident in the python mmv bindings:

diff --git a/src/python/pcp/mmv.py b/src/python/pcp/mmv.py
-LIBPCP_MMV.mmv_lookup_value_desc.restype = pmAtomValue
+LIBPCP_MMV.mmv_lookup_value_desc.restype = POINTER(pmAtomValue)

which triggers a stack-smashing crash under glibc/ffi

fche:
and this one too, discovered concurrently by Anothony and me:

--- a/src/python/pcp/mmv.py
+++ b/src/python/pcp/mmv.py
@@ -143,6 +143,7 @@ LIBPCP_MMV = ctypes.CDLL(ctypes.util.find_library("pcp_mmv"))
             helptext = helptext.encode('utf-8')
         if type(shorttext) != type(b''):
             shorttext = shorttext.encode('utf-8')
+        self.name = name

taintedkernel:
Actually see a few other issues as well, going to continue to work on this.

natoscott:
Oops, meant to comment on close: "all should be resolved in pcp git master branch now ... lemme know if issues persist"

PR #91, opened 2016-05-13
Python bindings: Fixing bug where name isn't set in mmv_metric

taintedkernel:
Fixes problem at https://github.com/performancecopilot/pcp/issues/90


taintedkernel:
Found other issues, will open new request once fixed.

PR #92, opened 2016-05-13
Fixes crashes and problems updating metrics

taintedkernel:
@fche and I were able to fix issues with the Python bindings.  These patches fixed the problems on my end.

### DIFF
--- a/src/python/pcp/mmv.py
+++ b/src/python/pcp/mmv.py
@@ -10,7 +10,7 @@
 # under the terms of the GNU General Public License as published by the
 # Free Software Foundation; either version 2 of the License, or (at your
 # option) any later version.
-# 
+#
 # This program is distributed in the hope that it will be useful, but
 # WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
 # or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
@@ -148,6 +148,7 @@ class mmv_metric(Structure):
         self.typeof = typeof
         self.indom = indom
         self.item = item
+        self.name = name
 
 ##
 # PCP Memory Mapped Value Services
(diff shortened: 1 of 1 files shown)


natoscott:
@taintedkernel thanks for looking into this.  It'd be good to add a regression test or two for this code - do you have some sample python code you've been testing with that you could send through?  (I'll add it into the QA suite if so, thanks!)

fche:
> do you have some sample python code?

Try the sample code - commented out - in `src/python/pcp/mmv.py`

natoscott:
| Try the sample code - commented out - in src/python/pcp/mmv.py 

(this is test case qa/704 and qa/src/test_mmv.py) - I'm more specifically looking for test code that exercises the failing situations here, thanks.

taintedkernel:
The code in mmv.py mostly works, but I had to make a few minor
modifications to get it to run correctly.  I'll be able to send over
tomorrow what I have put together thus far.

natoscott:
> I'll be able to send over tomorrow what I have put together thus far.

@taintedkernel great, thanks!

natoscott:
Ah, I also just found that test qa/704 is marked as "reserved" & not actively being run, so it may well become healthy with these fixes & provide the coverage we need - will take a look.

@taintedkernel any other test code you send through will still be helpful and used too though, thanks.

taintedkernel:
Taking a second look, there's only really 2 issues with the example code in
mmv.py.  The typeof and semantics constants weren't defined so I changed
them to reference the PM_ types in cpmapi.  Also, the call to
lookup_mapping requires an instance string as the second argument (None
will raise an AttributeError).

Attached is example code with the fixes.  I'm putting together a short demo
for my team with this functionality and may be able to also send over that
code afterwards.

natoscott:
| The typeof and semantics constants weren't defined so I changed them to reference the PM_ types in cpmapi

Ah, the commented out sample code is missing an "import cmmv" - I'll fix that up.  In fact, maybe a different sample program entirely is in order here.

| Also, the call to lookup_mapping requires an instance string as the second argument (None will raise an AttributeError). 

Yes, I came across that one yesterday when working on regressions tests here - so that one's fixed already.

PR #96, opened 2016-06-17
Document default pmwebd context timeout

scotte:
The default pmwebd context timeout of 300 seconds is not documented in the manpage.


fche:
thanks, lgtm!

PR #97, opened 2016-07-06
pcp-pidstat: initial commit of PCP implementation of pidstat

sitaramshelke:


natoscott:
Nice work Sitaram, looks good!

I'm seeing a failure in the new test qa/1078 - looks like two small things:
- We need to use -z to provide deterministic timestamps in the test output (I have a fix for this one)
- The individual processes are output in a different order for me - I guess different python version or different dictionary/hashing algorithm, or something like that.  Is there a way to sort the pcp-pidstat output?  (e.g. by pid or by top cpu user, or something like that - just to get that deterministic output).  I'm not seeing it from a quick man page scan.

cheers.

PR #277, opened 2017-02-19
perfevent_pmda: Add PMU events exported from kernel

hkshaw1990:
The perfevent pmda gets the event definitions from the libpfm
library. So, to add a new platform counter, we need to add its
definition to the libpfm library and use it here. This patch is the
first step towards removing that dependency. It parses the kernel
exported PMUs and their events from /sys/bus/event_source/devices, and
calculates the config values for each of these events. However, we don't
want to open all the events at once. Only the events which are mentioned
in perfevent.conf under [dynamic] section will be opened. But, all the
dynamic events will be listed with pminfo.
For e.g. :
 $ pminfo
[...]
perfevent.hwcounters.software.context-switches.dutycycle
perfevent.hwcounters.software.context-switches.value
perfevent.hwcounters.software.page-faults.dutycycle
perfevent.hwcounters.software.page-faults.value
[...]
perfevent.hwcounters.power.energy-cores.dutycycle
perfevent.hwcounters.power.energy-cores.value
perfevent.hwcounters.power.energy-ram.dutycycle
perfevent.hwcounters.power.energy-ram.value
[...]
perfevent.hwcounters.msr.mperf.dutycycle
perfevent.hwcounters.msr.mperf.value
perfevent.hwcounters.msr.aperf.dutycycle
perfevent.hwcounters.msr.aperf.value
[...]
perfevent.hwcounters.cpu.mem-stores.dutycycle
perfevent.hwcounters.cpu.mem-stores.value
perfevent.hwcounters.cpu.tx-capacity.dutycycle
perfevent.hwcounters.cpu.tx-capacity.value
perfevent.hwcounters.cpu.mem-loads.dutycycle
perfevent.hwcounters.cpu.mem-loads.value
perfevent.hwcounters.cpu.branch-misses.dutycycle
perfevent.hwcounters.cpu.branch-misses.value
[...]

Now, say only the event cpu.branch-instructions is mentioned under
[dynamic] section in the perfevent.conf file. In this case, we can only
read the values for this event, all other dynamic events are disabled
and will return invalid value.

$ pmval perfevent.hwcounters.cpu.branch-misses.value

metric:    perfevent.hwcounters.cpu.branch-misses.value
host:      <>
semantics: cumulative counter (converting to rate)
units:     count (converting to count / sec)
samples:   all

              cpu0                  cpu1                  cpu2                  cpu3
            3.028E+05             3.772E+05             3.757E+05             2.638E+05
            4.382E+05             2.716E+05             2.888E+05

cpu.mem-loads is not added in the perfevent.conf file, so we won't be
able to get any values out of it, even though, we can see it under pminfo.

$ pmval perfevent.hwcounters.cpu.mem-loads.value

metric:    perfevent.hwcounters.cpu.mem-loads.value
host:      <>
semantics: cumulative counter (converting to rate)
units:     count (converting to count / sec)
samples:   all
No values available
No values available

This patch also adds software events whose configuration values are
available under linux/perf_event.h :
[...]
perfevent.hwcounters.software.cpu-migrations.dutycycle
perfevent.hwcounters.software.cpu-migrations.value
perfevent.hwcounters.software.context-switches.dutycycle
perfevent.hwcounters.software.context-switches.value
perfevent.hwcounters.software.page-faults.dutycycle
perfevent.hwcounters.software.page-faults.value
[...]

### DIFF
--- a/man/man1/pmdaperfevent.1
+++ b/man/man1/pmdaperfevent.1
@@ -46,6 +46,11 @@ The configuration
 file allows different counters to be programmed on different CPUs and supports
 round-robin assignment of uncore counters needed for some AMD chips.
 .PP
+The PMDA also counts for events exposed in the kernel via
+/sys/bus/event_source/devices directory. The PMU device name and the event
+name must be mentioned in the configuration file. Otherwise, the metric won't
+be available to monitor through this PMDA.
+.PP
 The PMDA configures the counters to count events in both user and kernel mode.
 This means that the hardware counters are unavailable to use by normal
 unprivileged user applications when they are in use by the PMDA.
--- a/qa/perfevent/GNUmakefile
+++ b/qa/perfevent/GNUmakefile
@@ -32,6 +32,7 @@ SRCS = perf_event_test.c \
        rapl-interface.c \
        configparser.yytest.c \
        $(SRCDIR)/perfinterface.c \
+	$(SRCDIR)/parse_events.c
 
 THREAD_SRCS = threadtest.c \
 	      $(SRCDIR)/perfmanager.c \
--- a/qa/perfevent/config/test_dynamic_counters.txt
+++ b/qa/perfevent/config/test_dynamic_counters.txt
@@ -1,0 +1,8 @@
+[pmuname ]
+INSTRUCTIONS_RETIRED cpu
+UNHALTED_REFERENCE_CYCLES cpu_rr
+
+[dynamic]
+pmu1.bar
+pmu2.cm_event1
+pmu2.cm_event3
--- a/qa/perfevent/perf_event_test.c
+++ b/qa/perfevent/perf_event_test.c
@@ -9,6 +9,7 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include "parse_events.h"
 #include "mock_pfm.h"
 
 void test_init()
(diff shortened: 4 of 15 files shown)


natoscott:
Hi Hemant,

Sounds good.  What is the docs & QA plan?  i.e.  Do the existing tests still work?  Does the existing pmdaperfevent.1 man page need an update?  And what tests can we add to exercise this new functionality?

Thanks mate!

hkshaw1990:
Hi Nathan,
Will add some tests to it but not sure what's the best way to do that? Shall I just create a mock sysfs for the perf events so that this agent can read from there? Will that be fine? I will send a separate patch regarding the docs and qa, once you let me know this patch is fine.

Thanks!

natoscott:
The approach used seems valid to me @hkshaw1990 

For testing, esp. fetch_events, parse_event_string and friends - we could have a testing mode where we prefix DEV_DIR ("/sys/bus/event_source/devices/") with a path, allowing injection of test paths/files.  See the Linux PMDA and the LINUX_STATSPATH env variable which is used by a number of tests.

In this case, a LD_LIBRARY_PATH interceptor for perf_event_open so that actual kernel calls are not made for the faked event counters would be needed too I think.

Unrelated note, but this part may not be valid ...

--- a/src/pmdas/perfevent/pmda.c
+++ b/src/pmdas/perfevent/pmda.c
@@ -543,7 +545,7 @@ static char *normalize_metric_name(const char *name)
     {
         for(p = res; *p != '\0'; p++)
         {
-            if( !isalnum((int)*p) && *p != '_')
+            if( !isalnum((int)*p) && *p != '_' && *p != '.' && *p != '-')
             {
                 *p = '_'; /* "underscore" - new name */
             }

The metric name components should not contain dots or hyphens - see the
SYNTAX section on pmns(5) man page.

fche:
In addition to any pre-baked /sys/bus fake tarball for qa purposes, please consider having another test that actually exercises the code against the real running kernel.  We've had situations where the pre-baked data gives a false assurance, hiding breakage against changes in the real system.

hkshaw1990:
@natoscott thanks for the review Nathan.
Will add the suggested QAs for the perfevent.
While I agree that, a hyphen is illegal for a metric name to have, but, why a dot? The reason I am a bit unhappy to change everything to underscore because of some resultant metrics names. There is a "cpu" pmu  in /sys/bus/event_source/devices and "cpu-cycles" is one of its events. If every disallowed character is changed to underscore then the resultant name is :
perfevent.hwcounters.cpu_cpu_cycles.value
So, atleast, can we have the dot separating the PMU name and the event name?

@fche Agreed to have some test case to check against a real running kernel. But, one issue which we can face there is how do we compare the test results? Since, we don't know what or how many events are exposed in that kernel. Can you please provide some suggestions to it?

Thanks!

fche:
> how do we compare the test results?

You'd have to have some logic in the test driver shell script to sanity-check the output, instead of naive comparison against an .out file.  You could compare the number of counters between pcp and perf -l; you could assert a minimum number of non-zero counts; that sort of thing.

natoscott:
@hkshaw1990 yes, dots are OK as component separators within metric names - just its not within a metric name component (this is a low level detail when adding metric names on the fly - wasn't clear if it will affect you here, so just thought I'd mention it).  You can test this by ensuring the PMNS can be expanded (e.g. via pminfo/pmprobe) at any point where this is a dot.

Regarding testing, you can have arbitrary complexity in a QA test via custom qa/src code, then have the helper program produce deterministic "OK" output if all is well and detailed diagnostics on failure.  This still fits in with the "naive comparison" model just fine.  In the perfevent case though, it may be overkill though - one can sometimes end up spending more time chasing kernel/hardware issues, test workarounds and platform-idiosyncrasies than doing useful testing - so, up to you to make a call there as the code author.

The final thing to keep in mind is that there is likely to be other people (like myself, and the other PCP maintainers) who run the tests frequently and will need to debug test failures.  So there's a balance to be struck here, it's not a one-size-fits all situation.

PR #278, opened 2017-02-20
Tidy up of Vagrantfile & Start of OSX Sierra base box

tallpsmith:
My intent started off to get a working stable cleanroom environment for building PCP on OSX.

I've leveraged a pattern found on the interwebs for using an OSX Sierra base image which is then bootstrapped further with some other installations via Homebrew.

I've elected to tidy up the Vagrantfile in the process, pulling out a lot of inline scripts into dedicated files so that the file is easier to read.

At the moment, the provisioning of the OSX Sierra box takes FLIPPING AGES:

* Firstly, the base image is 20GB, so get a coffee and chill with Netflix while that downloads the first time...

* the `brew install` of several of the needed components takes _*FOREVER*_ - in particular `gettext` and `qt5`.  My strategy will be to supersede this base image with a newer pre-built OSX Base box with these pre-installed.  This won't help the initial download of the base box though..  I'm not sure where we can store this image, perhaps bintray as Nathan outlined.

----

For some reason this is coming up with a merge conflict but I don't know why...


ryandoyle:
Given the licensing issues with using the OSX box as per https://github.com/AndrewDryga/vagrant-box-osx#os-x-licensing and http://images.apple.com/legal/sla/docs/macOS1012.pdf, perhaps we should make it harder to use the box without the user implicitly doing so.

The sections of the PDF that are interesting are 2.A and 2.B. The way I read it is that you can only use a VM _if_ your license to the software is via 2.B. If your OSX license is granted by the way of it being pre-installed (2.A), then I don't think the allowable 2 VMs applies.

tallpsmith:
@ryandoyle I've updated the PR with the changes we discussed on Friday, so would appreciate another eyeballing.

ryandoyle:
A couple of things @tallpsmith 
- Vagrant is aborting for me when I run a `vagrant status`. I'm falling through to the `elsif(!platformOkToUseOSX)` conditional. You probably want to use `STDERR.puts` for non-fatal errors.
- A few stylistic things, Ruby convention is lower-case snake-case

I think the logic could be simplified to only inform the user that they _can_ use the OSX box if they are actually on OSX.

tallpsmith:
Ryan, factored in your feedback.

However it ends up only nagging and not _preventing_ the running...?

Several beers and boolean logic are not a good combination for me, will catch you tomorrow.

ryandoyle:
Yep looks better now :+1:

PR #279, opened 2017-02-23
initial implementation of speed container

suyash:
@natoscott I have added a simple container that installs speed and builds all examples.

Go installation is from the official library image.

### DIFF
--- a/build/containers/pcp-go-app/Dockerfile
+++ b/build/containers/pcp-go-app/Dockerfile
@@ -1,0 +1,60 @@
+#
+# Sample Dockerfile with speed installed and all the example binaries built
+#
+# Build:   $ docker build -t pcp-go-app .
+# Name:    $ sudo mkdir -p /run/containers/pcp-go-app
+# Run:     $ `docker inspect --format='{{.Config.Labels.RUN}}' pcp-go-app`
+#
+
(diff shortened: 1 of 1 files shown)


natoscott:
This is good, Suyash.  The final container is missing a running pmcd (along side the running Go program), but I realise I didn't explain that very well earlier - the end goal here is an application running in a container, with pmcd also running in the container, that could be deployed into a cloud service.

More importantly, I see the Acme example that I'd like to run here is a little different between Parfait and Speed.  In Speed we based it on the Python Acme example, which exits straight away.  What we need here is an instrumented daemon/service, written in Go and instrumented with Speed.  Could we extend the Acme example in Speed to be long-running (frequently updating the metrics), like Parfait?  (i.e. don't exit right away, but run in a loop and keep updating metric values)  e.g.

https://github.com/performancecopilot/parfait/tree/master/examples/acme/src/main/java

Assuming we can do that, I'll push an update to your PR which completes the container build (adds in pmcd, and runs it alongside the Go Acme service) and adds it into the set we release with PCP.

suyash:
@natoscott will implement a separate *acme_forever* example for speed and update this following that

natoscott:
@suyash thanks!  Or maybe a command line option to run acme --forever?

PR #285, opened 2017-04-02
Implementation of Prometheus PMDA

jronak:
Experimental implementation of a PMDA to import Prometheus metrics into PCP.
Looking forward to the suggestions/feedback.
Thanks


natoscott:
@jronak looking good!  Worked first time for me, which is always nice.  :)

The last two steps before an initial merge will be:
1. some documentation (a man page) - you might find src/pmdas/gluster/pmdagluster.1 useful as a template; in particular, the special pmdaprometheus command line options need documenting, and the 'Installation'' section of the man page should take a user through exactly what's involved.  The 'Files' section should also cover all the configuration files.
2. regression test(s) - see qa/README, the qa/new script, and the pmdagluster/pmdajson tests as some examples (egrep 'pmda.gluster|pmda.json' qa/group)

At that point, I think we'll have a perfect base from which we can (collectively) build - e.g. we could look into extending this PMDA with any number of things ...

- checking that some of the more common endpoints work well (like etcd, kubernetes, etc);
- some of the more automatic enabling of metrics that's been requested, when no configs exists;
- label support - great to see you've got prometheus labels embedded in compound instance names already, and its immediately clear these will map nicely to the pending PCP label concept too.

cheers.

ps: I'll send you a little patch with some small cleanups I noticed.  Also, down the track you might want to run pylint(1) on the script - it's very opinionated in some regards (re code style), but does pick up some genuine problems too sometimes, so worth going through the volumes of warnings it usually issues.

src/pmdas/prometheus/pmdaprometheus.1
```diff
@@ -0,0 +1,107 @@
...
+        "semantic": "instantaneous",
+        "prometheus_name": "go_goroutines",
+        "desc": "Number of goroutines that currently exist.",
+        "units_str": ""
+      }
+    ]
+  }
+.fam T
+.fi
+.PP
```
natoscott: @ronak at this point I think it would be good to document the meanings of each of the fields, and the allowable values for each.  e.g. semantics -> counter, instantaneous and discrete, units -> mention pmParseUnitsStr(3) and so on.

src/pmdas/prometheus/pmdaprometheus.1
```diff
@@ -0,0 +1,107 @@
...
+.nf
+.fam C
+  {
+    "metrics": [
+      {
+        "type": "double",
+        "name": "go_goroutines",
+        "semantic": "instantaneous",
+        "prometheus_name": "go_goroutines",
+        "desc": "Number of goroutines that currently exist.",
```
natoscott: @jronak I suggest using a different name to "desc" here, because that could be confused with "descriptor", as in pmLookupDesc(3).  Perhaps "oneline_text" and "help_text" for the pmLookupText(3) fields.

src/pmdas/prometheus/pmdaprometheus.1
```diff
@@ -0,0 +1,107 @@
...
+The associated metadata file would look like the following:
+.PP
+.nf
+.fam C
+  {
+    "metrics": [
+      {
+        "type": "double",
+        "name": "go_goroutines",
+        "semantic": "instantaneous",
```
natoscott: In all the PCP docs, the term "semantics" tends to be used, so I'd make this match up (rather than "semantic")

src/pmdas/prometheus/pmdaprometheus.1
```diff
@@ -0,0 +1,107 @@
...
+.fi
+.PP
+The associated metadata file would look like the following:
+.PP
+.nf
+.fam C
+  {
+    "metrics": [
+      {
+        "type": "double",
```
natoscott: Good example.  Based on the help text and metric name here, I'd imagine this metric is actually a "uint64" internally in Go, and not a "double" value?  (i.e. its describing an integer, not a floating point value, so I'd be very surprised if this is maintained as a "double" within the golang runtime).

src/pmdas/prometheus/pmdaprometheus.1
```diff
@@ -0,0 +1,107 @@
...
+.TH PMDAPROMETHEUS 1 "PCP" "Performance Co-Pilot"
+.SH NAME
+\f3pmdaprometheus\f1 \- Prometheus PMDA
+.SH DESCRIPTION
+\fBpmdaoprometheus\fR is a Performance Metrics Domain Agent (PMDA) which imports
+metrics from Prometheus endpoints which expose text based Prometheus metrics.
+.PP
+Further details on the Prometheus exposition format can be found at https://prometheus.io/docs/instrumenting/exposition_formats
+.SH METADATA SOURCES
+\fBpmdaprometheus\fR scans the $PCP_PMDAS_DIR/prometheus/metadata directory, looking for metadata files (*.json). These JSON metadata files describe the names, types, etc. of their associated Prometheus metrics for a given endpoint. For example, let's assume the following simple Prometheus metric:
```
natoscott: Paths are usually printed underlined in PCP man pages, so that would be:

.B pmdaprometheus
scans the
.I $PCP_PMDAS_DIR/prometheus/metadata
directory, looking for metadata files (\c
.IR *.json
).

Each line should be at most 80 characters (in the raw troff format), so I'd split the above line into multiple lines for text editor readability.

src/pmdas/prometheus/pmdaprometheus.1
```diff
@@ -0,0 +1,107 @@
...
+.fam C
+  {
+    "metrics": [
+      {
+        "type": "double",
+        "name": "go_goroutines",
+        "semantic": "instantaneous",
+        "prometheus_name": "go_goroutines",
+        "desc": "Number of goroutines that currently exist.",
+        "units_str": ""
```
natoscott: "units_str" would be more readable as simply "units"?

src/pmdas/prometheus/pmdaprometheus.1
```diff
@@ -0,0 +1,107 @@
...
+    ]
+  }
+.fam T
+.fi
+.PP
+.SH METADATA GENERATOR
+.PP
+PMDA Prometheus metadata generator helps in generating the heuristic metadata for the Prometheus metrics endpoint.
+.PP
+.B $PCP_PMDAS_DIR/prometheus/pmdaprometheus.python
```
natoscott: BTW, to make it easier for the user to understand (and easier to document), this could be a shell script with a more descriptive name ("$PCP_PMDAS_DIR/prometheus/generator"?), ...

#!/bin/sh
. /etc/pcp.env
exec $PCP_PYTHON_PROG $PCP_PMDAS_DIR/prometheus/pmdaprometheus.python -g "$@"

src/pmdas/prometheus/pmdaprometheus.1
jronak: @natoscott as we have seen the Golang struct of prometheus metric, it was quite clear that all the values are float64(double) except for the suffix _count metric in summary and histogram is of U64. Also, I have seen metrics represented in integer format when it doesn't have value after decimal point.

src/pmdas/prometheus/pmdaprometheus.1
jronak: I have renamed 'desc' to 'description'.

src/pmdas/prometheus/pmdaprometheus.1
jronak: units could be confusing as if it were a list of numbers [0,0,1,-1,1], rather _str makes it clear that the units will be in the string format.

src/pmdas/prometheus/pmdaprometheus.1
jronak: Should this be added as a separate script in the Prometheus folder or only for the man page ?

natoscott: Separate script which we install alongside pmdaprometheus, and document in the man page.  Some of the other PMDAs do similar things - e.g. pmdammv had a little mmvdump helper utility we install alongside it; pmdaoracle has a helper "setup" checking script, and so on.

src/pmdas/prometheus/pmdaprometheus.1
natoscott: Ah I see - yep, makes sense then if we're supporting both forms (which seems like a good idea to me).

src/pmdas/prometheus/pmdaprometheus.1
natoscott: Right, that was the prometheus exporter code we looked at though, which is simply exporting underlying values from the Go runtime in this case (blindly exporting most things as float64).  And in this particular case, we can be quite certain this value will not exist as a float64 within the Go runtime, that just wouldn't make sense.

So, it's a good example for the man page.  It shows the need to allow overrides on a case-by-case basis, so we can correctly represent values like this one (and no doubt many others) with appropriate PCP types, semantics, units - even when the heuristics fail.

PR #301, opened 2017-05-26
A couple of new pmmgr options

minnus:
hostid-static:
For when the host to be monitored doesn't know what it should be called.  Used in monitoring a VM farm where the name should be an externally generated uuid.

pmlogger-timefmt:
Allow the time format in the filename to be specified.

We have been running both of these for a while, and have seen no ill effects.


fche:
LGTM generally; the "hostid-static" docs should disclose that setting it will force pmmgr to interpret -all- target pmcds it encounters as the same single static hostid, so this is only useful in a single-target configuration.

minnus:
Thanks for the feedback Frank.  Should be addressed now.

With respect to the pmcd -H, I wasn't aware of that, and good to now know.  But in practice that would be cumbersome to implement in this specific environment.

natoscott:
@minnus looks like these changes needs some automated QA?

minnus:
OK, I think this should exercise the 2 new new options

PR #302, opened 2017-05-26
pmlogger label hostname override

minnus:
Not ready for merge yet, but just looking for comments if there would be any negative consequence to doing his.  When monitoring a remote VM farm, the hostname can be changed by the user to any value.  When logging many VMs centrally, these leads to 2 issues:

1:
We cannot uniquely identify an archive by hostname, since multiple machines may have the same hostname.

2:
pmlogextract fails to merge archives with different hostnames in the label.

When monitoring these hosts, by setting the hostname in the label to an externally generated uuid, we can solve both of these issues.

We have been running this for a week or so and it seems to work fine.


natoscott:
Is pmlabel(1) use an option here @minnus ?  If not, seems OK to me.  Might want to add a check for exceeding PM_LOG_MAXHOSTLEN too.

| We have been running this for a week or so and it seems to work fine.

This'll need an automated regression test - should be fairly straight-forward though (use pmlabel and/or pmdumplog -l to verify the right hostname is used).

minnus:
@natoscott OK, should be all set now with QA and man page

fche:
(Probably too late now, but I wonder if perhaps such an option would be even more appropriately placed within pmlogextract itself - that way, if given, it would both set the output archive's loglabel.hostname field, and suppress the 'must have same hostnames on the input archives' check.)

minnus:
And a few more fixes.  I missed the pmnewlog stuff and associated QA fallout.  I don't necessarily understand which options should get passed through by pmnewlog, but I think I got it right.  Let me know if I missed anything else.

PR #310, opened 2017-06-15
Additional slave metrics for MySQL PMDA

rlex:
Corresponding lines from SHOW SLAVE STATUS:
```
SQL_Delay: 0
SQL_Remaining_Delay: NULL
```

One problem i noticed is that SQL_Delay and SQL_Remaining_Delay can be both int and NULL.
This also applies to Seconds_Behind_Master which is PM_TYPE_U32 right now.

### DIFF
--- a/src/pmdas/mysql/pmdamysql.pl
+++ b/src/pmdas/mysql/pmdamysql.pl
@@ -1801,6 +1801,12 @@ $pmda->add_metric(pmda_pmid(3,33), PM_TYPE_U32, PM_INDOM_NULL,
 $pmda->add_metric(pmda_pmid(3,34), PM_TYPE_U32, PM_INDOM_NULL,
 		  PM_SEM_DISCRETE, pmda_units(0,0,0,0,0,0),
 		  'mysql.slave_status.master_ssl_allowed_num', '','');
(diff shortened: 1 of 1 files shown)


natoscott:
@rlex I've tweaked the metric semantics to instant (these aren't discrete values) and the units to reflect the documentation (seconds).

I believe the behavior of seeing null in those two fields would be that PM_ERR_APPVERSION is returned for that metric when no value is available (can you confirm?).  We can tweak that behavior by testing for undef values for those metrics in mysql_fetch_callback() if you think it should be different?  (perhaps an error code like PM_ERR_AGAIN - "Information not currently available" might be more appropriate here?

cheers.

rlex:
I'm not sure. There is situations when seconds_behind_master = NULL is normal situations, IE if you stopped replication for some reason (upgrade, maintenance, whatever) or stopped one of replication threads. SQL_Remaining_Delay will be NULL if you don't use delayed replication (which is like 9 from 10 replication installations), so that's OK as well, so i don't think treating Null as error is good idea.
SQL_Remaining_Delay and SQL_Delay is available from 5.6 AFAIK: https://dev.mysql.com/doc/refman/5.6/en/replication-delayed.html

PR #322, opened 2017-07-16
pmdaprometheus: Parallel http endpoint fetches

jronak:


src/pmdas/prometheus/pmdaprometheus.python
```diff
@@ -662,6 +660,7 @@ class PrometheusPMDA(PMDA):
 		PMDA.__init__(self, pmda_name, domain)
 		self.connect_pmcd()
 		self.numfetch = 0
+		self.num_threads = 8
```
fche: Question: how did you come up with the number 8?  Are the cost of these python threads so high that we can't create one for each "cluster" (even if there are hundreds)?

PR #326, opened 2017-08-02
Add nfsd worker thread status values to Linux PMDA

jumitche:
The values from /proc/fs/nfsd/pool_stats allow the monitoring of the nfs server workload and can show if insufficient threads have been configured


lberk:
Thanks for adding the QA!  Do you mind adding it to the pmda.linux group (in the qa/group file).  128 is now failing for me on platforms without nfs metrics.

src/pmdas/linux/proc_fs_nfsd.c
```diff
@@ -0,0 +1,98 @@
...
+			&woken, &timedout)) == 5) {
+
+		    /* poolid is not important, just count them */
+		    proc_fs_nfsd->pool_cnt++;
+		    proc_fs_nfsd->pkts_arrived += arrived;
+		    proc_fs_nfsd->sock_enqueued += enqueued;
+		    proc_fs_nfsd->th_woken += woken;
+		    proc_fs_nfsd->th_timedout += timedout;
+	    	}
+	    	if (proc_fs_nfsd->pool_cnt < 1)
```
lberk: Why < 1 ?
On a local box exporting nfs stats the proc file seems to be (properly) exporting 0 for pool count

src/pmdas/linux/pmda.c
```diff
@@ -2006,6 +2008,36 @@ static pmdaMetric metrictab[] = {
...
+    PMDA_PMUNITS(0,0,1,0,0,PM_COUNT_ONE) } },
+
+/* nfs.server.threads.pools */
+  { &proc_fs_nfsd.pool_cnt,
+    { PMDA_PMID(CLUSTER_NET_NFS,72), PM_TYPE_U32, PM_INDOM_NULL, PM_SEM_COUNTER,
+    PMDA_PMUNITS(0,0,1,0,0,PM_COUNT_ONE) } },
+
+/* nfs.server.threads.requests */
+  { &proc_fs_nfsd.pkts_arrived,
+    { PMDA_PMID(CLUSTER_NET_NFS,73), PM_TYPE_U32, PM_INDOM_NULL, PM_SEM_COUNTER,
```
lberk: Checking the kernel source, svc_xprt.c ~ svc_pool_stats_show some of these look like unsigned longs.  In pcp we have KERNEL_ULONG instead of the PM_TYPE_* to deal with that.  Mind changing this one as well as sockets-enqueued, threads_woken and threads_timeout?

src/pmdas/linux/proc_fs_nfsd.c
jumitche: That code should only be reached if the thread count is non-zero, in which case there must be at least one pool of threads defined. If pool_threads is zero then pool_stats is not a valid file (No such device) and we never get that far.

are you perhaps confusing the first column of pool_stats which is a pool id number.  I do not attempt to report that number, instead i count how many pools there are and aggregate the scores. even though technically there is only ever 0 or 1 pools right now.

jumitche:
I would appreciate some feedback if i got the right qa number to add my tests to?  128 initially looked the most likely, but the more i look at it the more it seems that the rest of the test does not work under current linux distros anyway, with get_nfsstat outputting nothing. should i move my tests elsewhere ?

lberk:
Hey @jumitche,

LGTM outside of qa.

qa/128 isn't quite right.  Looks like it's an irix specific testcase (check out qa/group for the general listings).

It's probably easiest to write a new test (check out the ./new script to make this a bit easier) and add it to the pmda.linux group.  We can tailor this test to specifically target the nfs.server bits.
First, I might tar the two /proc files of interest (with interesting values), we can use this tarball to check nfs.server metrics function properly even on a box that don't have nfs exports enabled.  qa/732 is a good example here, note that LINUX_STATSPATH is set and exported to point pmdalinux at the testcase /proc.
Please feel free to let me know if you have any questions

jumitche:
@lberk thanks for the guidance on the qa tests. It turned out that my patch caused qa/732 to break as that already covers the nfs.server tree, so i have added my additional values to that test instead and squashed the commits down to one. Hope that works better.

lberk:
Merged upstream, thanks!

PR #336, opened 2017-08-15
Labels support for pmwebd and pmdaprometheus

jronak:


fche:
Curious about what y'all have planned with respect to labels support in the pwmebd prometheus-exporter code.  There was a marker in the code where the current labels could be fetched & relayed.

natoscott:
@jronak I've merged the PMDA changes into the labels branch, they all look good - thanks!

I think the cleanest web API extension will involve adding a new _labels API rather than adding to just _metric and _indom.  Those are only parts of the label hierarchy, so don't really make sense exported on their own like that.  The API response change for _metric and _indom might cause headaches for some users too I imagine.

Similarly the Prometheus side of things will need more work than we have currently - the code only pmGet*'s labels for leaf nodes of the hierarchy, which doesn't represent the situation correctly (more details in the pmLookupLabels(3) and pmMergeLabelSets(3) man pages).

In general though, I think we have the cart before the horse here.  There's alot of missing foundation code we still need (archive support, optional context labels, fetch protocol extension, query support) that we need to have in place before we can decide how  best to tackle the higher level things like pmwebd.

So for now let's set the web API aspects aside and continue on with that core functionality first, then circle back to the web APIs later.

fche:
Of all those foundational missing pieces, which ones impact pmwebd's ability to fetch labels for a given metric/instance?  Archives and queries don't relate.  What do you have in mind with respect to fetches?  Any new PMAPI?

PR #338, opened 2017-08-30
pmrep modularization, add pcp2{es,json,xml,xlsx,zabbix}

myllynen:
A new tightly coupled PMConfig class for Python utilities is introduced (perhaps think of C++ friend classes) which will take care of reading the configuration file, populating the metrics data for the utility, reading command line (remember that this required quite some effort earlier to get it right for all cases), registering derived metrics, and validating the metrics in the PMNS. Utilities will use the "standard" pmrep config file format and remain mostly compatible with pmrep and each other (utility specific options are of course not recognized by other tools, e.g., zabbix_port makes only sense with pcp2zabbix).

Another module with methods that can be called directly is added to set timezone, step mode, and PMAPI context connection. There seems to be little in this area which would be different for most Python utilities.

Zabbix output support is dropped from pmrep (it was marked as experimental in the man page so should be fine) and a set of new utilities are being added to demonstrate how straightforward it becomes to add new bridge/connector utilities feeding PCP metrics data to external systems:

 * pcp2es
 * pcp2json
 * pcp2xml
 * pcp2xlsx
 * pcp2zabbix

### DIFF
--- a/qa/1068
+++ b/qa/1068
@@ -1,8 +1,8 @@
 #!/bin/sh
 # PCP QA Test No. 1068
-# Exercise pmrep Zabbix export reporting modes.
+# Exercise pcp2zabbix.
 #
-# Copyright (c) 2015 Red Hat.
+# Copyright (c) 2015-2017 Red Hat.
 #
 
 seq=`basename $0`
@@ -16,7 +16,7 @@ $python -c "from collections import OrderedDict" >/dev/null 2>&1
 [ $? -eq 0 ] || _notrun "python collections OrderedDict module not installed"
 
 which socat >/dev/null 2>&1 || _notrun "socat binary not installed"
-which pmrep >/dev/null 2>&1 || _notrun "pmrep not installed"
+which pcp2zabbix >/dev/null 2>&1 || _notrun "pcp2zabbix not installed"
 
 status=1	# failure is the default!
 signal=$PCP_BINADM_DIR/pmsignal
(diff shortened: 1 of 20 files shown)


natoscott:
Looks good Marko.  A few notes I made as I read through the changes ...

- pmconfig.PMConfig
The naming convention here looks a bit odd compared to e.g. "pmapi.pmContext"
Could we make that pmconfig.pmConfig instead?  Or maybe pmConfigFile so as not to be confused with pmGetConfig(3)?  Also, perhaps this should be another "convenience class" in pmcc.py instead of a separate module on its own?

- pcp2es
"es" is not very descriptive - let's use either pcp2elastic (as in elastic.co) or pcp2elasticsearch (as in pmdaelasticsearch(1) already in PCP). - I slightly prefer the latter I guess, but either or.

- pmcommon
There doesn't seem to be enough code here to justify a separate module?  Perhaps shift this directly into pmapi.py, or pmcc.py.  If the former - could use either @staticmethods on pmContext (there's a number of those already toward the end of the class), or as a separate class (or static helpers) in that file.

In the case of do_connect() I think that could make sense residing somewhere close to pmapi.pmContext.fromOptions() since there's so much similarity between the two.

On a somewhat related note, as we discussed earlier I've annotated pmsubsys as deprecated now, and will remove it from the pcp-4 branch shortly.

Other than the above little things, this all looks like its headed in the right direction to me.  Needs some docs, maybe more tests?, and I think we'll need to tweak the spec files to create some new packages with the correct import module dependencies.

cheers.

myllynen:
Thanks for the feedback and sorry for the delay, I've been busy and side-tracked with some other issues.

- Yes, braino/typo, let's use pmConfig, since we're also validating metrics and such. I'd like to avoid mixing this with pmcc.py as it has its own metrics from the early days instead of pmfg.

- Ok, let's go pcp2elasticsearch to make it unambiguous.

- Ok, will add to pmapi.py. I'll rename do_connectio() to set_connect_opts() as that what it actually does and place it in an appropriate location in pmapi.py.

Once these are ok we certainly need man pages for each new command. Best way to write a new utility with this approach is probably to copy an existing command, diff few existing ones (like diff -u pcp2elasticsearch pcp2xml) to see what's the difference there and adjust the new tool as needed. Not sure where this would be best to document. Perhaps we can add a short note about these new tools to the Quick Guide and mention this in a passing.

QA is also needed but probably more emphasis on valid output compared to e.g. current pcp2influxdb handling of invalid metric names as pmrep covers that well already.

I'll try to push an updated pull request next week.

Thanks.

natoscott:
> Thanks for the feedback and sorry for the delay, 

No problem at all.

> [...] Once these are ok [...]

(that all sounded good to me.)

> I'll try to push an updated pull request next week.

Awesome, looking forward to it - thanks Marko.

cheers.

PR #348, opened 2017-09-20
All environment variables must be PCP_

JohnPeacockMessageSystems:
In order to override the default UNBOUND_STATS command, it must be set as an environment variable but pcp.env clearly states that any variable set in pcp.conf will be ignored if not prefixed by PCP_.


natoscott:
This environment variable is intended solely for regression testing purposed (see qa/754 which would become broken by that patch).  What's the underlying issue here - is /usr/sbin/unbound-control not the correct path on your system?

Usually if we need agent-specific configuration, a PMDA config file would be used, and the PMDA code modified to read it at startup - that's possibly what we need here?

PR #400, opened 2017-12-19
pcpqa pmwebd: use json_pp for canonicalizing (661,1225), forget blinkenlights (661,663)

fche:
Use json_pp (a perl module) to canonicalize json documents coming from
pmwebd, so the output is reproducibly sorted and still proper &
readable JSON.  While in the vicinity, drop references to the
blinkenlights demo, since it is not necessary to demonstrate 302
redirection etc.


fche:
> check_vm ... other perl JSON modules ... can they not provide reproducible output instead?

Maybe.  AFAIK they don't have an executable wrapper as a standard component.

> is to add a pmjson helper utility

As a philosophical point, I would disfavour bundling even more platform library reimplementation code into PCP.  YMMV.

> Also, this commit seems to have multiple unrelated changes - these would be better as separate commits.

Yes, two.  OK.

> I see no other automated regression checks for blinkenlights too, so am loathe to remove what little we have there.

I have argued in the past that blinkenlights should itself be removed.  The current qa code doesn't constitute any sort of operational or even installation test.

fche:
> which you've not responded to 

The "Any interest in writing pmjson?" one?  No.

> maximise the number of platforms that tests run on

The JSON::PP package is easily available via CPAN if nothing else.  Shall I add it to `admin/check_vm` to proceed?

natoscott:
| The "Any interest in writing pmjson?" one?

Not that, rather the "please don't add new packages unnecessarily" point.  Don't worry about it.

| Shall I add it to admin/check_vm to proceed?

Not necessary, thanks though.  Please open a new PR/issue for any blinkenlights discussions you'd like to have, and with details of any earlier (private?) discussions.  One of the PCP maintainers will resolve the JSON output aspects in a more PCP-appropriate fashion shortly.

cheers.

PR #405, opened 2017-12-21
pmmgr: tolerate pmcds with empty hostname

fche:
An assertion failure was formerly triggered, possibly by pmcds
advertising a blank hostname.  This code now assigns them a
hostid of "-".

### DIFF
--- a/man/man1/pmmgr.1
+++ b/man/man1/pmmgr.1
@@ -147,7 +147,7 @@ instances, how to uniquely identify them, and where state
 such as log files should be kept for each.  Ideally, a persistent and
 unique hostid string is computed for each potential target pmcd from
 specified metric values.  This hostid is also used as a subdirectory
-name for locating daemon data.
+name for locating daemon data.  The rare empty hostid is mapped to "-".
 .TP
 .I hostid\-static
 This file contains one or more lines specifying the static string that
--- a/src/pmmgr/pmmgr.cxx
+++ b/src/pmmgr/pmmgr.cxx
@@ -458,6 +458,9 @@ pmmgr_job_spec::compute_hostids (const pcp_context_spec& ctx) const
 	}
     }
 
+  if (sanitized == "") // treat empty hostid as '-', from hostid-metrics separator
+    sanitized = "-";
+
   hostids.insert(pmmgr_hostid (sanitized));
   return hostids;
 }


goodwinos:
Frank, wouldn't a pmcd advertising a blank hostname be a configuration issue (pmcd side)? That pmcd would have to be running with:  -H ""  which is an invalid hostname. The default is from pmhostname(1), which would be a system configuration issue if it was blank.

Also, what are the consequences of the assertion failure? Can that be better handled as an error?

fche:
> wouldn't a pmcd advertising a blank hostname be a configuration issue (pmcd side)?

Probably.

> That pmcd would have to be running with: -H "" which is an invalid hostname.

Maybe, but pmcd does not complain if so invoked.

> Also, what are the consequences of the assertion failure? Can that be better handled as an error?

An assertion failure, well, exits the process as usual for assert(3).  This is now not an error nor an assertion failure; a sentinel hostid ("-") is assigned instead, and pmmgr motors on.

goodwinos:
> An assertion failure, well, exits the process as usual for assert(3). This is now not an error nor an assertion failure; a sentinel hostid ("-") is assigned instead, and pmmgr motors on.

yes, what I ment was: what are the consequences of "motoring on" with an invalid hostname? Wont this lead to further peculiar behavior or errors later? We might be better with "unknown" rather than "-".

Might also be worth augmenting pmcd's handling of -H for rudimentary checking of the supplied hostname string .. since there are downstream consumers of the pmcd.hostname metric such as pmmgr that might launch something that might try to connect to it. This would verify the string as an ip addr or [a-zA-Z][a-zA-Z0-9-.]* or some such. Regardless, pmcd should do some checking, currently does none and fallback to gethostname() if the supplied string is invalid.

This of course wouldn't help pmmgr connecting to a downrev pmcd.

fche:
> what are the consequences of "motoring on" with an invalid hostname?

pmmgr per se is fine; it'll just create a funny-named directory; pmie doesn't seem to care; pmlogger is given the "-H -" option and it seems to go on too.

I chose "-" instead of an english term like "unknown" or "none" or "anonymous" or "whatever" because it is unlikely to collide with real hosts of that name, and because "-" is already used as separator in hostids that are constructed from multiple metric values.

goodwinos:
Leaving this open because I'll be also improving pmcd -H hostname checking (there is no checking currently, but pmcd should at least check the named hostname string complies with host naming standards)

fche:
> pmcd should at least check the named hostname string complies with host naming standards

Which standards?

goodwinos:
> Which standards?

well, s/standards/conventions/

especially the one that says _'thou shall not have a blank hostname'_ :)

Perhaps translate http://commons.apache.org/proper/commons-validator/apidocs/src-html/org/apache/commons/validator/routines/InetAddressValidator.html to C, or use a suitable regex, etc.

Merry Christmas

fche:
Yeah ... except system hostnames don't need to be IP addresses, names resolvable by DNS, or any other directory system.  That "thou shalt" is good advice, but I don't see it in e.g. POSIX gospel.  The linux kernel tolerates a `sethostname("",0);` system call just fine.

goodwinos:
Will add some pmcd -H hostname basic checking and appropriate warning messages

fche:
To address this scenario fully, one would also need to check against a blank gethostname(2), in case pmcd's `-H` was not used.  And a sethostname(2) need not be ASCII or friendly isalnum().
If we do not reject such machines outright, then we should check to what extent PCP's tooling (e.g., pmlogger_daily?) tolerates them.  pmmgr sanitizes them in `pmmgr_job_spec::compute_hostids`.

PR #409, opened 2018-01-05
Add PCP BCC PMDA to extract metrics from eBPF/BCC Python modules

myllynen:
PCP BCC PMDA is a plugin which extracts live performance data from eBPF
programs by using the BCC (BPF Compiler Collection) Python frontend and
provides them to any PCP client for archiving, monitoring, exporting,
and analysis purposes. It loads and acts as a bridge for any number of
configured, separate BCC Python modules embedding BPF code. Existing BCC
Python tools and modules should be possible to convert as PCP BCC PMDA
modules with reasonable effort.


natoscott:
Hi Marko,

Overall this looks good to me.  The main issue that we need to address is the duplication of the embedded bpf code from bcc/tools - that looks like a long-term maintenance headache we should tackle up front.

Have you chatted to Brendan about this PMDA yet?  If not, let's contact him to discuss ways we could approach sharing that code (e.g. extracting it into a bpf/ directory, like the lib/ directory some of the other bcc/tools scripts use, perhaps) so the pcp python code can also make use of that code without duplicating it into pcp itself.

cheers.

myllynen:
Hi Nathan,

Thanks. The purpose of these few selected modules was mainly to serve as examples/references how to create new PCP BCC PMDA modules using different kind of BPF/BCC data structures, either from scratch or from some existing BCC Python modules (either part of BCC or not). I wasn't planning to convert every BCC tool to PCP (some of them don't even produce anything resembling performance metrics). Also, some of the BPF code had to modified slightly (some BCC tools modify the BPF program on the fly based on command line parameters), I'm not sure how feasible it would be to try dynamically adjust BPF module code on the fly on PCP side if the modules ever get updated, considering also that some BCC tools offer options to provide different kind of metrics based on their command line options.

We exchanged few emails from Brendan so he's aware of this effort. Perhaps we could check what he thinks about this kind of idea, allowing PCP BCC PMDA modules merely to import bpf_text from BCC modules and perhaps calling a function to sanitize the program for PCP would in theory be nice, we could make it optional for BCC tools to support such feature so not to put too much burden on BCC contributors who may or may not be familiar with PCP.

Thanks.

myllynen:
I've now pushed a new version which splits out BPF code into separate files in preparation to use the _-e_ of the tools once available so that the BPF code does not necessarily have to be copied anymore. Thanks.

PR #424, opened 2018-01-23
pmrep.conf: Add sar-n-SOCK config

wcohen:
Add a equivalent configuration for pmrep for "sar -n SOCK".  The "pmrep :sar-n-SOCK" metrics output matches up with the sar command that it modeled after.

### DIFF
--- a/src/pmrep/pmrep.conf
+++ b/src/pmrep/pmrep.conf
@@ -609,7 +609,22 @@ network.interface.out.fifo    = txfifo/s,,,,
 
 #[sar-n-NFS]
 #[sar-n-NFSD]
-#[sar-n-SOCK]
+
+[sar-n-SOCK]
+header = yes
+unitinfo = no
+globals = no
+timestamp = yes
+width = 9
+precision = 2
+delimiter = " "
+network.sockstat.total        = totsck,,,,
+network.sockstat.tcp.inuse    = tcpsck,,,,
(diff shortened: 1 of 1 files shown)


src/pmrep/pmrep.conf
```diff
@@ -609,7 +609,22 @@ network.interface.out.fifo    = txfifo/s,,,,
...
+
+[sar-n-SOCK]
+header = yes
+unitinfo = no
+globals = no
+timestamp = yes
+width = 9
+precision = 2
+delimiter = " "
+network.sockstat.total        = totsck,,,,
```
myllynen: To align with sar -n SOCK, the width of the first column should be 12.

src/pmrep/pmrep.conf
```diff
@@ -609,7 +609,22 @@ network.interface.out.fifo    = txfifo/s,,,,
...
+network.sockstat.tcp.inuse    = tcpsck,,,,
+network.sockstat.udp.inuse    = udpsck,,,,
+network.sockstat.raw.inuse    = rawsck,,,,
+network.sockstat.frag.inuse   = ip-frag,,,,
```
myllynen: Preferably only one space between the widest metric name and the equal sign.

wcohen: Follow on patch fixes the output column alignment and the space between the metric name and the '=' has been push onto the branch.

PR #428, opened 2018-01-26
pmwebd file service: add/check ETag header for 304 Not Modified results

fche:
When pmwebd is used for ordinary file service, it's desirable to allow
web browsers to cache those files.  This patch introduces simple
tagging of these files with the ETag: header (with some readily
available fstat(2) values as keys).  It now sends a 304 result code
instead of a 200 and a whole new copy of unmodified files.

### DIFF
--- a/qa/661
+++ b/qa/661
@@ -52,7 +52,7 @@ echo "crashers"
 _filter_curl_i()
 {
     tr -d '\r' |                      # drop ^M from httpd results
-    sed -e 's,^Date:.*$,Date: XXX,' -e 's/302 Moved Temporarily/302 Found/' |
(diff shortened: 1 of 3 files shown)


goodwinos:
Applied this patch, rebuilt/installed and re-ran QA (on the stable system) for the pmwebapi group again - there is consistently one new failure:
```
[8%] 661 6s ... - output mismatch (see 661.out.bad)
180,184d179
<         },
<         {
<             "name": "cpu",
<             "path": "archives/naslog.kernel.cpu.",
<             "is_leaf": 0
```
@fche, can you think of any reason this patch would cause qa/661 to remove archives/naslog.kernel.cpu compared to the qualified output? If that is expected, then I'll just remake 661, but it seems strange that only that one metric is affected.

PR #429, opened 2018-01-26
pmwebd: improve graphite archive-cache performance w.r.t. syscalls

fche:
It was reported that on a large collection of pcp archives, which
included a number of corrupt (0-byte ones), the graphite
metric-enumeration query took too long.  One source of this was
excessive effort on
- frequently retrying opening corrupt archives, and
- fstat'ing all files under -A $DIR

We no longer do either.  Corrupt archives are treated as though they
were fresh at the moment of pmwebd startup, but containing no content.
The -A directory's transitive contents are no longer routinely
fstat()d, only readdir() enumerated.  This costs us the ability to
follow subdirectory symlinks, but it's a pretty big win otherwise.
No QA impact, only performance.

### DIFF
--- a/man/man1/pmwebd.1
+++ b/man/man1/pmwebd.1
@@ -95,6 +95,8 @@ protocols, if possible.
 .TP
 \f3\-A\f1 \f2archdir\f1
 Limit remote access to archives to only those beneath the given directory.
+For performance, symbolic links to other directories may not
+be followed.
 By default, only files beneath the initial
 .B pmwebd
 working directory may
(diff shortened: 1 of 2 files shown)


goodwinos:
ran baseline (master) QA for group 'pmwebapi' and 1090 and 1388 failed. Then built and installed with the patches from this PR and #428, and now 661 and 1042 fail. That's on the same machine, so maybe there is residual pmwebapi QA pollution between test runs or something. Needs more investigation.

goodwinos:
Earlier report mentioning ".. residual pmwebapi QA pollution between test runs" has proven correct - I've applied this patch and run QA on a different system, and it's fine now. So I'll be merging this PR shortly.

goodwinos:
Merged into upstream master.

PR #440, opened 2018-02-08
Add a pmrep configuration to provide breakdown of CPU usage by NUMA node

wcohen:
The grouping of CPUs to nodes (for example processors 0,4,8,12 are in
node 0) may make it more difficult to spot when there is a performance
issue due to NUMA load imbalance.  To make it easier to spot
performance problems due to load imbalance between NUMA nodes the
pmrep pernodecpu configuration provides a per node breakdown of CPU
usage similar to the pmrep sar-u-ALL-P-ALL configuration.


myllynen:
Or perhaps prefix all these NUMA related sets with `numa-` especially if more are planned.

PR #458, opened 2018-03-17
golang: add initial PMAPI support

ryandoyle:
Adds initial golang support for PMAPI.

### DIFF
--- a/src/golang/pmapi/c_glue.c
+++ b/src/golang/pmapi/c_glue.c
@@ -1,3 +1,23 @@
+//Copyright (c) 2018 Ryan Doyle
+//
+//Permission is hereby granted, free of charge, to any person obtaining a copy
+//of this software and associated documentation files (the "Software"), to deal
+//in the Software without restriction, including without limitation the rights
+//to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//copies of the Software, and to permit persons to whom the Software is
+//furnished to do so, subject to the following conditions:
+//
+//The above copyright notice and this permission notice shall be included in all
+//copies or substantial portions of the Software.
+//
+//THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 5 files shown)


ryandoyle:
Hey there - I had this sitting around in my local branch of PCP for a while.

There is still a bit of functionality missing from the PMAPI but there is more than enough to query for metrics/units/etc... I pulled this code from my simplified golang based interface to PCP  https://github.com/ryandoyle/pcpeasygo. Let me know if the licensing is okay (MIT).

natoscott:
@ryandoyle in terms of language, build system, CI, and so on, this seems like it would be a better match for merging with Speed (https://github.com/performancecopilot/speed) rather than core PCP ...?  (Speed also happens to be MIT licensed as well).  Or perhaps as a standalone project building on PCP.

ryandoyle:
Hey @natoscott. Fair enough - it might make including it in projects easier too. How about a separate repository and keep Speed standalone as it is?

What would you like the repository to be? `pcp-go`, `pmapi-go`, something else?

natoscott:
| How about a separate repository and keep Speed standalone as it is?

Sounds fine to me.

| What would you like the repository to be?

I'd recommend a name that suits expansion beyond just PMAPI (so not "pmapi-go") - some of the code will be common for the PMDA API, which might be a follow-on project for you/someone, so best to not lock it into just the PMAPI (IMO).

So, "pcp-go" would be my choice of those two you suggested, or something else entirely - might want to think about when its being promoted and discussed, something catchy / recognisable like "Vector" or "Parfait" really helps.

I recommend we add it as a top-level project under the pcp github org though, to help get it visible - all up to you as the author, just let me know what you decide.

ryandoyle:
Hey @natoscott, I'm happy with `pcp-go` as it's really just go bindings for PCP and not a new product like Vector or Parfait IMO.

Having it on the top-level PCP github org would be good as you've mentioned. I don't have access to do that, if yourself or someone else could that would be great.

Cheers - Ryan

natoscott:
@ryandoyle yep, np - do you have a standalone git repo with this code?  (i.e. not part of a PCP repo with all of the PCP history).  If not, this can be cleanly extracted ... let's chat on slack tomorrow.

ryandoyle:
@natoscott, yep I can extract my commits and push them standalone so thats all good.

natoscott:
@ryandoyle ok mate, once you have a clean pcp-go repo, send me a pointer and I'll set it all up with you as owner, permissions to run CI, etc.

ryandoyle:
Hey @natoscott, I've setup https://github.com/ryandoyle/pcp-go as what I'd push to the one under the pcp org. Let me know if it needs adjustments etc..

ryandoyle:
Now at https://github.com/performancecopilot/pcp-go

PR #464, opened 2018-03-23
qa: add test 1116 for BCC PMDA biotop tool

andihit:
this test writes a testfile to `/var/tmp/pcp_qa_1116`, because a block device is required for the bcc tool (`/tmp` is tmpfs)
is this ok, or should I use another location for the testfile?

### DIFF
--- a/qa/1116
+++ b/qa/1116
@@ -1,0 +1,113 @@
+#!/bin/sh
+# PCP QA Test No. 1116
+# Exercise the BCC PMDA - install, remove and values.
+#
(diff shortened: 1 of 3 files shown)


natoscott:
|  is this ok, or should I use another location for the testfile?

Thats OK - but best to access it via $PCP_TMPFILE_DIR (from /etc/pcp.conf) - I've made that tweak, and also removed a bash-ism (assuming that /bin/sh == /bin/bash) in the test in one spot (please double check that for me?  Thanks).

andihit:
looks good, thanks for merging!

PR #474, opened 2018-03-31
GH#439: add low-level-discovery support to pcp2zabbix

fche:
Add new option --ldd = -l, to set a mode whereby pcp2zabbix
sends zabbix lld metrics for each pcp metric that has an indom.
This lld metric allows zabbix admins to configure templates
that grow automatically with pcp instance domains.  doc, qa++.

### DIFF
--- a/qa/1068
+++ b/qa/1068
@@ -64,14 +64,16 @@ EOF
 
 cat $tmp.config >>$seq.full
 
-socat tcp-listen:$zabbix_port,reuseaddr - >$tmp.socat.out 2>$tmp.socat.err &
+# start a longer-lived copy of socat we can use for consecutive tests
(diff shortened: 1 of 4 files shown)


goodwinos:
Merged into upstream master, thanks Frank.

Note there is an issue on F27 where the module directory has been changed to /usr/lib64/modules by default. On F27 (at least), that directory doesn't exist, and so the zabbix_agentd config would need to be set up to specify the path to /usr/lib64/zabbix/agent/zbxpcp.so (or it's symlink /usr/lib64/zabbix/modules/zbxpcp.so).

And this has apparently been reworked upstream, so additional updates will be needed for PCP. We're looking into it, will probably need to make the config autoconf configured and platform specific.

fche:
(pcp2zabbix has no connection with the zbxpcp agent code)

PR #482, opened 2018-04-16
Committing code, archives and qa for process state tracking

nkshirsagar:
This feature described at http://tasks.pcp.io/b/cbEf5fxGPp8BrbGYS/roadmap/a6shtF8jBdQ3MLPXn and http://tasks.pcp.io/b/cbEf5fxGPp8BrbGYS/roadmap/8ZW4kzL57tkgZzwuj


goodwinos:
Hi Nikhil, I have merged your pcp-pidstat updates into my local tree and tested on F26 and F27, which use python3.6 by default. It all works with python2 (e.g. on RHEL6 and RHEL7), but fails with python3 as follows:

```
kilcunda:mgoodwin@~/src/pcp-clean[master]$ python3 /usr/libexec/pcp/bin/pcp-pidstat -a qa/archives/pcp-pidstat-process-states -B all
Linux  3.10.0-693.el7.x86_64  (unused)  04/10/18  x86_64    (8 CPU)
  File "/usr/libexec/pcp/bin/pcp-pidstat", line 821, in <module>
    sts = manager.run()
  File "/usr/lib64/python3.6/site-packages/pcp/pmcc.py", line 665, in run
    self._printer.report(self)
  File "/usr/libexec/pcp/bin/pcp-pidstat", line 798, in report
    report.print_report(timestamp, header_indentation, value_indentation)
  File "/usr/libexec/pcp/bin/pcp-pidstat", line 402, in print_report
    if process_state_info.has_key(key):
AttributeError: 'dict' object has no attribute 'has_key'
kilcunda:mgoodwin@~/src/pcp-clean[master]$ pmpython -V
Python 3.6.5

```
If you need assistance forward porting to python3 platforms, let me know. A follow-up commit to support python3 would be best, rather than re-doing the whole patch. You'd just need to change
 
`if process_state_info.has_key(key):`

to something like:

`if key in process_state_info:`

and then be sure to test on both python2 and python3

-- Regards

goodwinos:
OK, it's working with python3 now.

However, the new QA test qa/1396 is not passing anywhere except India timezone. This is because pcp-pidstat reports in the local timezone by default. The -z option has to be used to report in the timezone of  the test archive (rathe than the local timezone where the test is being run), so it will have reproducible qualified output when the QA test is run in other timezones. I've changed 1396 so it's run as `pcp -z pidstat -a ....` and pushed your changes along with my change up to https://github.com/performancecopilot/pcp.git so you can pull it down from there. This has not been merged to master yet, but getting close.

Secondly, the metrics check is failing for all the other QA tests in the pidstat group (064, 1078 and 1040). This is because the pre-existing archives used by those tests do not have all the required metrics, e.g.

```
# ./check -l 1078
... [trimmed]
< 07:35:08         1000	32357	140	faked-sysv
< 07:35:08         1000	32364	140	rules
---
> Error: not all required metrics are available
> Missing ['proc.psinfo.sname', 'proc.psinfo.start_time', 'proc.psinfo.wchan_s']
```

The archive used by test qa/1078 doesn't have those three metrics in it. This affects tests 064, 1078, and 1140 in the same way. Either the archives for those tests will have to be re-made, or pcp-pidstat will have to be smarter about checking for needed metrics, i.e. make it conditional on which command line options are given (only the -B option actually needs those three metrics AFAICT).

nkshirsagar:
merged with https://github.com/goodwinos/pcp.git and submitted patches for python3 and new metric table for -B (ensuring old qa tests for pidstat dont break due to missing metrics in their collected archives)

PR #486, opened 2018-05-16
formatting strings: %I64i used instead of  %lld

Andrii-hotfix:
formatting strings: %I64i used instead of  %lld


Andrii-hotfix:
@kmcdonell  yes this is for Windows build. 
Thank you for your advice. I'll make suggested changes today!

Andrii-hotfix:
@kmcdonell thanks for the tips!
You have mentioned line 1371 at load.c (libpcp_web) but i guess you have meant line 1371 at query.c (libpcp_web). I've also added cast (__int64_t) at line 1366. Is that ok?

PR #487, opened 2018-05-16
pmdabcc: add runqlat tool, don't create empty instances (histogram buckets), QA tests for biolatency and runqlat and QA refactorings

andihit:


myllynen:
LGTM, thanks!

PR #488, opened 2018-05-21
pmdabcc: label support

andihit:


myllynen:
Also, please make sure pylint runs cleanly (methods need comments at least) and prefer to keep coding style consistent (%s instead of format() and "return {}" instead "return '{}'" if possible). Thanks.

andihit:
sry about the pylint/PEP, I'll add a check to travis

andihit:
I've updated the branch to reflect the changes
I'll create a separate PR for including pylint into travis as I have to update a few other files as well to not break the build immediately

myllynen:
LGTM, thanks! (As pointed out, the methods need to return strings not dicts.)

PR #489, opened 2018-05-21
build: include python3-bpfcc as a dependency for debian

andihit:


natoscott:
@kenj I think its safe to treat this the same was as PMDA_PERFEVENT and other similar cases in debian/fixcontrol.master...?  (i.e. conditionally injected into debian/control)

cheers.

kmcdonell:
Sorry, clicked close instead of merge ... will merge when CI is done.

PR #491, opened 2018-05-23
pmdabcc: ext4dist

andihit:


andihit:
@natoscott could you take a look at the makefile updates (updated makefile in `modules`, new makefile in `fs` dir) of the last commit? I'm not 100% sure if that's the PCP way to do it

natoscott:
@andihit those makefile changes look spot on.

PR #493, opened 2018-05-31
pmdabcc: xfsdist and zfsdist

andihit:


andihit:
I'll schedule the pid filtering option for the next release of PCP.

About the bitshift: there is a difference ([see bpf_get_current_pid_tgid docs](https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#4-bpf_get_current_pid_tgid)), but I'm also not sure what to use, as it differs from BCC tool to tool. Probably we should use `FILTER_PID` everywhere and add an option to the `apply_pid_filter` to use either the process id or thread group id (depending on the BCC tool).

PR #495, opened 2018-05-31
pmdabcc: rename runqlat metric name

andihit:
as `bcc.kernel.all.runnable` is inconsistent with the current `kernel.all.runnable` metric from PCP, I renamed it.

@natoscott do you have any better suggestions for the name?
The tool 
> measures the time a task spends waiting on a run queue for a turn on-CPU, and shows this time as a histogram. This time should be small, but a task may need to wait its turn due to CPU load.

(https://github.com/iovisor/bcc/blob/master/tools/runqlat.py)


natoscott:
@andihit that name looks good to me.

PR #501, opened 2018-06-04
Correct the semantics for proc_fs_nfsd.th_cnt and proc_fs_nfsd.pool_cnt

wcohen:
The proc_fs_nfsd.th_cnt and proc_fs_nfsd.pool_cnt were incorrectly
configured with counter semantics.  Both of the metrics should have
instanteneous semantics.


natoscott:
Thanks Will.  Changes in metric metadata need to be accompanied by pmlogconf(1) rules in case we have users recording these metrics with pmlogger(1) already.  For reference, see Mark's commit 362a6b2206f85.

If any help is needed there, please lemme know.

PR #505, opened 2018-06-11
pmdabcc: tcplife: track sessions independently, add more metrics + QA

andihit:
Updated tcplife to track each TCP session independently of the process and export all metrics visible in the BCC tcplife tool. The session_count configuration option controls how many TCP sessions should be stored in the cache.

@natoscott all instance values will never change, as they show information from already closed TCP sessions, therefore PM_SEM_DISCRETE is the correct semantic for all metric values I assume?

Screenshot of the corresponding vector widget:
vector_tcplife
(I'll rename RADDR/RPORT to DADDR/DPORT in the vector widget to keep it consistent with the BPF code and manpage of tcplife)

### DIFF
--- a/qa/1155
+++ b/qa/1155
@@ -1,0 +1,100 @@
+#!/bin/sh
+# PCP QA Test No. 1155
+# Exercise the BCC PMDA tcplife module - install, remove and values.
+#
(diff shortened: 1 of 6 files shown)


natoscott:
@andihit yep, I think it makes sense to use discrete here.

PR #508, opened 2018-06-12
Windows native build changes

Andrii-hotfix:
Changes required by native build process on Windows


natoscott:
@kmcdonell I have been chatting to @Andrii-hotfix & working through these issues on #windows on slack - just FYI.

The git installer Windows SDK creates a /proc for itself for some reason :( - seems to be a new development, and breaks the build - not sure we can sensibly configure this anymore as a result.

I believe the whitespace cleanups will be automated git tooling (probably on-by-default in git client on Windows) FWIW.

PR #510, opened 2018-06-15
pmdabcc: execsnoop

andihit:
execsnoop bcc tool + QA + some refactorings

### DIFF
--- a/qa/1155
+++ b/qa/1155
@@ -24,21 +24,6 @@ _install_filter()
     #end
 }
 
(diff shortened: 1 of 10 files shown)


andihit:
I've increased the timeout in `_pmdabcc_wait_for_value` to 30s (from 10s), works now on travis
but I'm not 100% happy with that solution, 10s should be more than enough for the first value to arrive from the poller thread

PR #511, opened 2018-06-20
pmdabcc: tcpretrans module

andihit:
add the tcpretrans module to the BCC PMDA
including 2 QA tests for the tracing mode and counting mode of the module


andihit:
seems like the test for execsnoop isn't very stable :/
the 2 new tests for tcpretrans ran successfully

PR #512, opened 2018-06-20
Adding support for -Zz flag. 

nagarevaibhav:
Adding support for -Zz flag. The -Z option changes the timezone to timezone in the format of the environment variable TZ. The -z option changes the reporting timezone to the local timezone as the host that is the source of the performance  metrics, as identified via either the -h or -a options.


nkshirsagar:
I guess you need to change the iostat reference in the man page to pidstat ?

nagarevaibhav:
Hello @nkshirsagar  , I have changed the pcp-iostat reference in the man page to pcp-pidstat.
Thank you

nkshirsagar:
looks good to me

PR #525, opened 2018-07-17
pmdabcc: profile module and store callback, python: release GIL

andihit:
The profile module records stack traces at a specific interval. It should be used in combination with Vector, which will display them as flame graphs:
profile

This PR also includes a commit which releases the GIL while the PMDA is waiting for new PDUs, otherwise the GIL would lock out background threads.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -136,7 +136,11 @@ disable=print-statement,
         no-else-return,
         too-many-instance-attributes,
         too-many-locals,
-        too-many-public-methods
+        too-many-public-methods,
+        too-many-branches,
+        invalid-name,
+        too-many-statements,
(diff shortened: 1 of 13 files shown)


andihit:
don't merge yet, @goodwinos found a bug with the GIL commit: when pmcd is sent a SIGTERM, all the python PMDAs end up exiting with a segfault

fche:
> when pmcd is sent a SIGTERM, all the python PMDAs end up exiting with a segfault

That's probably because in the case of a SIGTERM (or other error), `__pmdaMainPDU` returns an error rc before giving the "check" callback a call.  Via `pmdaMain`, this gives control back to python without having restored the `thread_state`.

PR #528, opened 2018-07-18
pmdabcc: move module options description to bcc.conf, add short description of each module

andihit:
Users shouldn't have to look into the source code of each module to get a description of the configuration options - therefore I moved them to `bcc.conf` (and grouped the modules + added a short description to each module as suggested by @myllynen)


andihit:
this commit also includes the description for the new profile tool (#525) - for the merge conflict please use `bcc.conf` from this commit and ignore the changes to `bcc.conf` in #525

PR #531, opened 2018-07-25
pcp-pidstat: Adding support for -l flag.

nagarevaibhav:
"Adding support for -l flag. The -l flag/option displays the process command name and all its arguments which works on live system and also on archives. In addition to this -l flag can also be used with other flags and multiple usage of -l is avoided by displaying incorrect message"


nkshirsagar:
<mgoodwin> mgoodwin@~/src/pcp/src/pcp/pidstat[nagarevaibhav-master]$ ./pcp-pidstat.py -l
<mgoodwin> pcp-pidstat.py: invalid option -- 'l'
<mgoodwin> have I missed something here? Doesn't seem to work??
<mgoodwin> the other thing that will need to be done is to measure the logging space overheads of adding the additional metric proc.psinfo.psargs to src/pmlogconf/zeroconf/pidstat
<mgoodwin> has that been measured? Some processes e.g. chrome have very long arguments
<nkshirsa> really? lemme check about it not working
<nkshirsa> ok there's a typo in there
<nkshirsa>         pmapi.pmOptions.__init__(self,"a:s:t:G:IU::p:RrkV?:f:B:l")
<nkshirsa>         pmapi.pmOptions.__init__(self,"a:s:t:G:IU::p:RrkVZ:z?:f:B:")
<nkshirsa> second one seems to override the first one, need to remove that second line not sure how it got in there

nagarevaibhav:
Resubmitting the commit, thus closing this request.
Thank you

PR #539, opened 2018-08-01
pmieconf/rules.c: Do not record build date

bmwiedemann:
Do not record the build date
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good.

The alternative fix would be to allow to override it like in
https://wiki.debian.org/ReproducibleBuilds/TimestampsProposal#C


bmwiedemann:
Taking a closer look, this change seems to actually not be required for our openSUSE packages. The fedora-paste patch seems to address the same thing.

There is another open issue with tutorial .tar.gz creation - will open a different PR for that.

PR #544, opened 2018-08-02
Fix perfevent agent to return correct cpu numbers in lower smt levels

anjutsudhakar:


natoscott:
@jpwhite4 @hkshaw1990 could you guys double check this please?  Looks OK to me.  (not sure if we still need to wrap sysconf(3) anymore for QA after this change, Joe?)

hkshaw1990:
@natoscott will do.

natoscott:
@anjutsudhakar @hkshaw1990 @jpwhite4 OK, merged - thanks.  I'm seeing a little QA fallout - some asserts are now tripping in the test code.  This patch fixes those (which I think is correct due to the use of four CPUs in test data - but am not certain):
https://paste.fedoraproject.org/paste/kDlMrScioBxgPheVRBsipw

However even with that, I'm still seeing the test fail to produce the expected output (8 CPU test machine):
https://paste.fedoraproject.org/paste/WI0jVZuu9jGON6~vAvQOcQ

Could some take a closer look?  Thanks!

hkshaw1990:
Will look into this today. @natoscott

hkshaw1990:
@natoscott the patch https://paste.fedoraproject.org/paste/WI0jVZuu9jGON6~vAvQOcQ is not required. It looks like there is a cpu mismatch issue as we are getting the CPUs from sysconf and from fakefs.
@anjutsudhakar is working on fixing that for the test cases, where fakefs cpu lists should be added.

PR #547, opened 2018-08-06
 Implement the trivial-pmda also as perl daemon 

christianhorn:
The trivial-pmda is currently the most simple pmda implementation, provided in C. Adding a perl variant which provides the same functionality.

### DIFF
--- a/src/pmdas/trivial/GNUmakefile
+++ b/src/pmdas/trivial/GNUmakefile
@@ -35,7 +35,7 @@ include $(BUILDRULES)
 
 install install_pcp:	default
 	$(INSTALL) -m 755 -d $(PMDADIR)
-	$(INSTALL) -m 755 Install Remove $(PMDADIR)
+	$(INSTALL) -m 755 Install Remove pmdatrivial.perl $(PMDADIR)
 	$(INSTALL) -m 644 GNUmakefile.install $(PMDADIR)/Makefile
 	$(INSTALL) -m 644 $(DFILES) root help pmns domain.h $(CFILES) $(PMDADIR)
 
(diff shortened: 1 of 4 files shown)


christianhorn:
pmdatrivial.perl is constantly running as child of pmdaroot, thus I thought 'daemon' would fit.

I see your point regarding the missing Python implementation, but as my python knowledge is limited, I was not successfull in modifying a simple/pdmasimple.python into a pdmatrivial.python.

christianhorn:
> I don't think it's needed to add Aconex or SGI copyright notices here?
Also the perl daemon is a downstripped version of simple/pmdasimple.perl, as per this I kept these parts.

natoscott:
@christianhorn @myllynen couple of small points ...

Many of the PCP docs do tend to describe any PMDA that is not in DSO form as a "daemon", which is perhaps a bit confusing but the code changes here are consistent with what we have in that regard.

I tend to agree it would be neat to have a python version as well - I'll have a crack at this if noone else volunteers.

Finally, we need some regression testing for the new code.  The current C trivial PMDA is tested out by qa/199 - so perl and python versions of this test will be handy.  Again, if noone else volunteers there I'll circle back and knock those out when I merge this.

cheers.

natoscott:
@christianhorn I've added tests, a Python variant and made some small updates to the Perl PMDA too - please double check my changes, thanks!

christianhorn:
Wow, awesome. Wanted to look at the perl test code, when you had already implemented the python part and the 2 tests :)
Rebuilding master, all trivial variants look good to me. I think this will help people getting custom metrics up and running quickly, I will refer to this new code.

PR #552, opened 2018-08-13
qa/perfevent: Fix the QA fallouts

anjutsudhakar:
Fix the qa fallouts happened due to the change in the way cpuids are
populated by retrieve_cpuinfo(), for perfevent pmda.

Signed-off-by: Anju T Sudhakar <anju@linux.vnet.ibm.com>


anjutsudhakar:
Do we still need `wrap_sysconf_retcode` defined in the test cases ? Because we are no more reading the cpulist from sysconf(_SC_NPROCESSORS_ONLN), instead we read it from /sys/devices/syste/cpu directory.

PR #559, opened 2018-09-04
modify lmsensors to perform autodetection of available sensors

christianhorn:
Not sure if this code is acceptable (or can be made acceptable), but let me try. I have implemented monitoring of temperature/fans on some of my systems, and revamped it so it might be usable for others.

- This is revamping the former C code, and using Perl instead. It relies on 'lm_sensors' being installed.
- The code is designed to require minimal user interaction: when the pmda starts, it will detect the available sensors (parsing 'sensors -u' output) and create metrics for them. These are different by hardware model, so one does not know beforehand which metrics become available. I still use file 'pmns' from the former implementation, but the contents do actually not apply.
- I looked at performance. Real used memory is around 10MB, should be ok. I looked at cpu cycles, and did verify the current (lm_sensors based) code with Perl directly parsing /sys directories. The latter one is ~15 to 20% faster, but comes to a downside: some sensors need 'finetuning afterwards': they need i.e. 20deg subtracted or such. lm_sensors is doing this abstraction internally, when doing the parsing directly then I would need to implement this directly.


christianhorn:
After looking more into the tests and reading the log https://api.travis-ci.org/v3/job/424231580/log.txt , I think that somehow the tests were performed on a tree which was not passing all tests at that time.
I think "./check -g pmda.bcc" failed, which I did not touch. Have not seen a separate test for lmsensors.

natoscott:
@christianhorn yes, sorry - those failures were not related to your new code at all.  Marko, Andreas and I had been looking into those in parallel and they're now resolved.

As discussed on IRC the other day, please let me know if you need help with lmsensors PMDA testing - don't worry about those (unrelated) CI failures though.

christianhorn:
Ok, will do a new pull request and see how the testing goes.

PR #564, opened 2018-09-17
modify lmsensors to perform autodetection of available sensors, v2

christianhorn:
Not sure if this code is acceptable (or can be made acceptable), but let me try. I have implemented monitoring of temperature/fans on some of my systems, and revamped it so it might be usable for others.

- This is revamping the former C code, and using Perl instead. It relies on /usr/bin/sensors being installed, usually via package lm_sensors.
- The code is designed to require minimal user interaction: when the PMDA starts, it will detect the available sensors (parsing 'sensors -u' output) and create metrics for them. These are different by hardware model, so one does not know beforehand which metrics become available. I still use file 'pmns' from the former implementation, but the contents do actually not apply.
- I looked at performance. Real used memory is around 10MB, should be ok. I looked at cpu cycles, and did verify the current (lm_sensors based) code with Perl directly parsing /sys directories. The latter one is ~15 to 20% faster, but comes to a downside: some sensors need 'finetuning afterwards': they need i.e. 20deg subtracted or such. lm_sensors is doing this abstraction internally, when doing the parsing directly then I would need to implement this directly.


goodwinos:
Hi Christian,

the code looks fine, and builds and works fine. Couple of things:

1. need to update copyright notices to add your copyright (most likely Red Hat 2018), retaining existing copyright notices

2. do you want to stick with a softdep check at run-time? Otherwise the pcp-pmda-lmsensors package should Require: lm_sensors  in build/rpm/pcp.spec.in (and fedora.spec). If you want to stick with the softdep, then the Install script should probably check if /usr/bin/sensors exists rather than having this check in the PMDA itself (or both is fine too)

3. qa test - there don't appear to be any qa tests for this PMDA at all (which I think pre-dates your new perl version). Let me know if you need some assistance with this bit!

Cheers
-- Mark

christianhorn:
Thanks! Working on a new pull request, additionally to these points the PMDA gets reimplemented in Python.

PR #567, opened 2018-10-03
lmsensors rework, v4

christianhorn:
An implementation of monitoring of temperature/fans.

- This is revamping the former C code, and using Python instead. It relies on /usr/bin/sensors being installed, usually via package lm_sensors. The pmda checks for the binary, and a rpm dependency was added.
- The code is designed to require minimal user interaction: when the PMDA starts, it will detect the available sensors (parsing 'sensors -u' output) and create metrics for them. These are different by hardware model, so one does not know beforehand which metrics become available. I still use file 'pmns' from the former implementation, but the contents do actually not apply.
- Used real memory is around 18MB, should be ok. Directly parsing /sys directories is ~15 to 20% faster, but comes to a downside: some sensors need 'finetuning afterwards': they need i.e. 20deg subtracted or such. lm_sensors is doing this abstraction internally, when doing the parsing directly then I would need to implement this directly in the pmda.


fche:
>  The code is designed to require minimal user interaction: when the PMDA starts, it will detect the available sensors (parsing 'sensors -u' output) and create metrics for them.

excellent

goodwinos:
Merged though not with the original commits (due to commits by root), so I applied as patches and committed with --author="Christian Horn <chorn@redhat.com>"

PR #581, opened 2018-11-25
pcp2elasticsearch: Support es_search_type flag

rgolangh:
Support the **es_search_type** config flag and allow to override the built-in
**pcp-metric**. This is specially useful for projects like ViaQ who maintian
their own types and indexes.


sradco:
Please review @lberk @natoscott

src/pcp2elasticsearch/pcp2elasticsearch.py
```diff
@@ -424,7 +430,7 @@ def write_es(self, timestamp):
                         insts.append({inst_key: name, last_part: value})
 
         try:
-            url = self.es_server + self.es_index + "/pcp-metric"
+            url = self.es_server + self.es_index + self.es_search_type
```
sradco: Need to add "/"
@rgolangh

src/pcp2elasticsearch/pcp2elasticsearch.py
```diff
@@ -253,6 +256,9 @@ def option(self, opt, optarg, _index):
             self.es_index = optarg
         elif opt == 'X':
             self.es_hostid = optarg
+        elif opt == 's':
+            self.es_search_type = optarg
+            print("search type {}".format(optarg))
```
sradco: Need to remove print

src/pcp2elasticsearch/pcp2elasticsearch.py
```diff
@@ -178,6 +180,7 @@ def options(self):
         opts.pmSetLongOption("es-host", 1, "g", "SERVER", "Elasticsearch server (default: " + ES_SERVER + ")")
         opts.pmSetLongOption("es-index", 1, "x", "INDEX", "Elasticsearch index for metric names (default: " + ES_INDEX + ")")
         opts.pmSetLongOption("es-hostid", 1, "X", "HOSTID", "Elasticsearch host-id for measurements")
+        opts.pmSetLongOption("es-search-type", 1, "", "ES_SEARCH_TYPE", "Elasticsearch search type measurements")
```
sradco: missing "for" measurements

rgolangh:
@myllynen What about dropping the short flag and supporting only the long ones? For one, we would run out of flags some time, and two, it is not that it is easy to remember. -p is a weak hint for 'search type'.

myllynen:
I think it's best to have a short option defined as long as there are letters available. If you think `-p` is not intuitive then perhaps use another one, see e.g. `pcp2graphite specific options` in `pcp2graphite`(1) for options others are using.

Whatever short option you pick, you'll need to add it to shell completions to avoid breaking QA 1100.

Thanks.

natoscott:
@rgolangh @sradco @lberk  I've fixed an issue related to short option handling, updated the test case a bit (and updated the expected test output), and merged.  Please double check my changes in commit 504b117d96 - thanks!

PR #582, opened 2018-12-02
pmwebd: tolerate live archives with multi-hour-old volumes

fche:
Prior to this patch, it was observed that in a live (being written-to)
multi-volume archive, pmwebd would stop looking for the last volume#
too early.  The timestamp on that wrong "last" volume is used in a
subsequent freshness heuristic.  Because the actual last volume file
may not be found during the search, pmwebd could be temporarily blind
to newer data available in newer volumes of the same archive,
returning falsely empty results.  This patch improves this volume
enumeration logic to find the last one, via an explicit loop.  This
leaves pmwebd with a max-60-second blind spot for fresh values.

qa/1389 tests with a pmlogger with HUP-triggered volume switching, and
an explicit /bin/touch to simulate longer-running previous volumes.

### DIFF
--- a/qa/1389
+++ b/qa/1389
@@ -1,0 +1,107 @@
+#! /bin/sh
+# PCP QA Test No. 1389
+# checks pmwebd graphite support for multi-volume archives
+#
+# Copyright (c) 2018 Red Hat.
+#
+seq=`basename $0`
+echo "QA output created by $seq"
+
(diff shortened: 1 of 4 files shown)


goodwinos:
Merged with a small tweak to qa/1389, which runs for ~ 2m10s (a bit long, but it passes and does the job)

PR #584, opened 2018-12-05
pmda: perfevent: Add support for raw perf events

sandip4n:
This adds support for using perf events which might not be exposed via standard mechanisms such as `sysfs` or `libpfm`. The alternative way to access these hardware counters is by using the architecture
dependent raw event codes directly.

Raw events can be specified in the perfevent configuration using the following syntax:
```
  RAW:[EVENT_NAME] rawcode=[RAW_EVENT_CODE]
```

E.g.
```
  $ tail -n 1 /var/lib/pcp/pmdas/perfevent/perfevent.conf
  RAW:foo rawcode=0x1e

  $ pminfo | grep foo
  perfevent.hwcounters.RAW_foo.dutycycle
  perfevent.hwcounters.RAW_foo.value
```

### DIFF
--- a/src/pmdas/perfevent/configparser.h
+++ b/src/pmdas/perfevent/configparser.h
@@ -37,6 +37,7 @@ typedef struct pmcsetting {
     double scale;    /* Currently, only used by derived events */
     int need_perf_scale;  /* Currently, only used by derived events */
     int chip;	/* Currently, only used by hv_24x7 dynamic events */
+    unsigned long rawcode;  /* Currently, only used by raw events */
     struct pmcsetting *next;
 } pmcsetting_t; 
 
--- a/src/pmdas/perfevent/configparser.l
+++ b/src/pmdas/perfevent/configparser.l
@@ -272,7 +272,7 @@ static void add_pmcsetting_name(configuration_t *config, char *name)
 
     entry = &config->configArr[config->nConfigEntries-1];
 
-    newpmcsetting = malloc(sizeof *newpmcsetting);
+    newpmcsetting = calloc(1, sizeof *newpmcsetting);
     newpmcsetting->name = strdup(name);
     newpmcsetting->cpuConfig = CPUCONFIG_EACH_CPU;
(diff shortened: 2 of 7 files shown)


src/pmdas/perfevent/configparser.l
```diff
@@ -270,7 +270,7 @@ static void add_pmcsetting_name(configuration_t *config, char *name)
 
     entry = &config->configArr[config->nConfigEntries-1];
 
-    newpmcsetting = malloc(sizeof *newpmcsetting);
+    newpmcsetting = calloc(sizeof *newpmcsetting);
```
natoscott: This is not going to compile as-is.  :)

src/pmdas/perfevent/configparser.h
```diff
@@ -36,6 +36,7 @@ typedef struct pmcsetting {
     int cpuConfig;
     double scale;    /* Currently, only used by derived events */
     int need_perf_scale;  /* Currently, only used by derived events */
+    unsigned long rawcode;  /* Currently, only used by raw events */
```
natoscott: This is going to patch-conflict with the recent change from @anjutsudhakar I think - could you pull those changes in (merging upstream now) and update this PR please?

src/pmdas/perfevent/configparser.l
sandip4n: Dang! Last minute change.

src/pmdas/perfevent/configparser.h
sandip4n: Sure.

sandip4n:
Thanks @anjutsudhakar

natoscott:
@sandip4n could you add a regression test for this?  (along the lines @anjutsudhakar did in commit a1250484ae9 ?)

sandip4n:
@natoscott Sure

sandip4n:
@natoscott Rebased on top of current master branch and added test cases.

Issue #419, opened 2018-01-16
libpcp-opened file descriptors should be O_CLOEXEC


fche:
A multithreaded program that deals with pcp archives and/or sockets, and forks child processes, may unintentionally pass open file descriptors to the children.  These somewhat harmless (in that the child process will leave them alone, if not buggy), except that they'd hold references to files/sockets that can't be completely released for perhaps an extended period.  It would be better not to pass them in the first place.

One possible cure would be to tweak the child process forking step in all programs that exec, so as to manually close all file descriptors between the fork() and exec() ... except that there aren't really good & simple & portable ways of doing that.

Another possible cure would be for libpcp to call fcntl(FD_CLOEXEC) on its file descriptors.  For files on modern linux, glibc's fopen(3) offers an "e" character to do the same fcntl under the covers.

fche:
pmmgr patch that works around problem
[patch.txt](https://github.com/performancecopilot/pcp/files/1637431/patch.txt)

natoscott:
No further work is planned here.

PR #591, opened 2018-12-17
pmmgr: workaround for missing libpcp O_CLOEXEC, issues/419

fche:
Manually close fd's 3..sysconf(_SC_OPEN_MAX) before exec'ing
subordinate daemon processes.

issue #419


natoscott:
This PR lacks QA and as described is more workaround than fix.  Longer term we'll be replacing the standalone pmmgr daemon with improvements to the PCP systemd integration, particularly for pmie and pmlogger farms, so closing this one out now as no further work is planned in this area.

PR #593, opened 2018-12-17
pmwebd: add -r NUM option to retain open archive contexts

fche:
This option permits pmwebd to avoid incurring potentially long libpcp
delays when opening archive contexts, by keeping some number of them
open.  This is done on an LRU basis, and governed by a new "-r NUM"
option.  This makes a huge difference on archives with O(4GB) .meta
files, such as reported on RHBZ1619708.

### DIFF
--- a/src/pmwebd/main.cxx
+++ b/src/pmwebd/main.cxx
@@ -1,7 +1,7 @@
 /*
  * JSON web bridge for PMAPI.
  *
- * Copyright (c) 2011-2017 Red Hat.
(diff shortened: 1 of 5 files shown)


natoscott:
There's been no QA provided exercising the new code in the months since this pull request was created.  
pmwebd is also deprecated and in the process of being replaced by pmproxy - it will be removed in a future release.  Since there are no further updates to pmwebd planned, I'm closing this out now.

Issue #177, opened 2016-10-26
richer systemd support (Bugzilla #1096)


pcpemail:
On 2014-12-05 09:52:58 -0600, Frank Ch. Eigler wrote:

> Building on the basic oneshot/ExecStart level of systemd support,
> we could go a little farther.  For example:
> - have the daemons use stderr/syslog to record trace output into the journal
> - invoke the daemons directly rather than through the sysvinit scripts
> - use systemd timer units instead of cron jobs
> - use unit templates for hard-configured multi-instance daemons (pmlogger, pmie)
> - socket activation for pmcd, pmproxy, pmwebd
> - add extras like Restart=restart-always for key daemons

On 2015-01-28 13:56:43 -0600, Frank Ch. Eigler wrote:

> Also related: the way in which systemctl is invoked is a bit
> cargo-culty and incomplete.
> 
> As per https://bugzilla.redhat.com/show_bug.cgi?id=1152535 ,
> we need some 'systemctl daemon-reload' in %post and %postun.
> We don't need all the '--no-reload's.  See e.g. the fedora
> openssh.spec for more modern invocation runes.

fche:
It has been reported that the pmcd sysvinit script breaks if invoked on a host running containers with other pmcd processes running in them, probably because of its /bin/ps based searching for related processes.  In a proper systemd world, these scripts would not need to resort to ps-based hacks at all, and therefore would be insensitive to non-system pcp processes being run by ordinary users or within containers.

myllynen:
FWIW, when testing something locally I created a systemd unit file for `pmcd`:

```
[Service]
Type=forking
Restart=always
EnvironmentFile=/etc/pcp.conf
EnvironmentFile=-/etc/sysconfig/pmcd
WorkingDirectory=/var/log/pcp/pmcd
ExecStart=/usr/libexec/pcp/bin/pmcd
KillMode=mixed
TimeoutSec=30
```

This seemed to work well and might be a good starting point if someone wants to work on this area. The one feature missing is parsing `pmcd` options in a backwards-compatible manner from `/etc/pcp/pmcd/pmcd.options`.

myllynen:
I experimented a bit and came up with the following scheme for retaining backword-compatible handling of `pmcd.options`. There might be some corners to improve and this is only for `pmcd` (`pmlogger` etc would need their own counterparts) but perhaps this could be a starting point for someone trying to finish this.

[...]

PR #599, opened 2019-01-07
offer systemd timers as an alternative to crontab

ddiss:
This patchset converts (non-docker) pmie and pmlogger crontab entries to systemd timer + service files, which are installed when ENABLE_SYSTEMD is configured.

### DIFF
--- a/qa/255
+++ b/qa/255
@@ -15,7 +15,7 @@ echo "QA output created by $seq"
 . ./common.filter
 . ./common.check
 
-_check_cron
+_check_job_scheduler
 [ -d $PCP_PMDAS_DIR/simple ] || _notrun "simple PMDA directory is not installed"
 
 rm -f $seq.out
@@ -138,11 +138,8 @@ _cleanup()
 	echo '=== pmcd log' >>$seq.full
 	cat $tmp.log >>$seq.full
     fi
-    if [ -f $tmp.cron ]
-    then
-	echo "=== Restoring crontab ==="
-    	_restore_cron $tmp.cron $sudo
-    fi
+    echo "=== Restoring crontab ==="
+    _restore_job_scheduler $tmp.cron $tmp.systemd $sudo
     [ -f $PCP_PMDAS_DIR/simple/simple.conf.$seq ] && $sudo mv $PCP_PMDAS_DIR/simple/simple.conf.$seq $PCP_PMDAS_DIR/simple/simple.conf
 
     echo "=== /tmp files" >>$seq.full
@@ -154,10 +151,10 @@ _cleanup()
     _wait_for_pmlogger
 }
 
-# Removing cron entries that may collide
+# Removing systemd timers or cron entries that may collide
 #
 echo "=== Removing potential cron conflicts ==="
-_remove_cron $tmp.cron $sudo
+_remove_job_scheduler $tmp.cron $tmp.systemd $sudo
 
 # Install broken pmda namespace
 #
--- a/qa/common.check
+++ b/qa/common.check
@@ -1617,8 +1617,8 @@ _remove_cron()
     then
 	:
     else
-	# error, make sure the backup is empty so no changes are made
-	echo >$rc_backup
+	# error, remove the backup file so no changes are made
+	rm -f $rc_backup
     fi
 
     if [ -s $rc_backup ]
--- a/qa/common.check
+++ b/qa/common.check
@@ -1600,30 +1600,46 @@ EOF
   fi
 }
 
-# comment pmlogger_check and pmsnap entries in the crontab file
-# (also cron.pmcheck and cron.pmsnap entries for backwards compatibility)
-# Usage: _remove_cron backup sudo
+# disable systemd timers and / or comment pmlogger_check and pmsnap entries in
+# the crontab file (also cron.pmcheck and cron.pmsnap entries for backwards
+# compatibility)
+# Usage: _remove_job_scheduler cron_backup systemd_state sudo
 #
-# backup - where to keep the old crontab file
+# cron_backup - where to keep the old crontab file
+# systemd_state - where to keep systemd timer state
 # sudo - location of sudo
 #
-_remove_cron()
+_remove_job_scheduler()
 {
-    rc_backup=${1:-crontab}
-    rc_sudo=${2:-sudo}
+    rc_cron_backup=${1:-crontab}
+    rc_systemd_state=${2:-systemd.state}
+    rc_sudo=${3:-sudo}
 
-    $rc_sudo rm -f $rc_backup
-    if $rc_sudo crontab -l 2>/dev/null >$rc_backup
+    $rc_sudo rm -f $rc_cron_backup $rc_systemd_state
+
+    if systemctl cat pmie_daily.timer >/dev/null 2>&1; then
+	for i in pmie_daily.timer pmie_check.timer pmlogger_daily.timer \
+		pmlogger_daily_poll.timer pmlogger_daily_check.timer \
+		pmlogger_daily_report.timer pmlogger_daily_report_poll.timer; do
+	    $rc_sudo systemctl is-active "$i" > /dev/null || continue
+	    $rc_sudo systemctl stop $i >/dev/null
+	    echo "$i" >> $rc_systemd_state
+	done
+    fi
+
+    which crontab >/dev/null 2>&1 || return	# systemd timers only
+
+    if $rc_sudo crontab -l 2>/dev/null >$rc_cron_backup
     then
 	:
     else
 	# error, remove the backup file so no changes are made
-	rm -f $rc_backup
+	rm -f $rc_cron_backup
     fi
 
-    if [ -s $rc_backup ]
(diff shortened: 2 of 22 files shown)


fche:
great - see also issue #177

ddiss:
Ping, any feedback on this one?

natoscott:
> Ping, any feedback on this one?

Hi Dave,

No issues from me, still recovering from end-of-year backlog a bit ... but, what is the rationale for this change OOC? - is cron going away?  or are there situations cron is unavailable?  (containers based on systemd?)  or does this mechanism provide other functionality of use in the PCP context?

There's a small issue in that qa/common.check provides (test 255) an ability to stop cronjobs interfering with its work ... the equivalent would need to be implemented for this systemd mechanism.

Has the upgrade path been thought through?  What of the situation where people have modified the PCP cronjobs to add their own customisations?  (e.g. additional command options, different log compression, and so on).

How would someone go about adding / changing the options used in these systemd files?  (e.g. the "-X xz -x 3" part)  And if that involves editing them directly, is there package manager support for not overwriting those changes, like there is for the /etc cron files?  (an alternative might be to move the explicit use of "-X xz -x 3" into an /etc/pcp file set instead of hard-coding them there)

cheers.

ddiss:
Hi Nathan,

> No issues from me, still recovering from end-of-year backlog a bit ... but, what is the rationale for this change OOC? - is cron going away? or are there situations cron is unavailable? (containers based on systemd?) or does this mechanism provide other functionality of use in the PCP context?

We're trying to move towards using systemd timers exclusively for system packages in future SUSE releases.

> There's a small issue in that qa/common.check provides (test 255) an ability to stop cronjobs interfering with its work ... the equivalent would need to be implemented for this systemd mechanism.

Ah okay, I'll take a look at that.

> Has the upgrade path been thought through? What of the situation where people have modified the PCP cronjobs to add their own customisations? (e.g. additional command options, different log compression, and so on).

This patchset only covers new installations. I guess we could add some rpm script magic to handle cron->systemd timer upgrades. I'll have a think about it.

> How would someone go about adding / changing the options used in these systemd files? (e.g. the "-X xz -x 3" part) And if that involves editing them directly, is there package manager support for not overwriting those changes, like there is for the /etc cron files? (an alternative might be to move the explicit use of "-X xz -x 3" into an /etc/pcp file set instead of hard-coding them there)

If we want these parameters to be user tunable, then using a systemd EnvironmentFile, like your suggested alternative, would make sense.

Thanks for the feedback!

kmcdonell:
Hi @ddiss ... I have basically the same questions as Nathan
- what happens after an upgrade if both cron and systemd think they are driving?
- leaving the religious issues aside, systemd has proven to be a total pain for QA and by default we avoid it whenever possible ... we'd need a big hammer to stop systemd doing async "good" things, like checking and restarting daemons or pmlogger, while QA is running

myllynen:
> systemd has proven to be a total pain for QA and by default we avoid it whenever possible ... we'd need a big hammer to stop systemd doing async "good" things

I think we should acknowledge that PCP's systemd support is not complete and may play a part here. For better or worse, a systemd developer described one of the PCP startup scripts as "horror" (in https://bugzilla.redhat.com/show_bug.cgi?id=1650999) which I suppose was meant to compare the current PCP approach for using systemd to full systemd support implemented by most of the other services running on recent Linux distributions (which does away with (most) custom shell scripts).

Some aspects to consider around PCP / systemd:

1. My Fedora 29 VMs would otherwise shutdown in ~2 seconds but with latest git master with `redis` and `pmproxy -t` in use (for `pmseries`), stopping `pmproxy` takes up to 18 (!) seconds so there's a significant delay caused by PCP init script for system shutdown
1. When running QA, I see often PCP processes being started outside of systemd realm after QA runs (meaning that the PCP services are running but `systemctl status` shows them being stopped)
1. The needed steps to properly support systemd in PCP are AFAICS described in https://github.com/performancecopilot/pcp/issues/177#issuecomment-452209191

Thanks.

ddiss:
Hi @kmcdonell 

> what happens after an upgrade if both cron and systemd think they are driving?

It looks like we should be able to use 
ConditionPathExists=!@CRONTAB_PATH@
ConditionPathExists=!@CRONTAB_DAILY_REPORT_PATH@
...
in the timer service files to avoid this condition. Would you consider this appropriate protection?
If so, I'll resubmit with these changes.

> leaving the religious issues aside, systemd has proven to be a total pain for QA and by default we avoid it whenever possible ... we'd need a big hammer to stop systemd doing async "good" things, like checking and restarting daemons or pmlogger, while QA is running

As discussed with Nathan, I'll look into having qa/common.check handle systemd timers when available.

ddiss:
> It looks like we should be able to use
> ConditionPathExists=!@CRONTAB_PATH@
> ConditionPathExists=!@CRONTAB_DAILY_REPORT_PATH@
> ...
> in the timer service files to avoid this condition.

I did some testing with this - the service is silently skipped rather than logged as a failure, but otherwise it works well.

ddiss:
This new patchset includes the systemd service !ConditionPathExists cron guards and adds pmlogger_/pmie_timers sysconfig EnvironmentFile templates, which can be used to override the default service parameters.
I haven't gotten to the test suite changes yet, so this should probably be considered an RFC for now.

ddiss:
It probably also makes sense to squash these changes down before merge. I haven't done that yet to make it a little easier for reviewers.

natoscott:
> For better or worse, a systemd developer described one of the PCP startup scripts as "horror"

Childish name calling from either side is more likely to hinder than help improve the situation, so please don't repeat such things here - let's stick to facts and discussion of actual patches.

>     [...] stopping `pmproxy` takes up to 18 (!) seconds so there's a significant delay caused by PCP init script for system shutdown

This one is a bug in the new pmproxy code, being worked on (amongst many other things) - dropping into a debugger on pmproxy suggests it is unrelated to the init scripts, and it appears to be more of an async I/O completion issue.

cheers.

ddiss:
@natoscott skipping 255 is most likely not what you had in mind, but I'll leave it here until I have a better idea of what _remove_cron() and _restore_cron() are actually used for in this test.

natoscott:
@ddiss yeah, not that - the equivalent functionality should be achievable in a similar way to the existing cron code - i.e. moving aside the systemd timer files while the test is running (via _remove_cron), and replacing them at the end (via _restore_cron)?  The aim is to prevent interference from the system scripts while the test runs as that would cause the test to fail.

ddiss:
@natoscott I could disable the individual timers, which should achieve something like that.

ddiss:
This new round drops the _check_cron() change and updates test 255 to disable systemd timers throughout the test.

ddiss:
@natoscott this new round attempts to address your proposals. I added an extra systemd.state file so that backup / restore can be handled independently for systemd timers and crontab.

ddiss:
another update with the generated pmlogger/pmie service files added to .gitignore.

natoscott:
@ddiss I'm seeing some fallout from this change.  Upgrading an existing systemd-based system to current PCP via rpm packages results in no cron scripts, no active timers but primary pmlogger and/or pmie service still running.  This means no daily rollover, no daily reports, no log compression, no log culling, and enospc eventually ensues ... generally, badness.

I think we're missing dependencies between the timers and the associated pmlogger/pmie services.  I've taken a stab at adding this based on some systemd docs reading, but would appreciate a closer eye over these changes (will push after a little more testing here.) - thanks!

ddiss:
@natoscott Sorry for the headache - enablement of systemd services is generally handled by the package post-install scripts. SUSE have a few distro specific spec file macros for this, so I wasn't too sure what to do for Fedora, etc. Proper dependencies between the timers and core services would definitely be nice to have. I'm happy to take a look at whatever changes you have staged.

natoscott:
@ddiss this seems to do the right thing (commit 9947fbf6) - no additional spec file changes needed, AFAICT?  (can you describe the SuSE changes a little more so we can audit build/rpm/fedora.spec for missing magic?) - thanks!

ddiss:
> no additional spec file changes needed, AFAICT?

Thanks Nathan. This looks good and should avoid any need to touch the timer services in postinst :)

natoscott:
@ddiss thanks Dave - I think I've overlooked deps for the new services added to replace cron scripts (non-timers) - will update those too.

I also noticed that all other timers seem to use a convention of hyphens rather than underscores in their names.  I wonder if we should follow suit?  Now is probably the best time to decide as these have not been released yet.

ddiss:
> I also noticed that all other timers seem to use a convention of hyphens rather than underscores in their names. I wonder if we should follow suit?

I'm fine with either. I guess it makes sense to use the common convention.

> Now is probably the best time to decide as these have not been released yet.

Agreed. Will you submit with the renames, or would you like me to?

natoscott:
> Agreed. Will you submit with the renames, or would you like me to?

If you could, that'd be great Dave - I'm keen to focus on other QA and docs efforts for the 4.3.1 release later this week.

PR #609, opened 2019-01-31
pmns: don't install .NeedRebuild on Linux

ddiss:
The existing filter already covers RedHat, Debian and SUSE distros, but
depends on os-release file based distro detection.
Change the .NeedRebuild install check to cover all Linux build targets,
so that os-release distro detection isn't necessary.

Signed-off-by: David Disseldorp <ddiss@suse.de>

### DIFF
--- a/src/pmns/GNUmakefile
+++ b/src/pmns/GNUmakefile
@@ -60,7 +60,7 @@ install:	default
 	$(INSTALL) -m 755 Rebuild $(PMNS_VAR_DIR)/Rebuild
 	$(INSTALL) -m 755 Make.stdpmid $(PMNS_VAR_DIR)/Make.stdpmid
 	$(INSTALL) -m 644 $(STDPMID) $(PMNS_VAR_DIR)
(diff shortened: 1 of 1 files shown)


fche:
I wonder why this `.NeedRebuild` business is distro-specific at all.

natoscott:
@ddiss this change has the side-effect of breaking any distribution which is not using (rpm/deb) packaging to handle the namespace rebuild.  There are several of those - Arch, etc.

The underlying change (not depending on the configure.ac mechanisms to detect the distro) is also going to mean SuSE platforms are no longer detected correctly in the build, so any future makefile change needing this information may get a surprise - possibly leading to bugs.

Is there any other reliable way we can detect suse platforms in configure.ac?  In the worst case, I guess you could force PACKAGE_DISTRIBUTION to be set in the spec file, before configure is run, to ensure configure.ac still comes up with the correct result.

ddiss:
I'll go down the PACKAGE_DISTRIBUTION route and withdraw this change. Thanks for the feedback.

PR #622, opened 2019-03-18
To show correct time difference from UTC

ShivamJ2047:
PCP is showing correct time zone but the time difference shown is negative of the actual one. This change will show correct time difference.


myllynen:
I think this is by design due to POSIX convention, see e.g. https://unix.stackexchange.com/questions/104088/why-does-tz-utc-8-produce-dates-that-are-utc8. Note that ISO 8601 is probably more easily understood non-programmers, see https://bugzilla.redhat.com/show_bug.cgi?id=1146838. Thanks.

ShivamJ2047:
ohh i didn't know about that. I will close it then.

PR #685, opened 2019-06-27
pmdabcc: add selinux rules to fix bcc syscall prefix detection (fixes…

andreasgerstmayr:
… execsnoop)

get_syscall_fnname of bcc relies on /proc/kallsyms to detect the syscall
prefix (depending on the architecture). To read the addresses of
/proc/kallsyms the PMDA requires the CAP_SYSLOG capability, which
needs to be granted by selinux


lberk:
@andreasgerstmayr thanks! It's always appreciated when folks help out with selinux stuff.  One small thing, do you have the original AVC still?  We try to capture them all both; directly above the corresponding rule in `pcpupstream.te.in` as well as in the regression test in `qa/1622`  I believe ken added a `src/selinux/next-xxx-yyy` script to help with that.

lberk:
Oh, and please do a sanity check on el6 or el7 era distros.  Often times newer selinux rules won't build there, it'd be good make sure this rule change is compatible.  Thanks!

andreasgerstmayr:
> One small thing, do you have the original AVC still?

There was no AVC, the kernel is using `has_capability_noaudit` (https://github.com/SELinuxProject/selinux-kernel/blob/master/kernel/kallsyms.c#L653), which doesn't produce any audit logs (which made the bug even trickier to resolve).

> Oh, and please do a sanity check on el6 or el7 era distros. Often times newer selinux rules won't build there, it'd be good make sure this rule change is compatible.

Ok, I'll setup vagrant and check out your provisioning scripts :)

Issue #13, opened 2015-04-14
make fails after sudo make install


rvernica:
When compiling version 3.10.3 from source, after running `sudo make install`, running `make` fails due to `Permission denied`. Here are the steps to reproduce:

```
$ wget ftp://ftp.pcp.io/projects/pcp/download/pcp-3.10.3.src.tar.gz
...
$ ./configure --prefix=/usr --sysconfdir=/etc --localstatedir=/var
$ make
$ sudo make install
$ make
...
=== pmieconf ===
gcc -fPIC -fno-strict-aliasing -D_GNU_SOURCE -fstack-protector-all -D_FORTIFY_SOURCE=2 -I../../src/pmie/src -Wall -O2 -g -DPCP_DEBUG -DPCP_VERSION=\"3.10.3\" -I../../src/include -I../../src/include/pcp -o pmieconf  -Wall -L../../src/libpcp/src -L../../src/libpcp_pmda/src   pmieconf.o rules.o io.o     -lpcp  
rm: cannot remove `rules/cpu/load_average': Permission denied
rm: cannot remove `rules/cpu/system': Permission denied
rm: cannot remove `rules/cpu/util': Permission denied
rm: cannot remove `rules/cpu/context_switch': Permission denied
rm: cannot remove `rules/cpu/low_util': Permission denied
rm: cannot remove `rules/global/parameters': Permission denied
rm: cannot remove `rules/global/pcp_actions': Permission denied
rm: cannot remove `rules/pernetif/packets': Permission denied
rm: cannot remove `rules/pernetif/errors': Permission denied
rm: cannot remove `rules/pernetif/util': Permission denied
rm: cannot remove `rules/pernetif/collisions': Permission denied
rm: cannot remove `rules/memory/swap_low': Permission denied
rm: cannot remove `rules/memory/exhausted': Permission denied
rm: cannot remove `rules/filesys/filling': Permission denied
rm: cannot remove `rules/percpu/some_util': Permission denied
rm: cannot remove `rules/percpu/many_util': Permission denied
rm: cannot remove `rules/percpu/system': Permission denied
mkdir: cannot create directory `rules': File exists
make[2]: *** [local] Error 1
make[1]: *** [default_pcp] Error 2
make[1]: Leaving directory `/home/debian/pcp-3.10.3/src'
make: *** [default_pcp] Error 2
```

I would guess the permissions need to be reset after `sudo make install` runs.

kmcdonell:
[...]

Issue #10, opened 2015-04-10
Completion of error handling


elfring:
Would you like to add more error handling for return values from functions like the following?
- [pthread_create](http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_create.html "Thread creation") ⇒ [rpm_init](https://github.com/performancecopilot/pcp/blob/f04f926f4c5dc9cea1c88d0da385d37c42a2c9e4/src/pmdas/rpm/rpm.c#L596)
- [pthread_mutex_init](http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_mutex_init.html "Initialize a mutex") ⇒ [monitor_init](https://github.com/performancecopilot/pcp/blob/cec143a52ffd8deac816b5d3ad4ea2caf2d67f3d/src/pmdas/perfevent/perfmanager.c#L94)

natoscott:
Yes.  Do you want to send a patch or pull request?

kmcdonell:
According to my man page ... pthread_mutex_init always returns 0.
The pthread_create one though is a bug that I'll fix.

PR #693, opened 2019-07-09
Fix segfault in pmdaFetch().

shogo-matsumoto:
I try to fix the following issue.
https://github.com/cockpit-project/cockpit/issues/6108

Sometimes cockpit-pcp crashes due to segfault in __pmFindProfile().
I think the cause is in __pmFetchLocal() of src/libpcp/src/fetchlocal.c.

Cockpit calls pmdaFetch() periodically. pmdaFetch() refers to pmda->e_prof
via __pmdaCountInst() to calculate the number of instances. 
Just before calling pmdaFetch() pmda->e_prof has a chance to be updated in
__pmFetchLocal() but it should satisfy a condition "ctxp->c_sent != dp->domain".

This condition is not appropriate I think because there is a following case:

 1. pmDestroyContext() frees pmda->e_prof
 2. pmda->e_prof area is reallocated and used for other purpose
 3. __pmFetchLocal() is called
 4. __pmFetchLocal() does not update pmda->e_prof due to the condition
 5. __pmFetchLocal() calls pmdaFetch()
 6. pmdaFetch() accesses corrupted area by referring to pmda->e_prof

I'm not sure it is best answer but I suggest a fix by removing the condition
in order to replace pmda->e_prof with a new profile at any time before using it.

### DIFF
--- a/src/libpcp/src/fetchlocal.c
+++ b/src/libpcp/src/fetchlocal.c
@@ -95,25 +95,23 @@ __pmFetchLocal(__pmContext *ctxp, int numpmid, pmID pmidlist[], pmResult **resul
 	    /* based on domain, unknown PMDA */
 	    sts = PM_ERR_NOAGENT;
 	else {
-	    if (ctxp->c_sent != dp->domain) {
-		/*
-		 * current profile is _not_ already cached at other end of
-		 * IPC, so send get current profile ...
-		 * Note: trickier than the non-local case, as no per-PMDA
-		 *	 caching at the PMCD end, so need to remember the
-		 *	 last domain to receive a profile
-		 */
-		if (pmDebugOptions.fetch)
-		    fprintf(stderr, 
-			    "__pmFetchLocal: calling ???_profile(domain: %d), "
-			    "context: %d\n", dp->domain, ctx);
-		if (dp->dispatch.comm.pmda_interface >= PMDA_INTERFACE_5)
-		    dp->dispatch.version.four.ext->e_context = ctx;
-		sts = dp->dispatch.version.any.profile(ctxp->c_instprof,
-						dp->dispatch.version.any.ext);
-		if (sts >= 0)
-		    ctxp->c_sent = dp->domain;
-	    }
+	    /*
+	     * current profile is _not_ already cached at other end of
+	     * IPC, so send get current profile ...
+	     * Note: trickier than the non-local case, as no per-PMDA
+	     *       caching at the PMCD end, so need to remember the
+	     *       last domain to receive a profile
+	     */
+	    if (pmDebugOptions.fetch)
+	        fprintf(stderr, 
+	                "__pmFetchLocal: calling ???_profile(domain: %d), "
+	                "context: %d\n", dp->domain, ctx);
+	    if (dp->dispatch.comm.pmda_interface >= PMDA_INTERFACE_5)
+	        dp->dispatch.version.four.ext->e_context = ctx;
+	    sts = dp->dispatch.version.any.profile(ctxp->c_instprof,
+	                                    dp->dispatch.version.any.ext);
+	    if (sts >= 0)
+	        ctxp->c_sent = dp->domain;
 	}
 
 	/* Copy all pmID for the current domain into the temp. list */


shogo-matsumoto:
Hi @kmcdonell,
Thank you for the feedback. 
I will review my commit but it is helpful if you could update it.
Thanks.

kmcdonell:
@shogo-matsumoto Sorry for the long silence here, I've not been able to spend much time on PCP recently.
I've reread you're original comment, and realize I overlooked/misunderstood one very important aspect ... this is in libpcp_pmda, not libpcp.  There has been no threading work in libpcp_pmda at all, only in libpcp.

Now just to make sure I am chasing the right bug this time, could you please confirm that the application environment you're seeing this in is a multi-threaded client (so using libpcp and the PMAPI), but it is also using contexts created with PM_CONTEXT_LOCAL, and these contexts are being dynamically created and destroyed across multiple threads.  Is that correct?

kmcdonell:
@shogo-matsumoto please ignore my last comment ... this cannot be a multi-threaded app as the pmFetch would fail.
On digging a little deeper I can see https://github.com/cockpit-project/cockpit/issues/6108 ... and
this mentions https://bugzilla.redhat.com/show_bug.cgi?id=1235962 and this BZ contains this traceback from testing on F22:
Truncated backtrace:
Thread no. 1 (10 frames)
 #0 __pmFindProfile at profile.c:144
 #1 __pmInProfile at profile.c:163
 #2 __pmdaNextInst at callback.c:148
 #3 pmdaFetch at callback.c:514
 #4 linux_fetch at pmda.c:5714
 #5 __pmFetchLocal at fetchlocal.c:131
 #6 pmFetch at fetch.c:147
 #7 cockpit_pcp_metrics_tick at src/bridge/cockpitpcpmetrics.c:349
 #8 on_timeout_tick at src/bridge/cockpitmetrics.c:178
 #13 g_main_context_iteration at gmain.c:3869

Now I need to check with Juan Antonio Clavero and Martin Pitt to make sure we're all agreed that this the one and same issue.

Assuming for the moment that it is, then I think the failure scenario is ...
+ using PM_CONTEXT_LOCAL
+ single threaded
+ create a new context
+ do some fetches
++ in __pmFetchLocal if we are using a profile, the pointer ctxp->c_instprof gets copied to pmda->e_profile via a call path that looks like __pmFetchLocal() -> dispatch.version.any.profile() -> pmdaProfile() ... at the end the pointer copy looks like pmda->e_prof = prof;
+ close context
++ pmDestroyContext calls _pmFreeProfile(ctxp->c_instprof) and this leaves pmda->e_prof in the PMDA dangling in the breeze
+ next pmFetch is potentially kaboom

I can see how to easily fix this, but first I want to make sure all the interested parties agree on the analysis above.

shogo-matsumoto:
Hi @kmcdonell,
Thank you for sharing the information.
I agree with you that the issue does not depend on multi-threaded.

The program (cockpit-pcp) which uses libpcp.so and libpcp_pmda.so in my environment(Fedora 30 + cockpit github version) has two threads but it seems that only one thread (Thread 1) calls pcp functions such as pmDestroyContext() or __pmFetchLocal().

In my test, contexts are created/destroyed when I change data-range (5 minutes, 1 hour, 6 hours, ...) on dashboard page of cockpit.

kmcdonell:
Thanks for the update @shogo-matsumoto.
Do you know if the cockpit-pcp application is using more than one PMAPI context within the one thread that's doing the PCP work?
Would it be possible to run cockpit-pcp with -Dcontext,fetch,profile (assuming it implements the standard PCP debugging options from the command line), and send me the stderr output for the failing case?

kmcdonell:
Good news ... I have developed a test program that reproduces the SEGV ...

(gdb) where
#0  0x00007ffff7f02f9a in __pmFindProfile (indom=251658247, prof=0x555555574410) at profile.c:145
#1  0x00007ffff7f02ff1 in __pmInProfile (indom=251658247, prof=0x555555574410, inst=0) at profile.c:164
#2  0x00007ffff7fac061 in __pmdaNextInst (inst=0x7fffffffd750, pmda=0x555555565ce0) at callback.c:161
#3  0x00007ffff7fac1a2 in __pmdaCountInst (dp=0x7ffff78d2668, pmda=0x555555565ce0) at callback.c:188
#4  0x00007ffff7facf0c in pmdaFetch (numpmid=1, pmidlist=0x555555564a20, resp=0x7fffffffdab0, pmda=0x555555565ce0) at callback.c:569
#5  0x00007ffff78b448e in ?? () from /var/lib/pcp/pmdas/linux/pmda_linux.so
#6  0x00007ffff7f29437 in __pmFetchLocal (ctxp=0x55555555a2a0, numpmid=1, pmidlist=0x5555555592a8 <pmid>, result=0x7fffffffdcf8) at fetchlocal.c:129
#7  0x00007ffff7ee97d4 in pmFetch_ctx (ctxp=0x55555555a2a0, numpmid=1, pmidlist=0x5555555592a8 <pmid>, result=0x7fffffffdcf8) at fetch.c:160
#8  0x00007ffff7ee9c38 in pmFetch (numpmid=1, pmidlist=0x5555555592a8 <pmid>, result=0x7fffffffdcf8) at fetch.c:234
#9  0x0000555555555516 in do_inner () at profilecrash.c:122
#10 0x00005555555561f8 in main (argc=5, argv=0x7fffffffde78) at profilecrash.c:345

Now, onto the fix.

shogo-matsumoto:
The issue seems to be fixed at the commit aa14df7. Thank you for your help.

PR #714, opened 2019-08-03
making dictDelete function available

Erbenos:
I will need dictDelete function available in StatsD PMDA, therefore making it available.


Erbenos:
Moved.

PR #724, opened 2019-08-21
tweak systemd service scripts

goodwinos:
The pmlogger_check service is not required by pmlogger.service. This was causing them to run in parallel. Also, add a 120s timeout in all service scripts for both start and stop to avoid timeouts in Azure CI checks.

### DIFF
--- a/src/pmlogger/pmlogger.service.in
+++ b/src/pmlogger/pmlogger.service.in
@@ -7,6 +7,7 @@ BindsTo=pmlogger_check.timer pmlogger_daily.timer pmlogger_daily-poll.timer
 
 [Service]
 Type=forking
+TimeoutSec=120
 Restart=always
 ExecStart=@PCP_RC_DIR@/pmlogger start
 ExecStop=@PCP_RC_DIR@/pmlogger stop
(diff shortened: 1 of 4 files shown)


goodwinos:
> I don't follow how 120s timeouts help CI improve? What happens if the script hits that timeout - surely the test still fails, right?

Paste from IRC conversation -

<nathans> what happens if it hits that timeout - surely the test still fails right?  so, don't follow how it helps
<nathans> mgoodwin, will ask in the gh issue
<mgoodwin> the assertion is CI often fails because the default timeout (30s) isnt long enough
<mgoodwin> removing the RequireBy=pmlogger in pmlogger_check.service is because Ken is seeing failures that I think are due to this
<mgoodwin> those two services should not run in parallel
<mgoodwin> same as pmlogger and pmlogger_daily
<nathans> ok - remind me - if I enable pmlogger, how do the daily and check scripts get enabled?
<mgoodwin> by their timers
<nathans> how do their timers get enabled?
<nathans> (off by default, aren't they?)
<mgoodwin> they're requiredby=pmlogger.service
<mgoodwin> so if pmlogger.service is enabled and active, the timers are start too
<nathans> definitely?  ok
<nathans> sounds good
<mgoodwin> yep
<mgoodwin> next step here would be to convert over to type=notify
...
<mgoodwin> looks like CI passed, except on rawhide (gpgkeys issue of some sort) and debian-testing (no sane version of 'ping' available)
<nathans> cool
<nathans> those were pre-existing issues, i noticed em too
<mgoodwin> that's an improvement at least anyway
<nathans> ping may have moved to some non-default package on debian testing i guess
<nathans> +1

PR #725, opened 2019-08-21
pmdastatsd: generate man page from pod

lzap:
This stops the insanity of maintaining man page.


natoscott:
@lzap there's some possible down-sides here, e.g. this man page will no longer appear on online resources like pcp.io and man7.org (due to the way they're built) ... is it worth it?  Up to you and @Erbenos

lzap:
Damn, it was painful to convert it back and I can hardly asses how painful it was to write it directly.

Here is an alternative proposal - an extra target that generates it manually so we would commit the generated `.1` file into git with some comment the file was autogenerated so people won't accidentaly change it.

If you agree I will refactor this.

Erbenos:
I am actually doing some doc changes in my fork and since that one still has no pod doc files, can we wait with this, until .1 doc is merged with my version?

lzap:
Of course, this is not important. If you prefer to maintain `.1` file manually I can also just throw this away.

PR #735, opened 2019-08-27
Added support for scientific notation for metric values in StatsD payloads

Erbenos:
+ ...and small fix for underflow detection

Payloads with values such as these are valid:
:+1:
```
test:1e1|g
test:+1e+1|g
test:-1e-1|g
test:1.1e1|g
```
These are not:
:-1:
```
test:-.2e2|g
test:.2e2|g
test:e2|g
test:2e2.2|g
```

### DIFF
--- a/src/pmdas/statsd/README.md
+++ b/src/pmdas/statsd/README.md
@@ -356,7 +356,6 @@ These names are blacklisted for user usage. No messages with these names will pr
 
 # Roadmap
 - Make sure code is optimized
-- Allow _value_ to be expressed in _e_ notation
 
 # FAQ
 
--- a/src/pmdas/statsd/src/aggregator-metric-gauge.c
+++ b/src/pmdas/statsd/src/aggregator-metric-gauge.c
@@ -57,29 +57,33 @@ int
 update_gauge_value(struct agent_config* config, struct statsd_datagram* datagram, void* value) {
     (void)config;
     double old = *(double*)value;
-    double new_value;
+    double increment;
     switch (datagram->explicit_sign) {
         case SIGN_MINUS:
-            new_value = -1.0 * datagram->value;
+            increment = -1.0 * datagram->value;
             break;
         default:
-            new_value = datagram->value;
-    }
-    // check for overflow
-    if (old > 0 && new_value > DBL_MAX - old) {
-        VERBOSE_LOG(2, "Caught double overflow.");
-        return 0;
-    }
-    // check for underflow
-    if (old < 0 && new_value < DBL_MAX - old) {
-        VERBOSE_LOG(2, "Caught double underflow.");
-        return 0;
+            increment = datagram->value;
     }
     if (datagram->explicit_sign == SIGN_NONE) {
-        *(double*)(value) = new_value;
-    } else {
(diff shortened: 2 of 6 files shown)


natoscott:
@Erbenos looks like something that would be good to test - can we extend the testsuite to exercise this change?

Erbenos:
You are right, will do next. While working on this I noticed that with
statsd Makefile the rules for unit tests for parsers are gone, so will
probably add that and integrate those as well.

natoscott:
> You are right, will do next. While working on this I noticed that with statsd Makefile the rules for unit tests for parsers are gone, so will probably add that and integrate those as well.

Oh, not gone - they reside in the PCP QA suite now - see test 1599.  You can run them (and all statsd tests) via:

```
$ cd qa
$ ./check -g pmda.statsd
```

Erbenos:
The unit tests for parsers work differently then the e2e tests from 1599
(unless I am missing something)

They are defined in src/pmdas/statsd/src/parser-basic.c at bottom of the
file (same for parser-ragel.rl). Idea was to build the binary with those
test specific D flags and then check for their output.

natoscott:
> The unit tests for parsers work differently than the e2e tests from 1599 (unless I am missing something). They are defined in src/pmdas/statsd/src/parser-basic.c at bottom of the file (same for parser-ragel.rl). Idea was to build the binary with those test specific D flags and then check for their output.

Ah yes, you're right - some makefile tweaks will be needed there then.

cheers.

lzap:
Out of curiosity, do tests try to launch the agent both with regular and ragel parsers?

Erbenos:
The tests in my pmdastatsd repo do (those in my PCP fork too), but those
don't contain scientific notation payload. The tests in main PCP repository
for statsd are non-existant as of now.
Ideally unit tests will be extended as well to include scientific notation
payload. There are no tests written for this as of now.

lzap:
It works great on my development setup. I haven't done anything big, about 1k packets delivered and parsed. One nitpick tho, units do not seem to be reported correctly. Since statsd protocol is always in miliseconds, the agent should set the units correctly so PCP tools show it properly:

```
metric:    statsd.fm_rails_http_request_total_duration
host:      box.home.lan
semantics: instantaneous value
units:     none
samples:   all
full label for instance[0]: /percentile99::action=index::controller=hosts_controller
```

This only applies to counters and timings. For gauge the unit is really unknown as statsd does not provide much info about the metrics.

Erbenos:
> It works great on my development setup. I haven't done anything big, about 1k packets delivered and parsed. One nitpick tho, units do not seem to be reported correctly. Since statsd protocol is always in miliseconds, the agent should set the units correctly so PCP tools show it properly:
> 
> ```
> metric:    statsd.fm_rails_http_request_total_duration
> host:      box.home.lan
> semantics: instantaneous value
> units:     none
> samples:   all
> full label for instance[0]: /percentile99::action=index::controller=hosts_controller
> ```
> 
> This only applies to counters and timings. For gauge the unit is really unknown as statsd does not provide much info about the metrics. Do you want me to report a bug/issue?

Fixed in 8fcd8b9

PR #736, opened 2019-08-28
libpcp: fix multi-archive replay error handling with corrupted archive(s)

goodwinos:
Changing libpcp/interp is fraught with danger - I'd appreciate review for this one!
There may be other or better ways to solve the problem; improving the error
handling seems to be the best option but I'm open to alternative suggestions,
thanks in advance.

RHBZ#1673053

libpcp multi-archive replay fails to propagate PM_ERR_LOGREC errors back
up the stack to the caller, causing pmrep and other tools to continue
forever (unless given a -T finishing time window bound). Without the fix,
the archives in qa/archives/multi-corrupted demonstrate this.

New QA test 1671 exercises this with qa/archives/multi-corrupted.
Without the fix, 1671.full will be rather large (but limited to 10000
samples). The interp algorithm doesn't seem to make progress and just
keeps looping at posn=60624.

Other QA fallout - have tested QA groups logutil, pmlogextract, archive,
multi-archive, libpcp, decompress-xz and pmlogcheck.  There is one new
failure in qa/914 - this is a new test and fails with/without the fix.


natoscott:
LGTM, great work @goodwinos

PR #739, opened 2019-09-02
pmdalmsensors: initial rework to use json output instead of parsing r…

lherbolt:
Hi,
sensors have options to print out the output in json. Which is much easier 
to format and get proper output. Here is initial working update to use json output. 

Thanks for review.

### DIFF
--- a/src/pmdas/lmsensors/pmdalmsensors.python
+++ b/src/pmdas/lmsensors/pmdalmsensors.python
@@ -19,6 +19,7 @@
 
 # pylint: disable=bad-continuation,line-too-long,too-few-public-methods
 
+import json
 import subprocess
 import re
 import argparse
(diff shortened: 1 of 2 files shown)


christianhorn:
Great, much better than parsing the output.
-                if int(sensorvalues[sensornames[item]]) < 0:
+                if int(sensorvalues[sensornames[item]]) <= 0:
Why sort out also 0? Isn't it a valid value to be returned? I thought that just negative values make no sense, so should be "None".

lherbolt:
> Great, much better than parsing the output.
> 
> ```
> -                if int(sensorvalues[sensornames[item]]) < 0:
> +                if int(sensorvalues[sensornames[item]]) <= 0:
> ```
> 
> Why sort out also 0? Isn't it a valid value to be returned? I thought that just negative values make no sense, so should be "None".

Originally I did it b/c some of my fan are not present and it sensors reports 0 rpm. Now, thinking about it 
it is probably better to sort out just the negative values.

christianhorn:
> [..] Now, thinking about it
> it is probably better to sort out just the negative values.
Thanks, I think so to. For example, the thinkpads fan here is just spinning up from time to time, having 0 reported as UKN would be bad.
When working on this implementation, I wondered weather invalidating of values should be done in pmda-lmsensors or in lmsensors, which is already doing preprocessing like doing value corrections for some sensors. I did not see this configurable in lmsensors as of today, so doing it in the pmda was the pragmatic solution. We could consider a more flexible implementation with i.e. a small configfile in pcp/pmdas/lmsensors, where a user can choose which sensor values should be considered UKN. I did not really have a need so far.

src/pmdas/lmsensors/pmdalmsensors.python
```diff
@@ -41,87 +42,25 @@ def lmsensors_get():
...
-            continue
-
-        tmp = re.sub(r'_input: .*', "", outline2)
-        level3 = re.sub(r'\s+', "", tmp)             # i.e. temp1
+        p = subprocess.Popen(["/usr/bin/sensors", "-j"], stdout=subprocess.PIPE)
+        output = json.loads(re.sub("[0-9]_","_", p.communicate()[0].decode("utf-8")))
         if args.debug:
-            if args.debug == 2:
-                print("    assigned level3:", level3)
+            print("function lmsensors_get(), json object after removing redudant naming:\n", output)
```
natoscott: Little typo there, "redudant" -> "redundant"

christianhorn:
When we keep the "if int(sensorvalues[sensornames[item]]) < 0:", then it looks good to me. I do not see a hint on a new pull request, so writing it explicitly.

Keeping qa/1480 unchanged does not hurt us for now. If lmsensors has bugs where '-u' and '-j' contents differ, then qa/1480 might eventually start to fail, but as of today it looks good. 

I remember that when I worked on the current pmda-lmsensors, lmsensors did still not provide the nicely machine parsable '-j'.

We might need to consider the versions: '-j' came into lm_sensors in April 2017. 
For example, RHEL7.* all ship lm_sensors-3.4.0 versions without '-j'. RHEL8.0 and onwards know '-j', and these are also lm_sensors-3.4.0 versions. We should take this as a hint that pmda-lmsensors might get used on systems with older lmsensors.. to the least we should state something in readme/manpage?
Unfortunately, just looking at the lm_sensors version is as seen above not enough to know if lmsensors supports '-j'.

lherbolt:
> If lmsensors has bugs where '-u' and '-j' contents differ, then qa/1480 might eventually start to fail, but as of today it looks good.

This is quite unlikely as the -j output is just pretty printed -u. Atleast it looks to me like it when I was 
checking the lmsensors source code. I will fix the typo and change back to report zero values as zeros.

PR #741, opened 2019-09-05
pmdabpftrace: don't use cached instance

andreasgerstmayr:
if a dict is used for indom.set_instances, the instance names get cached
using the pmdaCacheOp functionality. However, for bpftrace maps, the keys
should not be cached. For example, at some point the metric
bpftrace.scripts.script100.data.bytes can contain a histogram, and at another
time the same metric can contain an indom with process names as the instance
names.

TODO: somehow the first instance doesn't show up in pminfo, but the log file shows that all instances are passed to the `set_instances` and `replace_indom` functions?


andreasgerstmayr:
eaa5122f8785577eb273d5c1b9d4da0062d1b3c7 fixes the instance problems mentioned in the TODO -> ready to merge

PR #761, opened 2019-10-03
python pmdas: enable authentication support

andreasgerstmayr:
modified QA 1388 to test pmproxy authentication (currently failing)

is a general `pmdaSetCommFlags(&dispatch, PMDA_FLAG_AUTHORIZE);` for python PMDAs fine or should this be optionally enabled on a per-(python)-pmda-basis? none of the python pmdas call `set_attribute_callback` as of now (as this method is new in this commit), so the overhead should be minimal?


andreasgerstmayr:
Thanks for the review! I've added a `set_comm_flags` method to the Python PMDA base class and exported the `PMDA_FLAG_AUTHORIZE` and `PMDA_FLAG_CONTAINER` flags.

PR #805, opened 2019-12-07
configure: make build reproducible

jelly:
pcp.lsm embeds the build date which makes the package un reproducible,
use SOURCE_DATE_EPOCH to always generate a deterministic builddate when
reproducing pcp.

Motiviation: https://reproducible-builds.org/

Signed-off-by: Jelle van der Waa <jelle@archlinux.org>


natoscott:
@jelly thanks!  I've tweaked it a little to provide a fallback for platforms that do not provide a long-option 'date' command - please double check?  cheers.

Issue #625, opened 2019-04-02
pcp pidstat crashes on a pcp-zeroconf collected archive


mbaldessari:
I collected some archives on rhel8 beta via pcp-zeroconf. I am now trying to analyze it on my F30 laptop (pcp-4.3.1-2.fc30.x86_64):
$ pcp -a 20190330 pidstat -f '%y-%m-%d %H:%M:%S' -t30m -r > 0330-pistat.txt
Traceback (most recent call last):                                           
  File "/usr/libexec/pcp/bin/pcp-pidstat", line 916, in <module>             
    sts = manager.run()                                                      
  File "/usr/lib64/python3.7/site-packages/pcp/pmcc.py", line 665, in run    
    self._printer.report(self)                                               
  File "/usr/libexec/pcp/bin/pcp-pidstat", line 846, in report               
    ncpu = self.get_ncpu(group)                                              
  File "/usr/libexec/pcp/bin/pcp-pidstat", line 829, in get_ncpu             
    return group['hinv.ncpu'].netValues[0][2]                                
IndexError: list index out of range                                          

Archive is here: http://acksyn.org/files/tripleo/undercloud-0.redhat.local/

In this case it was the 20190330 one.

mbaldessari:
So just to be more complete, it seems that when pmlogger was collecting the 20190330 archive, the system might have been under a lot of memory pressure and it might very well be that the OOM killer went on a spree and the archive might not have all the expected data

PR #822, opened 2020-01-20
pcp-pidstat: fix out of range error when reading metrics from an archive

shogo-matsumoto:
pcp pidstat fails with IndexError when proc.psinfo.utime metric is found prior to
kernel.uname.* or hinv.ncpu metrics.
Check whether those metrics exist before starting to show process information.

Resolves: https://github.com/performancecopilot/pcp/issues/625

The error in my environment is as follows.

```
# pcp pidstat -a archive
  File "/usr/libexec/pcp/bin/pcp-pidstat", line 916, in <module>
    sts = manager.run()
  File "/usr/lib64/python3.6/site-packages/pcp/pmcc.py", line 665, in run
    self._printer.report(self)
  File "/usr/libexec/pcp/bin/pcp-pidstat", line 838, in report
    self.print_machine_info(group, manager)
  File "/usr/libexec/pcp/bin/pcp-pidstat", line 820, in print_machine_info
    header_string += group['kernel.uname.sysname'].netValues[0][2] + '  '
IndexError: list index out of range
```

### DIFF
--- a/src/pcp/pidstat/pcp-pidstat.py
+++ b/src/pcp/pidstat/pcp-pidstat.py
@@ -833,6 +833,9 @@ class PidstatReport(pmcc.MetricGroupPrinter):
             # need two fetches to report rate converted counter metrics
             return
 
+        if not group['hinv.ncpu'].netValues or not group['kernel.uname.sysname'].netValues:
+            return
+
         if not self.Machine_info_count:
             self.print_machine_info(group, manager)
             self.Machine_info_count = 1


natoscott:
@shogo-matsumoto looking at the code I see a few other metrics not catered for - print_machine_info() also uses kernel.uname.release, kernel.uname.nodename and kernel.uname.machine.

Perhaps instead of testing specific metrics in this way, we could add a try/except around print_machine_info() and return if IndexError is the exception?

I'll write a little QA test for this - for that, we could use an archive based on qa/archives/pcp-pidstat but with these metrics removed via pmlogextract(1).

shogo-matsumoto:
@natoscott Thank you for your comment. As you mentioned my fix is not perfect, so I think it's good idea to use try/exception to handle this issue. It seems that we need also to surround get_ncpu() considering the case of missing hinv.ncpu metric.

natoscott:
@shogo-matsumoto yes, I think you are correct.

natoscott:
@shogo-matsumoto how does this patch look to you?
https://paste.centos.org/view/3e180e8b

shogo-matsumoto:
@natoscott  Thank you. I haven't test yet, but  isn't it necessary to check IndexError for `ncpu = self.get_ncpu(group)` like the following?
```
diff --git a/src/pcp/pidstat/pcp-pidstat.py b/src/pcp/pidstat/pcp-pidstat.py
index 1c95267f7..2c0f85c26 100755
--- a/src/pcp/pidstat/pcp-pidstat.py
+++ b/src/pcp/pidstat/pcp-pidstat.py
@@ -833,16 +833,20 @@ class PidstatReport(pmcc.MetricGroupPrinter):
             # need two fetches to report rate converted counter metrics
             return

-        if not self.Machine_info_count:
-            self.print_machine_info(group, manager)
-            self.Machine_info_count = 1
+        try:
+            ncpu = self.get_ncpu(group)
+            if not self.Machine_info_count:
+                self.print_machine_info(group, manager)
+                self.Machine_info_count = 1
+        except IndexError:
+            # missing some metrics
+            return

         ts = group.contextCache.pmLocaltime(int(group.timestamp))
         timestamp = time.strftime(PidstatOptions.timefmt, ts.struct_time())
         interval_in_seconds = self.timeStampDelta(group)
         header_indentation = "        " if len(timestamp)<9 else (len(timestamp)-7)*" "
         value_indentation = ((len(header_indentation)+9)-len(timestamp))*" "
-        ncpu = self.get_ncpu(group)

         metric_repository = ReportingMetricRepository(group)
```

natoscott:
@shogo-matsumoto I don't think that's strictly necessary because print_machine_info() already calls get_ncpu() internally, so it will trigger the exception on first call.  However, your updated code does make the reason for the exception more clear so let's go with that.  Thanks!

shogo-matsumoto:
@natoscott 
> print_machine_info() already calls get_ncpu() internally,

I misunderstood that. Thank you for merging. I tested and confirmed that the issue does not reproduce at the latest commit.

PR #836, opened 2020-02-06
pmdastatsd new QA and small agent changes

Erbenos:
New QA has better coverage than the old one. Includes a Valgrind test.

pmdastatsd received small changes to _double_ metric values handling and improvements to clean up procedures.


qa/statsd/src/cases/04.py
```diff
@@ -0,0 +1,152 @@
...
+# -*- coding: utf-8 -*-
+
+# Exercises updates of non-config hardcoded metrics:
+# - statsd.pmda.received (before: 0, after: 75)
+# - statsd.pmda.parsed (before: 0, after: 45)
+# - statsd.pmda.aggregated (before: 0, after: 35)
+# - statsd.pmda.dropped (before: 0, after: 40)
+# - statsd.pmda.time_spent_parsing (before: 0, after: non-zero)
+# - statsd.pmda.time_spent_aggregating (before: 0, after: non-zero)
+# - statsd.pmda.metrics_tracked, with its counter, gauge, duration and total instances (before: 0, 0, 0, 0; after: 4, 4, 2, 10)
```
lzap: I don't know how these tests work, but is this supposed to check also the before/after state?

qa/statsd/src/cases/09.py
```diff
@@ -0,0 +1,81 @@
...
+
+expected_min = 0
+expected_max = 20000000
+expected_count = 10000000
+expected_average = 10000000
+expected_median = 10000000
+expected_percentile90 = 18000000
+expected_percentile95 = 19000000
+expected_percentile99 = 19800000
+expected_stddev = 5773500.278068
```
lzap: Ah disregard, I see how this is done here.

qa/statsd/src/cases/12.py
```diff
@@ -0,0 +1,157 @@
...
+    "session_duration:1wq|ms",
+    u"cache_cleared:4ěš|ms",
+    "session_started:1_4w|ms",
+    "session_started:2-1|ms",
+    "session_started:1|mss",
+    "cache_cleared:4|msd",
+    "cache_cleared:1|msa",
+    "session_started:|ms",
+    ":20|ms"
+]
```
lzap: Nitpick: If these "bad" metric names are not needed for these test, I'd probably drop these as these were already tested above.

qa/statsd/src/cases/13.py
```diff
@@ -0,0 +1,76 @@
...
+    time.sleep(3)
+    debug_output = utils.get_debug_file("debug_test")
+    print(debug_output)
+    utils.pmdastatsd_remove()
+    utils.restore_config()
+    utils.remove_debug_file("debug_test")
+
+run_test()
+
+
```
lzap: This test does only check of there are any errors I assume? Would it be feasible also to check those metrics created?

lzap:
Tests look good.

qa/statsd/src/cases/12.py
Erbenos: 12.py checks if the specificity of various verbose levels is ordered correctly. Payloads as above would generate additional err logs.

qa/statsd/src/cases/13.py
Erbenos: 08.py checks if labels are OK, the payload is nearly identical as in case of 13.py

qa/statsd/src/cases/09.py
```diff
@@ -0,0 +1,81 @@
...
+    labels_output = utils.request_metric("statsd.test_hdr")
+    output = utils.get_instances(labels_output)
+    for k, v in output.items():
+        status = False
+        number_value = float(v)
+        if k == "/average":
+            if utils.check_is_in_bounds(expected_average, number_value):
+                status = True
+        elif k == "/count":
+            # TODO: Ask Nathan, if this is OK
```
natoscott: Which part?  The 0.5?

qa/statsd/src/cases/06.py
```diff
@@ -0,0 +1,83 @@
...
+        utils.print_metric("statsd.pmda.dropped")
+        utils.print_metric('statsd.test_gauge')
+        overflow_payload = sys.float_info.max
+        sock.sendto("test_gauge2:+{}|g".format(overflow_payload).encode("utf-8"), (ip, port))
+        utils.print_metric("statsd.pmda.dropped")
+        utils.print_metric("statsd.test_gauge2")
+        sock.sendto("test_gauge2:+{}|g".format(overflow_payload).encode("utf-8"), (ip, port))
+        utils.print_metric("statsd.pmda.dropped")
+        utils.print_metric("statsd.test_gauge2")
+        # TODO: check if this is truly the desired behavior
```
natoscott: Now would be a good time to check, and figure out an answer (before this is merged) - otherwise, we're all likely to forget over time.

qa/statsd/GNUmakefile
```diff
@@ -2,14 +2,50 @@ TOPDIR = ../..
...
 
 ifeq "$(PMDA_STATSD)" "true"
 default setup default_pcp:
 
-install install_pcp: default $(RBFILES)
+install install_pcp: default $(TESTFILES)
 	$(INSTALL) -m 755 -d $(TESTDIR)
-	$(INSTALL) -m 644 $(RBFILES) $(TESTDIR)
+	# I wish I knew how to extract such listing from $(TESTFILES), not sure why this is needed yet
+	$(INSTALL) -m 644 src/cases/01.py $(TESTDIR)/src/cases/01.py
```
natoscott: In the same way TESTDIR is installed above, each new directory used also needs to be $(INSTALL)'d ... otherwise, the packaging ends up slightly incorrect (nothing 'owns' the directory is its not explicitly installed).

qa/statsd/GNUmakefile
```diff
@@ -2,14 +2,50 @@ TOPDIR = ../..
...
+	$(INSTALL) -m 644 src/cases/07.py $(TESTDIR)/src/cases/07.py
+	$(INSTALL) -m 644 src/cases/08.py $(TESTDIR)/src/cases/08.py
+	$(INSTALL) -m 644 src/cases/09.py $(TESTDIR)/src/cases/09.py
+	$(INSTALL) -m 644 src/cases/10.py $(TESTDIR)/src/cases/10.py
+	$(INSTALL) -m 644 src/cases/11.py $(TESTDIR)/src/cases/11.py
+	$(INSTALL) -m 644 src/cases/12.py $(TESTDIR)/src/cases/12.py
+	$(INSTALL) -m 644 src/cases/13.py $(TESTDIR)/src/cases/13.py
+	$(INSTALL) -m 644 src/cases/14.py $(TESTDIR)/src/cases/14.py
+	$(INSTALL) -m 644 src/cases/15.py $(TESTDIR)/src/cases/15.py
+	$(INSTALL) -m 644 src/configs/complex/0/pmdastatsd.ini $(TESTDIR)/src/configs/complex/0/pmdastatsd.ini
```
natoscott: Likewise for each of these new ini file dirs :|

Since these configs are so small, I'd consider generating them at runtime instead of committing them as individual files in the repo.

qa/1599
```diff
@@ -15,8 +15,9 @@ echo "QA output created by $seq"
 
 test -e $PCP_PMDAS_DIR/statsd/pmdastatsd || _notrun "statsd PMDA not installed"
```
natoscott: The comment at the start of this file refering to 'ruby' needs updating.

qa/1599
```diff
@@ -15,8 +15,9 @@ echo "QA output created by $seq"
```
natoscott: Consider replacing the above three lines with common.python now.

qa/1599
```diff
@@ -15,8 +15,9 @@ echo "QA output created by $seq"
 
 test -e $PCP_PMDAS_DIR/statsd/pmdastatsd || _notrun "statsd PMDA not installed"
 
-# NOTE: Miroslav is planning to re-work this in python/shell
-which ruby >/dev/null 2>&1 || _notrun "ruby not installed"
+which python >/dev/null 2>&1 || _notrun "python not installed"
```
natoscott: common.python will have checked this (python not installed) for you already.

qa/1599
```diff
@@ -28,41 +29,18 @@ status=1	# failure is the default!
...
-pmdastatsd_install
-
-cd $here/statsd
-scripts=`ls *.rb | LC_COLLATE=POSIX sort`
+cd $here/statsd/src
+scripts=`find cases -name *.py | LC_COLLATE=POSIX sort`
 for script in $scripts
 do
-    ruby $script
+    python $script $here/statsd/output
```
natoscott: Using common.python, this should then be $python rather than just python.

This lets us test both python2 and python3 more easily (we use python3 more aggressively - whenever available - over python2).

qa/statsd/src/cases/09.py
Erbenos: Yes, that one.

qa/statsd/src/cases/06.py
Erbenos: I actually already did the other day (we had small convo in Slack about it), emailed it trough with Lukas if it is OK and we resolved it. I will remove the comment in code.

Erbenos:
Thank you for your review so far, I will address the issues you brought up.

qa/statsd/src/cases/09.py
natoscott: Yes, that's fine - it's a common pattern - when writing shell tests we even have a shared _within_tolerance() function available for this purpose.

Erbenos: Alright, I will leave it as is and remove the TODO comment, thanks.

Erbenos: I had to set up a new VM to work on this on which I encounter following phenomena: even 0.5 toleration margin is not enough, particularly in relation to count of received UDP payloads, even after setting ```sysctl net.core.rmem_max``` and ```sysctl net.core.rmem_default``` to 26214400, which makes me worry about reliability of such a test - even though its UDP after all, what would you approach to this be?

natoscott: Set the acceptable boundaries sufficiently wide that any reasonable scenario would be catered for.  Add comments to the test discussing possible causes for values outside these boundaries, in case people have to come along later and widen the boundaries further.  Even with wide boundaries, its often still likely to be a worthwhile test.

Erbenos: Will do, thanks.

qa/1599
```diff
@@ -15,8 +15,9 @@ echo "QA output created by $seq"
 
 test -e $PCP_PMDAS_DIR/statsd/pmdastatsd || _notrun "statsd PMDA not installed"
 
-# NOTE: Miroslav is planning to re-work this in python/shell
-which ruby >/dev/null 2>&1 || _notrun "ruby not installed"
+which python >/dev/null 2>&1 || _notrun "python not installed"
+
+which valgrind >/dev/null 2>&1 || _notrun "valgrind not installed"
```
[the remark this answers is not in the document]
Erbenos: Addresed by 78b11bf

qa/statsd/GNUmakefile
Erbenos: I added GNUmakefile's to each qa/statsd dir in 78b11bf (as inspired by qa/slurm). I guess GNUmakefile.install are for post install steps once all test files are `$(INSTALL)`-ed? Will leave only `default setup install clean check` in them since python is interpreted.

Erbenos:
The commit above addressed issues mentioned but also disabled 13.py as I noticed the test case is not deterministic and will need to be filtered/processed in python further.

qa/statsd/GNUmakefile
natoscott: > I added GNUmakefile's to each qa/statsd dir in [78b11bf](https://github.com/performancecopilot/pcp/commit/78b11bfbc9987b9f97ed2b788ee88ae891abddce) (as inspired by qa/slurm).

Perfect.

> I guess GNUmakefile.install are for post install steps once all test files are `$(INSTALL)`-ed? 

This is the makefile that will be in the pcp-testsuite (rpm/deb) package, used when tests are run from /var/lib/pcp/testsuite

Issue #850, opened 2020-02-23
pcp-atop does not consider pids greater than 5-digits


miztake:
pcp-atop displays pid with "%5d".
However, since the pid is 2^22(approximately 4 million) at the maximum,
it is necessary to handle up to 7-digits.
When outputting 7-digit pid, the display is shifted as follows.

    00:11:12    pid command  cpu% |   pid command  cpu% |   pid command  cpu%_top3_
    00:11:12  1000048 pcp-atop 100% | 29765 gnome-sh  22% |  9259 X          6%
    00:22:32      1 systemd    0% |    12 watchdog   0% |   446 systemd-   0%

PR #851, opened 2020-02-23
pcp-atop: fix upper limit of output pid to 7-digits

miztake:
pcp-atop displays pid with "%5d".
However, since the pid is 2^22(approximately 4 million) at the maximum,
it is necessary to handle up to 7-digits.

Resolves https://github.com/performancecopilot/pcp/issues/850


natoscott:
@miztake looks good.  This affects regression tests qa/785 and qa/1390, which fortunately are testing the change you've made here.  I'll update the expected output of these tests when merging this, so please double check those changes and feel free to include updates to the tests in your PRs.  Thanks!

Issue #715, opened 2019-08-07
RFE: NFSv4 Server Clients Info


myllynen:
Linux kernel 5.3 NFSv4 server creates a directory under /proc/fs/nfsd/clients with information about current NFS clients, including which files they have open.

It might be nice to have have this information available as PCP metrics (at least current NFS clients, probably also open files). Thanks.

PR #857, opened 2020-02-27
Update nfs client statistics

jumitche:
This adds a number of additional nfs client statistics that the kernel exports


natoscott:
@jumitche thanks - looks good to me.

I'd like to extend the regression test coverage in qa/798 to cover some of these new metrics - could you add a sample mountstats file below qa/nfsclient/ which contains the new values from the kernel?

natoscott:
Would you be interested in working on these metrics too...?
https://github.com/performancecopilot/pcp/issues/715
(they're server metrics, code would live in src/pmdas/linux/proc_fs_nfsd.c I guess)

Alternatively, if you have a v4.2 server setup with interesting content in the directory @myllynen mentioned there, could you send me a tarball of that directory?  I'll work on it alongside some other kernel metrics I'm working on currently.  Thanks!

PR #860, opened 2020-03-16
pcp-atop: fix STDATE/STTIME in -v output

shogo-matsumoto:
STDATE/STTIME of pcp atop -v was wrong (e.g. STDATE=1970/01/01).
A new metric kernel.all.btime is added and used to fix the issue.

Note:
- This may not be best but I defined a new global function set_system_btime to pass btime to showprocs.c.
- ENDATE/ENTIME are also modified but not be tested because curstat->gen.elaps seems not to be implemented yet.

### DIFF
--- a/src/pcp/atop/atop.c
+++ b/src/pcp/atop/atop.c
@@ -164,6 +164,7 @@ char		**argvp;
 int		fetchmode;
 int		fetchstep;
 
+long long	system_boottime;
 
 struct visualize vis = {generic_samp, generic_error,
 			generic_end,  generic_usage,
@@ -587,7 +588,8 @@ engine(void)
 			if ((*vis.next)() < 0)
 				cleanstop(1);
 			goto reset;
-               }
+		}
+		system_boottime = cursstat->cpu.btime;
 
 		/*
 		** take a snapshot of the current task-level statistics 
--- a/src/pcp/atop/showprocs.c
+++ b/src/pcp/atop/showprocs.c
@@ -999,7 +999,11 @@ procprt_STDATE_ae(struct tstat *curstat, int avgval, int nsecs)
 {
         static char buf[11];
 
-        convdate(curstat->gen.btime, buf, sizeof buf);
+        if (system_boottime)
+                convdate(system_boottime + curstat->gen.btime / 1000, buf, sizeof buf);
+        else
+                strcpy(buf, "----/--/--");
+
         return buf;
 }
 
@@ -1011,7 +1015,11 @@ procprt_STTIME_ae(struct tstat *curstat, int avgval, int nsecs)
 {
         static char buf[9];
 
-        convtime(curstat->gen.btime, buf, sizeof buf);
+        if (system_boottime)
+                convtime(system_boottime + curstat->gen.btime / 1000, buf, sizeof buf);
+        else
+                strcpy(buf, "--:--:--");
+
         return buf;
 }
 
@@ -1033,7 +1041,10 @@ procprt_ENDATE_e(struct tstat *curstat, int avgval, int nsecs)
 {
         static char buf[11];
 
-        convdate(curstat->gen.btime + curstat->gen.elaps, buf, sizeof buf);
+        if (system_boottime)
+                convdate(system_boottime + curstat->gen.btime / 1000 + curstat->gen.elaps, buf, sizeof buf);
+        else
+                strcpy(buf, "----/--/--");
 
         return buf;
 }
@@ -1056,7 +1067,10 @@ procprt_ENTIME_e(struct tstat *curstat, int avgval, int nsecs)
 {
         static char buf[9];
 
-        convtime(curstat->gen.btime + curstat->gen.elaps, buf, sizeof buf);
+        if (system_boottime)
+                convtime(system_boottime + curstat->gen.btime / 1000 + curstat->gen.elaps, buf, sizeof buf);
+        else
+                strcpy(buf, "--:--:--");
 
         return buf;
 }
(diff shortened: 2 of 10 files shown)


src/pmdas/linux/help
```diff
@@ -775,6 +775,7 @@ e.g. mirroring the results from existing performance tools:
 @ kernel.all.sysfork fork rate metric from /proc/stat
 @ kernel.all.running number of currently running processes from /proc/stat
 @ kernel.all.blocked number of currently blocked processes from /proc/stat
+@ kernel.all.btime boot time from /proc/stat
```
natoscott: @shogo-matsumoto could we name this metric 'kernel.all.boottime' instead?

src/pmdas/linux/mk.rewrite
```diff
@@ -43,6 +43,7 @@ metric disk.dev.write_merge { type -> $KERNEL_ULONG }
 metric kernel.all.sysfork { type -> $KERNEL_ULONG }
 metric kernel.all.running { type -> $KERNEL_ULONG }
 metric kernel.all.blocked { type -> $KERNEL_ULONG }
+metric kernel.all.btime { type -> U64 }
```
natoscott: This line isn't necessary, its only for if we'd previously got the metadata incorrect (but this is a new metric, so we'll get it right first time)

src/pcp/atop/atop.c
```diff
@@ -588,6 +588,7 @@ engine(void)
 				cleanstop(1);
 			goto reset;
                }
+		set_system_btime(cursstat->cpu.btime);
```
natoscott: @shogo-matsumoto hmm, could this be done in various.c::setup_globals() rather than on every fetch iteration?

src/pcp/atop/showprocs.c
```diff
@@ -46,6 +46,10 @@ static char     *columnhead[] = {
 	[MSORTGPU]= "GPU",
 };
 
+static count_t sys_btime = 0;
```
natoscott: Let's declare a 'boottime' variable in atop.h with the other system-wide variables there (eg pagesize, etc).

natoscott:
> STDATE/STTIME of pcp atop -v was wrong (e.g. STDATE=1970/01/01).
> A new metric kernel.all.btime is added and used to fix the issue.

Great!

>     * This may not be best but I defined a new global function set_system_btime to pass btime to showprocs.c.
> 

Agreed, I think there's a better way - see review comments.

>     * ENDATE/ENTIME are also modified but not be tested because curstat->gen.elaps seems not to be implemented yet.

Ah, good find!  I think we can fix that too - the 'elaps' value there could be calculated as:
kernel.all.uptime - proc.psinfo.start_time
(uptime changed on every sample of course)

src/pcp/atop/atop.c
shogo-matsumoto: @natoscott Thank you for your feedback! It seems good idea but I'm concerned about if it works for an archive having multiple boot time which is generated by pmlogextract.

src/pmdas/linux/help
shogo-matsumoto: @natoscott Yes, of course.

src/pcp/atop/atop.c
natoscott: @shogo-matsumoto yes, you're right - needs to be checked on each sample then.  I still recommend use the 'boottime' global variable name, and updating it directly in photosyst(), rather than using a function called at the top level in atop.c::engine().

shogo-matsumoto: @natoscott Thank you for the reply.
> I still recommend use the 'boottime' global variable name
Yes, I agree.

shogo-matsumoto:
@natoscott Thank you for the feedback. I updated the commit. Could you review it?

natoscott:
> @natoscott Thank you for the feedback. I updated the commit. Could you review it?

@shogo-matsumoto It looks good to me.  I want to add a little test that uses 'pcp atop -v' from the command line, then I'll merge it.  Do you want to work on the curstat->gen.elaps issue you found?

shogo-matsumoto:
>Do you want to work on the curstat->gen.elaps issue you found?

I will try it if no one works on this issue. It may take some time. 
Is the following your comment correct?

>the 'elaps' value there could be calculated as: kernel.all.uptime - proc.psinfo.start_time

It does not seem correct for processes which had already finished. Do you mean that a new metric is needed for elaps which is updated every sample while the process is running?

natoscott:
> > Do you want to work on the curstat->gen.elaps issue you found?
> 
> I will try it if no one works on this issue. It may take some time.

OK, I'll also see if I can fit it in if noone else works on it first.

> Is the following your comment correct?
> 
> > the 'elaps' value there could be calculated as: kernel.all.uptime - proc.psinfo.start_time
> 
> It does not seem correct for processes which had already finished.

If the process is finished, pmFetch wont return a proc.psinfo.start_time instance/value for it.

natoscott:
> > > the 'elaps' value there could be calculated as: kernel.all.uptime - proc.psinfo.start_time

Since we don't currently sample kernel.all.uptime (/proc/uptime) in pcp-atop, we could alternatively perform the calculation in this way:

elapsed_time[pid] = (sample time from pmResult) - (kernel.all.boottime + proc.psinfo.start_time[pid])

shogo-matsumoto:
@natoscott Thank you for your comment. I will try to implement elaps.

natoscott:
> @natoscott Thank you for your comment. I will try to implement elaps.

Great!  I've also added commit qa/1347 in commit 56c8d93abc43c193894955b3cd76dfcc9dc626c4 which exercises the -v option now.  We should be able to extend that test to exercise the elaps fix too.

PR #867, opened 2020-03-24
pmlogger archive roll and strftime on basename

goodwinos:
Hi Ken, this (see below) is all done and pushed to my pmlogger-exec branch whilst I finish running QA and update docs/pmlogger(1). I've not yet updated the systemd configs, pending simplifying updates to the rc and logutil scripts, which I think is the next step.

> * argv munging so the args match the running instance ... this means
>     the strftime stuff may be internal to pmlogger (not requiring a
>     command line arg change) because the date+time pattern is really fixed
>   * loop with backoff in pmlogger in the case of collision on the output     archive name (think more than 1 SIGUSR1 inside a minute) ... which
>     would obviate the need for the -00, -01 ugliness in the names ...
>     pmlogger would just take a bit longer to get started until the
>     strftime re-evaluated name is unique ... some loop count and fail
>     dead man's brake would also help
>   * a new -m arg to indicate reexec

I've been testing this with custom `pmlogger.service` (for primary logger) and `pmlogger@.service `(template for farm logger instances), as follows -:

== pmlogger.service ==
```
[Unit]
Description=Performance Metrics Archive Logger
Documentation=man:pmlogger(1)
After=network-online.target pmcd.service

[Service]
Type=notify
NotifyAccess=all
TimeoutSec=120
Restart=always
ExecStart=/usr/libexec/pcp/bin/pmlogger -P -r -c config.default -v 100mb -m reexec /var/log/pcp/pmlogger/goblin/%%Y%%0m%%0d.%%0H.%%0M
PIDFile=/run/pcp/pmlogger.pid

[Install]
WantedBy=multi-user.target
```

And the template == pmlogger@.service ==
```
[Unit]
Description=Performance Metrics Archive Logger
Documentation=man:pmlogger(1)
After=network-online.target pmcd.service

[Service]
Type=notify
NotifyAccess=all
TimeoutSec=120
Restart=always
ExecStart=/usr/libexec/pcp/bin/pmlogger -h %i -r -c config.%i -v 100mb -m reexec -l/var/log/pcp/pmlogger/%i/pmlogger.log /var/log/pcp/pmlogger/%i/%%Y%%0m%%0d.%%0H.%%0M
PIDFile=/run/pcp/pmlogger-%i.pid

[Install]
WantedBy=multi-user.target
```

Obviously these aren't using any rc scripts - just for testing - and have no control file integration. To add a new farm logger : `systemctl enable pmlogger@somehost` and then start that service. These all manifest as separately managed service units, so we can start/stop individual loggers (including the primary), and on reboot they all start (if enabled) in parallel.


kmcdonell:
The missing library for Debian is -lsystemd although this needs wrapping in a macro that is conditionally set via configure (like $(LIB_FOR_PTHREADS)).

goodwinos:
> I don't see the point for the %0x instead of %x for the strftime pattern ... we currently use %Y%m%d.%H.%M (no zeroes) which I think will work just fine.

with %Y%m%d.%H.%M the template was longer than the resulting string, which upset the argv[argc-1] munging. So I inserted the 0's to pad it out.

> 
> For Debian-based systems, libsystemd-dev is marked (build optional) in qa/other-packages/manifest, but the build fails with Unmet build dependencies: libsystemd-dev ... either it is not optional for Debian-based systems, or the build needs to be fixed ... I suspect the latter, because libsystemd-dev will not be available for older (but still supported) Debian-based releases. 

Andreas fixed the debian library issue last night (see fd28dc287f300af2a058444ede0449407999f14a). He also updated qa/admin/other-packages/manifest

> There may be a similar issue for RPM builds, but I've not investigated this.

I haven't encountered any

> 
> On Ubuntu, the build fails for pmlogger with /usr/bin/ld: ../../../src/libpcp/src/libpcp.so: undefined reference to `sd_notify' ... there is some -l glue missing
> 
> I'll add more comments as I work through the build and testing.

Please note I've committed a few more updates to this PR (in the goodwinos:pmlogger-reexec branch)

kmcdonell:
SIGUSR1 handler does not seem to be installed properly.  I started pmlogger, checked the pid, did 
$ sudo kill -USR1 <pid>
and a new pmlogger process was running.  At the end of the pmlogger.log.prior file I see: 
Caught signal 10, exiting
Note the man page says SIGUSR1, but the code appears to be using SIGUSR2 (which does work sort of as expected, other than these errors in pmlogger.log:
__pmBind(/run/pcp/pmlogger.32564.socket): Address already in use
__pmBind(/var/lib/pcp/.pcp/run/pmlogger.32564.socket): No such file or directory
pmlogger: error creating primary logger symbolic link /var/lib/pcp/tmp/pmlogger/primary: File exists
pmlogger: error creating primary logger socket symbolic link /run/pcp/pmlogger.primary.socket: Bad address

kmcdonell:
Andreas's changes to not fix my issues above ... there is no LIB_FOR_SYSTEMD being setup and used.
The "add some number of 0s" is a gross hack (and not POSIX strftime compliant), there must be a cleaner solution, surely.
I'm pulling from your branch as soon as I see commits.

goodwinos:
> SIGUSR1 handler does not seem to be installed properly. I started pmlogger, checked the pid, did
> $ sudo kill -USR1
> and a new pmlogger process was running. At the end of the pmlogger.log.prior file I see:
> Caught signal 10, exiting
> Note the man page says SIGUSR1, but the code appears to be using SIGUSR2 (which does work sort of as expected, 

Changed to SIGUSR2 in 1cf9981484b310759120229e2baaa7317e59f57c (which included updates to the man page). pmnewlog uses SIGUSR1 to kill off the existing logger.

other than these errors in pmlogger.log after the reexec:
> __pmBind(/run/pcp/pmlogger.32564.socket): Address already in use
> __pmBind(/var/lib/pcp/.pcp/run/pmlogger.32564.socket): No such file or directory
> pmlogger: error creating primary logger symbolic link /var/lib/pcp/tmp/pmlogger/primary: File exists
> pmlogger: error creating primary logger socket symbolic link /run/pcp/pmlogger.primary.socket: Bad address

Running pmlogger as as root, I haven't been seeing the __pmBind errors/warnings. But running as non-root I do now. and qa/1896.full shows them too. The intent was to use the same sockets, FDs and pid files .. I'll investigate more

Also, https://pubs.opengroup.org/onlinepubs/9699919799/functions/strftime.html says 0 (zero) is a valid flag preceding a conversion character, but maybe that's not current posix?

kmcdonell:
Hi Mark ... wish we could get in the same room to discuss, but that's not possible at the moment.
- pmlogger runs as user pcp, so that's why I'm seeing the _pmBind errors/warnings
- pmnewlog will go away, so we could revert to SIGUSR1
- the ubuntu man page for strftime is misleading ... but I think %Y%0m%0d.%0H.%0M _only_ works because the %0m adds 1 char to be used in %Y, %0d adds another character to be used in %Y and the %0H and %0M add 2 characters of pad ... this is pretty obtuse!

kmcdonell:
I don't think the -m command line argument is getting changed to to "reexec" before the exec() ... in the ps output mine still appears as -m pmlogger_check after SIGUSR2.
And the logic here:
```
else if (note && strncmp(note, "reexec", 6) != 0)
        break; /* archive already exists and we're NOT using -m reexec */
```
seems incorrect ... I don't think there should be any conditional execution based on the -m argument "note", as this is purely for annotation when inspecting ps output.

And I've just found another wrinkle ... the Latest folio does not get updated if pmlogger gets a SIGUSR2 ... I think this needs to be added to the set of things pmlogger needs to do when started with one of the -m options that indicate this is launched via pmlogger_check or reexec.

I have changes for pmlogger_daily that rip out all of the convoluted logic around pmnewlog and just sends the signal ... we'll see how this goes tonight.

goodwinos:
> Hi Mark ... wish we could get in the same room to discuss, but that's not possible at the moment.
> 
> * pmlogger runs as user pcp, so that's why I'm seeing the _pmBind errors/warnings

maybe that would be if bind returned EACCESS. But I was getting EADDRINUSE but only when run as non-root user. This was because after the exec, the socket file still existed. So for my case, this patch fixed it :
```
diff --git a/src/pmlogger/src/ports.c b/src/pmlogger/src/ports.c
index 3e1b5771c..e303fbe83 100644
--- a/src/pmlogger/src/ports.c
+++ b/src/pmlogger/src/ports.c
@@ -311,6 +311,8 @@ GetPorts(char *file)
                    if (sts >= 0 || oserror() == EEXIST) {
                        __pmSockAddrSetPath(myAddr, socketPath);
                        __pmServerSetLocalSocket(socketPath);
+                       if (pmlogger_reexec && access(socketPath, F_OK) == 0)
+                           unlink(socketPath);
                        sts = __pmBind(fd, (void *)myAddr, __pmSockAddrSize());
```

I have changes to qa/1896 that test this patch (above).

> * pmnewlog will go away, so we could revert to SIGUSR1

OK I'll revert back to USR1 (once pmnewlog has gone away). How did the overnight testing go with those changes?

> * the ubuntu man page for strftime is misleading ... but I think %Y%0m%0d.%0H.%0M _only_ works because the %0m adds 1 char to be used in %Y, %0d adds another character to be used in %Y and the %0H and %0M add 2 characters of pad ... this is pretty obtuse!

have the 0 just for %0H and %0M ends up with the same length string after strftime, vis 
```
$ x=qa-%Y%m%d.%0H.%0M; pmdate $x; echo $x
qa-20200327.10.52
qa-%Y%m%d.%0H.%0M
```

I'll also fix the -m reexec issues

natoscott:
> > Hi Mark ... wish we could get in the same room to discuss, but that's not possible at the moment.
> > 
> > * pmlogger runs as user pcp, so that's why I'm seeing the _pmBind errors/warnings
> 
> maybe that would be if bind returned EACCESS. But I was getting EADDRINUSE but only when run as non-root user. This was because after the exec, the socket file still existed.

Sounds like pmlogger cleanup() is not being run before the exec?  (atexit handler not being run)  Looks like there's other cleanup that may need to be done too, not just unlinking the unix domain socket.

> So for my case, this patch fixed it :

So not sure that's a correct fix - seems more like a workaround?
 
cheers.

goodwinos:
this is all merged into master - deleting the branch. I have another change for pmlogger to write out the 'Latest' folio on startup (including after reexec) just prior to entering the main loop, but that can go directly into master.

PR #879, opened 2020-04-11
pmproxy: return authentication errors

andreasgerstmayr:
had to open this PR before I forget my findings after easter ;)

Looks like after creating a new context (`pmwebapi_new_context`) the first PDU pmproxy sends to the PMCD with authentication credentials comes from the `pmGetContextHostName_r` function (`pmLookupName_ctx` inside this function), which gets called from `pmwebapi_new_context` -> `pmwebapi_source_meta`.

Unfortunately this function doesn't return the `sts`, so pmproxy doesn't know about any authentication errors. This PR introduces `pmGetContextHostName_rr` which returns the `sts`, and fixes QA 1388.

Another issue remains however, valgrind reports the following when the authentication is unsuccessful:
```
==459557== Invalid write of size 8
==459557==    at 0x4C51FA3: uv_run (in /usr/lib64/libuv.so.1.0.0)
==459557==    by 0x112411: main_loop (server.c:856)
==459557==    by 0x11194E: main (pmproxy.c:427)
==459557==  Address 0x57f13f0 is 176 bytes inside a block of size 360 free'd
==459557==    at 0x483AA0C: free (vg_replace_malloc.c:540)
==459557==    by 0x48AB593: webgroup_new_context (webgroup.c:207)
==459557==    by 0x48AB593: webgroup_lookup_context (webgroup.c:260)
==459557==    by 0x48ACA4E: pmWebGroupContext (webgroup.c:299)
==459557==    by 0x4C4DBCD: ??? (in /usr/lib64/libuv.so.1.0.0)
==459557==    by 0x4C7D4E1: start_thread (in /usr/lib64/libpthread-2.30.so)
==459557==    by 0x4E406A2: clone (in /usr/lib64/libc-2.30.so)
==459557==  Block was alloc'd at
==459557==    at 0x483BB1A: calloc (vg_replace_malloc.c:762)
==459557==    by 0x48AB2CE: webgroup_new_context (webgroup.c:162)
==459557==    by 0x48AB2CE: webgroup_lookup_context (webgroup.c:260)
==459557==    by 0x48ACA4E: pmWebGroupContext (webgroup.c:299)
==459557==    by 0x4C4DBCD: ??? (in /usr/lib64/libuv.so.1.0.0)
==459557==    by 0x4C7D4E1: start_thread (in /usr/lib64/libpthread-2.30.so)
==459557==    by 0x4E406A2: clone (in /usr/lib64/libc-2.30.so)
==459557== 
```

PR #882, opened 2020-04-15
pmproxy: return authentication errors

andreasgerstmayr:
somehow I messed up the last PR (#879) and set the base branch of that PR to the master branch of my fork, instead of the `auth-changes` feature branch of my fork... here is it again:

Looks like after creating a new context (`pmwebapi_new_context`) the first PDU pmproxy sends to the PMCD with authentication credentials comes from the `pmGetContextHostName_r` function (`pmLookupName_ctx` inside this function), which gets called from `pmwebapi_new_context` -> `pmwebapi_source_meta`.

Unfortunately this function doesn't return the `sts`, so pmproxy doesn't know about any authentication errors. This PR introduces `pmGetContextHostName_rr` which returns the `sts`, and fixes QA 1388.

Another issue remains however, valgrind reports the following when the authentication is unsuccessful:
```
==459557== Invalid write of size 8
==459557==    at 0x4C51FA3: uv_run (in /usr/lib64/libuv.so.1.0.0)
==459557==    by 0x112411: main_loop (server.c:856)
==459557==    by 0x11194E: main (pmproxy.c:427)
==459557==  Address 0x57f13f0 is 176 bytes inside a block of size 360 free'd
==459557==    at 0x483AA0C: free (vg_replace_malloc.c:540)
==459557==    by 0x48AB593: webgroup_new_context (webgroup.c:207)
==459557==    by 0x48AB593: webgroup_lookup_context (webgroup.c:260)
==459557==    by 0x48ACA4E: pmWebGroupContext (webgroup.c:299)
==459557==    by 0x4C4DBCD: ??? (in /usr/lib64/libuv.so.1.0.0)
==459557==    by 0x4C7D4E1: start_thread (in /usr/lib64/libpthread-2.30.so)
==459557==    by 0x4E406A2: clone (in /usr/lib64/libc-2.30.so)
==459557==  Block was alloc'd at
==459557==    at 0x483BB1A: calloc (vg_replace_malloc.c:762)
==459557==    by 0x48AB2CE: webgroup_new_context (webgroup.c:162)
==459557==    by 0x48AB2CE: webgroup_lookup_context (webgroup.c:260)
==459557==    by 0x48ACA4E: pmWebGroupContext (webgroup.c:299)
==459557==    by 0x4C4DBCD: ??? (in /usr/lib64/libuv.so.1.0.0)
==459557==    by 0x4C7D4E1: start_thread (in /usr/lib64/libpthread-2.30.so)
==459557==    by 0x4E406A2: clone (in /usr/lib64/libc-2.30.so)
==459557== 
```

### DIFF
--- a/qa/1388
+++ b/qa/1388
@@ -153,39 +153,51 @@ _service pmproxy start >>$here/$seq.full 2>&1
 echo "Start test Python PMDA to check if username is in per-context state"
 cd pmdas/test_python
 $sudo ./Install </dev/null \
-    | _filter_pmda_install \
-    | sed -e '/^Waiting for pmcd/s/\.\.\.[. ]*$/DOTS/' \
-    | $PCP_AWK_PROG '
-/Check test_python metrics have appeared/ { if ($10 >= 0 && $10 <= 9) $10 = "N" }
-                                          { print }'
+| _filter_pmda_install
 cd $here
 
-echo && echo "=== pmproxy no authentication (expect errors) ==="
-curl -s -S "http://localhost:44322/pmapi/context?hostspec=localhost%3Fusername%3Dbob" | pmjson | _filter_json
-curl -s -S "http://localhost:44322/pmapi/context?hostspec=localhost%3Fpassword%3Dbib" | pmjson | _filter_json
-curl -s --user $username:bob "http://localhost:44322/pmapi/context?hostspec=localhost%3Fusername%3Dbob" | pmjson | _filter_json
+echo && echo "=== invalid username with pminfo ==="
+pminfo -f -h "pcp://127.0.0.1?username=nonexisting_user&password=bob" test_python.current_username
 
-echo && echo "=== pmproxy with authentication (expect success) ==="
-curl -s --user $username:y "http://localhost:44322/pmapi/context" | pmjson | _filter_json | _filter_credentials
-
-echo && echo "=== state without authentication ==="
-ctx_no_auth=`curl -s "http://localhost:44322/pmapi/context" | pmpython -c 'import sys,json; print(json.load(sys.stdin)["context"])'`
-curl -s "http://localhost:44322/pmapi/$ctx_no_auth/fetch?names=test_python.current_username" | pmjson | grep '"value"' | _filter_ctx
-
-echo && echo "=== state with authentication ==="
-ctx_with_auth=`curl -s --user $username:y "http://localhost:44322/pmapi/context" | pmpython -c 'import sys,json; print(json.load(sys.stdin)["context"])'`
-curl -s "http://localhost:44322/pmapi/$ctx_with_auth/fetch?names=test_python.current_username" | pmjson | grep '"value"' | _filter_ctx | _filter_username
-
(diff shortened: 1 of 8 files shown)


natoscott:
@andreasgerstmayr looks good, nice work!  I think I see the valgrind issue - its complaining about an address that is smack bang in the middle of a recently freed context structure, so I suspect that's the uv_timer_t inside our context_t (load.h line 48).  Which would mean we've registered the timer in new context and then its accessed back in uv_run() after freeing the context on auth failure.  I'll push a follow-up commit to yours to tackle this one.

In terms of the name pmGetContextHostName_rr - I'm not a big fan of the '_rr' :) - how about we go back to basics and call this (finally final!) API version pmGetHostName(3)?  Either way, we need a man page update here too.

Finally, this commit seems to have some unrelated changes to pmdatest_python.python in QA?  Could you double check thats supposed to be there?  Thanks.  (I've merged, so go right ahead and undo that part if its in error)

andreasgerstmayr:
Thanks for the review & merge!

> In terms of the name pmGetContextHostName_rr - I'm not a big fan of the '_rr' :) - how about we go back to basics and call this (finally final!) API version pmGetHostName(3)? Either way, we need a man page update here too.

Is it fine to change the signature of an existing API function? (`pmGetHostName_r` was created to not break the existing API, that's why I added a `pmGetHostName_rr` :D)

> Finally, this commit seems to have some unrelated changes to pmdatest_python.python in QA? Could you double check thats supposed to be there? Thanks. (I've merged, so go right ahead and undo that part if its in error)

Yes, that's supposed to be here - the QA creates multiple contexts and checks associated attributes (the auth username), so the internal mapping from a context to an attribute should be cleared when a context ends, otherwise we end up with an inconsistent state.

natoscott:
> > In terms of the name pmGetContextHostName_rr - I'm not a big fan of the '_rr' :) - how about we go back to basics and call this (finally final!) API version pmGetHostName(3)? Either way, we need a man page update here too.
> 
> Is it fine to change the signature of an existing API function? (`pmGetHostName_r` was created to not break the existing API, that's why I added a `pmGetHostName_rr` :D)

It's not fine to change signature of existing functions (but that's not what I suggested), we must keep existing APIs as-is for any third-party tools using the library.  As we're closing in on QA week, I've gone ahead and made the change I suggested with the alternate name (pmGetHostName) and updated docs, etc.

> > Finally, this commit seems to have some unrelated changes
> Yes, that's supposed to be here

Great, thanks.

andreasgerstmayr:
> > Is it fine to change the signature of an existing API function? (`pmGetHostName_r` was created to not break the existing API, that's why I added a `pmGetHostName_rr` :D)
> 
> It's not fine to change signature of existing functions (but that's not what I suggested), we must keep existing APIs as-is for any third-party tools using the library. 

Ah sorry, I didn't realize that the old functions have `context` inside the name, and the new suggested function name doesn't. Now it makes sense.

> As we're closing in on QA week, I've gone ahead and made the change I suggested with the alternate name (pmGetHostName) and updated docs, etc.

Thanks!

PR #892, opened 2020-04-22
Fix for new version of ActiveMQ broker

liufan:
In the new version of AcrtiveMQ (5.12+), when sending REST request through Jolokia API, it requires an 'Origin' value in HTTP request header(https://activemq.apache.org/rest) otherwise it will fail.
$ curl -u admin:admin http://localhost:8161/api/jolokia/read/org.apache.activemq:brokerName=localhost,destinationName=NOTIFICATIONS,destinationType=Queue,type=Broker
{"error_type":"java.lang.Exception","error":"java.lang.Exception : Origin null is not allowed to call this agent","status":403}

### DIFF
--- a/src/pmdas/activemq/RESTClient.pm
+++ b/src/pmdas/activemq/RESTClient.pm
@@ -41,7 +41,8 @@ sub get {
 	$json = $cached_json;
     }
     else {
-	my $response = $self->{_http_client}->get("http://" . $self->{_host} . ":" . $self->{_port} . $url);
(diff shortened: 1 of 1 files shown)


goodwinos:
Hi Liufan,

is this back-compatible with the current and previous version of ActiveMQ broker (< 5.12) ? If so, we'll merge it for the pcp-5.1.0 release.  Thanks for the contribution!

liufan:
Hi Mark,
Yes, I've tested it locally. It worked with the previous version of ActiveMQ broker.

PR #902, opened 2020-05-05
Export openvswitch metrics into pcp

not4win:
Export openvswitch bridge,port, flow metrics into pcp


qa/1680
```diff
@@ -0,0 +1,68 @@
...
+
+    echo
+    echo "=== openvswitch agent installation ==="
+    $sudo ./Install </dev/null >$tmp.out 2>&1
+    # Check metrics have appeared ... X metrics and Y values
+    _filter_pmda_install <$tmp.out \
+    | sed \
+        -e '/^Waiting for pmcd/s/\.\.\.[. ]*$/DOTS/' \
+    | $PCP_AWK_PROG '
+/Check rabbitmq metrics have appeared/  { if ($7 == 22) $7 = "X"
```
natoscott, line 50: rabbitmq?  ;-)

src/pmdas/openvswitch/Remove
```diff
@@ -0,0 +1,34 @@
...
+# under the terms of the GNU General Public License as published by the
+# Free Software Foundation; either version 2 of the License, or (at your
+# option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+# for more details.
+#
+# Remove the rabbitmq PMDA
```
natoscott, line 15: Should be openvswitch here too

src/pmdas/openvswitch/pmdaopenvswitch.1
```diff
@@ -0,0 +1,57 @@
...
+.\"
+.\" This program is distributed in the hope that it will be useful, but
+.\" WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+.\" or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+.\" for more details.
+.\"
+.\"
+.TH PMDAopenvswitch 1 "PCP" "Performance Co-Pilot"
+.SH NAME
+\f3pmdaopenvswitch\f1 \- Rabbit Message Queue subsystem PMDA
```
natoscott, line 18: Rabbit MQ cut+paste again

not4win:
Ah well, changes done. Just for testing purposes: Install openvswitch with:
`sudo dnf install openvswitch`
Enable connection:
`sudo /usr/share/openvswitch/scripts/ovs-ctl start`
Add a temp bridge with:
`sudo ovs-vsctl add-br mybridge`

src/pmdas/openvswitch/pmdaopenvswitch.py
```diff
@@ -0,0 +1,184 @@
...
+        insts = []
+        for idx, val in enumerate(self.bridge_names):
+            insts.append(pmdaInstid(idx, val))
+        self.add_indom(pmdaIndom(self.bridge_indom, insts))
+
+    def openvswitch_refresh(self, cluster):
+        """refresh function"""
+        if cluster == self.bridge_cluster:
+            # self.get_bridge_info_json()
+            insts = []
```
not4win, line 119: As of now just facing a bit of a hiccup here; Gives out an unknown error while uncommenting this.

src/pmdas/openvswitch/pmdaopenvswitch.py
```diff
@@ -0,0 +1,412 @@
...
+from pcp.pmda import PMDA, pmdaMetric, pmdaIndom, pmdaInstid
+from pcp.pmapi import pmUnits
+from pcp.pmapi import pmContext as PCP
+from cpmapi import PM_TYPE_STRING, PM_TYPE_U64, PM_TYPE_FLOAT
+from cpmapi import PM_SEM_COUNTER, PM_SEM_DISCRETE, PM_SEM_INSTANT
+from cpmapi import PM_COUNT_ONE, PM_SPACE_BYTE, PM_TIME_SEC
+from cpmapi import PM_ERR_APPVERSION
+from cpmda import PMDA_FETCH_NOVALUES
+
+PMDA_DIR = '/home/ashwin/github/pcp/src/pmdas/openvswitch'
```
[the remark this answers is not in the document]
not4win: Oh totally missed this one out!.Thanks!

src/pmdas/openvswitch/pmdaopenvswitch.py
```diff
@@ -148,7 +148,7 @@ def __init__(self, name, domain):
 
     def fetch_switch_info(self):
         """ fetches result from command line """
-        out = subprocess.Popen([self.pmda_dir+'/switch.sh'], stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+        out = subprocess.Popen([self.pmda_dir+'/openvswitch/switch.sh'], stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
Erbenos, line 151: One more thing I noticed - I am nitpicking here - it might be worthwhile to use OS specific separator in file paths where it's relatively less time intensive to implement like here. I am not sure about state of PCP on Windows, but making the agent little bit future proof shouldn't do any harm. Opinion @natoscott ?

not4win: Ah thanks again. I didn't think of this. Will resolve ASAP.

src/pmdas/openvswitch/switch.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+sudo ovs-vsctl --format=json list bridge
```
natoscott, line 3: This script needs to be removed, ovs-vsctl (and options) should be invoked directly by the PMDA.
As a general point, we should not assume sudo is available and configured in a way we can use it - the PMDA has root privileges by default so this is not an issue.

src/pmdas/openvswitch/pmdaopenvswitch.py
```diff
@@ -0,0 +1,411 @@
...
+                            self.flow_metrics[item][4],
+                            self.flow_metrics[item][4])
+
+
+        self.set_fetch_callback(self.openvswitch_fetch_callback)
+        self.set_refresh(self.openvswitch_refresh)
+
+    def fetch_switch_info(self):
+        """ fetches result from command line """
+        out = subprocess.Popen([os.path.join(PMDA_DIR, 'openvswitch', 'switch.sh')], stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
natoscott, line 151: The PMDA should invoke commands it needs directly wherever possible, not via a shell script.

src/pmdas/openvswitch/pmdaopenvswitch.py
```diff
@@ -0,0 +1,411 @@
...
+        stdout, stderr = out.communicate()
+        stdout = stdout.decode("utf-8")
+        return stdout, stderr
+
+    def get_switch_info_json(self):
+        """ Convert the commandline output to json """
+        stdout, _ = self.fetch_switch_info()
+
+        # Store it in a file
+        text_file = open("switch_output.txt", "w")
```
natoscott, line 161: I'm not understanding the need for this intermediate file - and its creation can lead to problems (e.g. ENOSPC) we don't want to have to deal with - ideally, just read directly from the ovs programs stdout.

src/pmdas/openvswitch/pmdaopenvswitch.py
```diff
@@ -0,0 +1,411 @@
...
+                                       self.flow_indom,
+                                       self.flow_metrics[item][2],
+                                       self.flow_metrics[item][3]),
+                            self.flow_metrics[item][4],
+                            self.flow_metrics[item][4])
+
+
+        self.set_fetch_callback(self.openvswitch_fetch_callback)
+        self.set_refresh(self.openvswitch_refresh)
+
```
natoscott, line 148: Since we're using compound instances here, we should add a labels callback which provides the labels for each component of the instance name - see simple/pmdasimple.python for example, and we'll need lines here like:
self.set_label(self.openvswitch_label)
self.set_label_callback(self.openvswitch_label_callback)

PR #912, opened 2020-05-16
Simple idea how to store timeseries replies from Redis

ElephantZhang:
To extend time-series query language with RATE() and ADD() functions, this PR is a very first thing I think about how to deal with the replies from Redis. (NOTE: the following code still has bugs, I will continue to fix them.) The following modified code part is to store timeseries values returned from Redis into the corresponding nodes in the parser tree.

A query statement will be transformed into a parser tree consist of struct node_t, then we will take processing on it. Since all the queries to Redis are async and I think there is no simple method to ensure the order of points' queries is correct according to the parser tree. Take an example, a query statement `metric.A + (metric.B + metric.C)[count:10]`, when we have the instances' values of A and going to compute them with B&C's, how can we know the instances' values of B&C's have been obtained? And there are other issues, assume that we do not use functions implemented in Redis, we request to Redis for values of `metric.A`, then `metric.B`, but when the reply to metric.A arrive, after the callback function being called we can never access the replies' values unless we save them somewhere.

That's the reason why I add `void **SID` and `pmSeriesValue ***series_values` into `struct node_t`. For each function-type node, use its child nodes' information to request to Redis and store the corresponding instance values into child nodes. And for the further functions' implementation, it's very convenient because for a function-type node, the values required for calculation have already existed in its child nodes.

### DIFF
--- a/src/libpcp_web/src/query.c
+++ b/src/libpcp_web/src/query.c
@@ -1389,6 +1389,353 @@ series_query_expr(void *arg)
     series_query_end_phase(baton);
 }
 
+static int
+series_instance_store_to_node(seriesQueryBaton *baton, sds series,
+	pmSeriesValue *value, int nelements, redisReply **elements, node_t *np, int idx_sample)
+{
+    char		hashbuf[42];
+    sds			inst;
+    int			i, sts = 0, idx_instance = 0, idx_series = np->result.nseries;
+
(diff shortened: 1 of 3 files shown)


natoscott:
@ElephantZhang let's continue our discussion on this topic over in #redis on pcp slack

PR #921, opened 2020-05-19
libpcp: open NSSDB readonly for clients

andreasgerstmayr:
I ran the tests from the secure group, they all pass.
The code doesn't change the pmproxy part (with the comment regarding read/write), which is in `secureserver.c`.


andreasgerstmayr:
ahkr, while testing this (using `make`/`sudo make install` after changes), the pmlogger executable gets the unconfined domain and is allowed to do anything... so any change I made "fixed" the issue...
but when building updated rpm packages and installing them, the selinux policy gets enforced and the access to PMCD doesn't work anymore.

After manually resetting the correct domain after make install, I've found another issue, the call to `__pmMakePath` which silently fails and subsequently `NSS_Init` isn't called and no certificates are loaded. I've amended the commit and pushed again, now it's fixed (and my selinux knowledge got updated :P)

PR #931, opened 2020-05-27
Make sure mkafm commands "check" and "remove" can work with compressed archives

kurik:
Make sure mkafm commands "check" and "remove" can work with compressed archives


kurik:
Solves https://bugzilla.redhat.com/show_bug.cgi?id=1839666

goodwinos:
The proposed fix looks good to me and the update to qa/671 exercises the fix. No other regressions in any other 'folio' group tests. Thanks Jan.

PR #941, opened 2020-06-08
Fix of 1845241 - Intermittent pmlogconf core dumps

kurik:
Variable `i` is compared with `vsp->numval` in iterrations of `evaluate_number_values`, `evaluate_string_values` and `evaluate_string_regexp` functions. As `vsp->numval` can have a negative value (an error code), the comparison if `i`is less then `vsp->numval` needs to use signed integers instead of unsigned ones.


natoscott:
@kurik thanks Jan, I've extended test qa/362 with additional error handling cases too.

Issue #922, opened 2020-05-21
 ENDATE/ENTIME of pcp atop -v is not implemented


shogo-matsumoto:
I am trying to implement ENDATE/ENTIME of pcp-atop as a related issue of [PR#860](https://github.com/performancecopilot/pcp/pull/860).
But I haven't found the best implementation for it. Could anyone give me a comment if there is a good idea?

man pcp-atop describes ENDATE/ENTIME as follows:

```
ENDATE   Date that the process has been finished.  If  the  process  is
         still running, this field shows `active'.

ENTIME   Time  that  the  process has been finished.  If the process is
         still running, this field shows `active'.
```

Original [atop](https://github.com/Atoptool/atop) command uses `acct` system call to get ENDATE/ENTIME.
`acct` has the following features:

- root privileges is needed to call 
- enable only one file at a time in the system

If pcp calls `acct`, it will disable `acct` already executed by another process, and vice versa.
The easy way I think to avoid the interference of `acct` among processes is to make a new PMDA which reads data from /var/account/pacct.
/var/account/pacct is generated and updated by psacct service if the service is enabled in the environment.

The problem of this idea is that /var/account/pacct file can't be read from non-root user due to its permission (0600).
Even if it is possible to run the PMDA as a root process, I'm concerned about whether it is appropriate
to collect data and pass it to non-root pcp commands while the data of /var/account/pacct is
originally accessed only from root user.

I thought other ideas:

- do not implement ENDATE/ENTIME and leave it for compatibility with (original) [atop](https://github.com/Atoptool/atop) command, or
- approximate ENDATE/ENTIME by the last sampling time when the process was active

I haven't come to the best conclusions.

natoscott:
> [...]
> Original [atop](https://github.com/Atoptool/atop) command uses `acct` system call to get ENDATE/ENTIME.
> `acct` has the following features:
> 
>     * root privileges is needed to call
>     * enable only one file at a time in the system
> 
> If pcp calls `acct`, it will disable `acct` already executed by another process, and vice versa.

Sounds like this is also true of the original 'atop'.  This is similar to how some hardware CPU counters work, it's not a problem (PCP actually helps, since it makes the values available to any clients).

> The easy way I think to avoid the interference of `acct` among processes is to make a new PMDA which reads data from /var/account/pacct.

Makes sense (alternatively, the existing 'proc' PMDA could be extended to export the new metrics)

[...]

PR #962, opened 2020-07-06
pmdaproc: add acct metrics

shogo-matsumoto:
Collect accounting information from process accounting file (see acct(5)).
pmdaproc(acct) will try to read data from /var/account/pacct if it is effective,
or use private accounting file (/tmp/pcp-pacct).
Currently version 3 format (struct acct_v3) is available.

Related to https://github.com/performancecopilot/pcp/issues/922

### DIFF
--- a/qa/031.out.linux
+++ b/qa/031.out.linux
@@ -65,6 +65,7 @@ children of "sample.longlong" ...
 pmid: PM_ID_NULL name: bozo.the.clown
 
 children of "" ...
+    acct                <s = 1>
     disk                <s = 1>
     event               <s = 1>
     filesys             <s = 1>
--- a/src/libpcp/src/hash.c
+++ b/src/libpcp/src/hash.c
@@ -126,6 +126,7 @@ __pmHashDel(unsigned int key, void *data, __pmHashCtl *hcp)
 	    else
 		lhp->next = hp->next;
 	    free(hp);
+	    hcp->nodes--;
 	    return 1;
 	}
 	lhp = hp;
--- a/src/pmdas/linux_proc/GNUmakefile
+++ b/src/pmdas/linux_proc/GNUmakefile
@@ -29,7 +29,7 @@ CFILES		= pmda.c cgroups.c proc_pid.c proc_runq.c proc_dynamic.c\
 		  getinfo.c contexts.c gram_node.c config.c error.c hotproc.c
 
 HFILES		= clusters.h indom.h \
-		  cgroups.h proc_pid.h proc_runq.h getinfo.h contexts.h hotproc.h gram_node.h config.h
+		  acct.h cgroups.h proc_pid.h proc_runq.h getinfo.h contexts.h hotproc.h gram_node.h config.h
 
 LFILES		= lex.l
 YFILES		= gram.y
--- a/src/pmdas/linux_proc/GNUmakefile
+++ b/src/pmdas/linux_proc/GNUmakefile
@@ -25,7 +25,7 @@ PMDAINIT	= proc_init
 PMDADIR		= $(PCP_PMDAS_DIR)/$(IAM)
 CONF_LINE	= "proc	3	pipe	binary		$(PMDADIR)/$(CMDTARGET) -d 3"
 
-CFILES		= pmda.c cgroups.c proc_pid.c proc_runq.c proc_dynamic.c\
+CFILES		= pmda.c acct.c cgroups.c proc_pid.c proc_runq.c proc_dynamic.c\
 		  getinfo.c contexts.c gram_node.c config.c error.c hotproc.c
 
 HFILES		= clusters.h indom.h \
@@ -103,6 +103,7 @@ proc_kernel_ulong.conf proc_jiffies.conf proc_kernel_ulong_migrate.conf: mk.rewr
 
 gram.tab.o:	hotproc.h
 lex.o:	hotproc.h
+acct.o pmda.o: acct.h
 cgroups.o pmda.o: clusters.h
 cgroups.o pmda.o:	cgroups.h
 cgroups.o pmda.o proc_pid.o proc_runq.o proc_dynamic.o:	proc_pid.h
@@ -113,7 +114,7 @@ pmda.o:	domain.h
 pmda.o:	getinfo.h
 pmda.o:	$(VERSION_SCRIPT)
 
-cgroups.o contexts.o pmda.o proc_dynamic.o proc_pid.o proc_runq.o:	$(TOPDIR)/src/include/pcp/libpcp.h
+acct.o cgroups.o contexts.o pmda.o proc_dynamic.o proc_pid.o proc_runq.o:	$(TOPDIR)/src/include/pcp/libpcp.h
 
 check::	$(CFILES) $(HFILES)
 	$(CLINT) $^
--- a/src/pmdas/linux_proc/acct.c
+++ b/src/pmdas/linux_proc/acct.c
@@ -17,7 +17,10 @@
 #include <sys/wait.h>
 #include "acct.h"
 
+#define MAX_ACCT_RECORD_SIZE_BYTES	128
 #define RINGBUF_SIZE			5000
+#define OPEN_RETRY_INTERVAL		60
+#define CHECK_ACCOUNTING_INTERVAL	600
 #define PACCT_SYSTEM_FILE		"/var/account/pacct"
 #define PACCT_PCP_PRIVATE_FILE		"/tmp/pcp-pacct"
 
@@ -236,6 +239,76 @@ static int open_pacct_file(void) {
 	return 0;
 }
 
+static void reopen_pacct_file(void) {
+	close_pacct_file();
+	open_pacct_file();
+}
+
+static void free_entry(__pmHashCtl *hp, int i_inst) {
+	__pmHashNode *node = __pmHashSearch(i_inst, hp);
+	if (node && node->data) {
+		__pmHashDel(i_inst, (void *)node->data, hp);
+		free(node->data);
+	}
+}
+
+static int free_ringbuf_entry(__pmHashCtl *hp, int index) {
+	if (!acct_ringbuf.buf[index].instid.i_inst)
+		return 0;
+	free_entry(hp, acct_ringbuf.buf[index].instid.i_inst);
+	memset(&acct_ringbuf.buf[index], 0, sizeof(acct_ringbuf_entry_t));
+	return 1;
+}
+
+static int next_ringbuf_index(int index) {
+	return (index + 1) % RINGBUF_SIZE;
+}
+
+static void acct_ringbuf_add(__pmHashCtl *hp, acct_ringbuf_entry_t *entry) {
+	free_ringbuf_entry(hp, acct_ringbuf.next_index);
+	acct_ringbuf.buf[acct_ringbuf.next_index] = *entry;
+	acct_ringbuf.next_index = next_ringbuf_index(acct_ringbuf.next_index);
+}
+
+static int ringbuf_entry_is_valid(time_t t, int index) {
+	return (t - acct_ringbuf.buf[index].time) <= acct_lifetime;
+}
+
+static int acct_gc(__pmHashCtl *hp, time_t t) {
+	int need_update = 0;
+	int i, index = acct_ringbuf.next_index;
+	for (i = 0; i < RINGBUF_SIZE; i++) {
+		if (ringbuf_entry_is_valid(t, index))
+			break;
+		need_update += free_ringbuf_entry(hp, index);
+		index = next_ringbuf_index(index);
+	}
+	return need_update;
+}
+
+static void copy_ringbuf_to_indom(pmdaIndom* indomp, time_t t) {
+	int i, index;
+	for (i = 0; i < RINGBUF_SIZE; i++) {
+		index = (acct_ringbuf.next_index - 1 - i + RINGBUF_SIZE) % RINGBUF_SIZE;
+		if (!ringbuf_entry_is_valid(t, index))
+			break;
+		indomp->it_set[i] = acct_ringbuf.buf[index].instid;
+	}
+	indomp->it_numinst = i;
+}
+
+static int get_file_size(const char* path) {
+	struct stat statbuf;
+	if (stat(path, &statbuf) < 0)
+		return -1;
+	return statbuf.st_size;
+}
+
+static int exists_hash_entry(int i_inst, proc_acct_t *proc_acct) {
+	__pmHashNode *node = __pmHashSearch(i_inst, &proc_acct->accthash);
+	return node && node->data ? 1 : 0;
+}
+
 void acct_init(proc_acct_t *proc_acct) {
 	init_acct_file_info();
 	open_pacct_file();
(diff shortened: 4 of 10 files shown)


natoscott:
@shogo-matsumoto this is a looking good.  I've made some minor cleanups (see below), so I'll merge your current work (later today) so that we can work together on these remaining items ...

- hertz - use existing variable (I've fixed this one)
- configuration variables (intervals, file size thresholds, paths) - better if external configurable not hard-coded (handy for testing), and perhaps even could be modified using pmStore(3) and perhaps proc.control.acct or acct.control.* metrics.
- QA - sample pacct file (compressed) that we could temporarily put in /var/account/pacct to test the PMDA?
     - perhaps add this file (perhaps several, different / interesting cases?) to the qa/linux/ pre-installed files set
     - also need 'live' test to check metrics work with the running kernel process accounting
- code style differs to rest of pmdaproc?  best if its consistent
- path for PACCT_PCP_PRIVATE_FILE needs to be somewhere more private than /tmp for security (/var/lib/pcp/tmp/pmcd/proc-pacct? /var/lib/pcp/tmp is also PCP_TMPFILE_DIR)
- PACCT_SYSTEM_FILE should be from PCP_ACCT_DIR/pacct and configure.ac like PCP_SA_DIR
- file size is not an int (get_file_size) - use off_t or size_t or unsigned long long
- acct_debug always does the varargs expansion and always incurs function call overhead -
  maybe use __pmNotifyErr (gives timestamps) with DEBUG level and move pmDebugOptions check
  to wrap each callsite (lowest overhead).
- what is the permissions model going to be?  not yet using PCP per-context authentication ... I suggest we go with the model of no-authentication == no access like rest of proc.* metrics.

natoscott:
@shogo-matsumoto correction:

>     * what is the permissions model going to be?  not yet using PCP per-context authentication ... I suggest we go with the model of no-authentication == no access like rest of proc.* metrics.

It *is* using per-context authentication, sorry, my mistake - although enforcing that only root can access these metrics values.  Should we allow any authenticated user, I wonder? (more like the proc.* metrics access)  If so, also the existing -A option to pmdaproc could be used to let a sysadmin allow all (incl. remote) access to the metrics, again consistent with proc.* metrics access.

```diff --git a/src/pmdas/linux_proc/pmda.c b/src/pmdas/linux_proc/pmda.c
index a100cc9e6..eb16153ba 100644
--- a/src/pmdas/linux_proc/pmda.c
+++ b/src/pmdas/linux_proc/pmda.c
@@ -1520,7 +1520,7 @@ proc_refresh(pmdaExt *pmda, int *need_refresh)
            refresh_cgroups2(cgroup, cgrouplen, need_refresh);
     }
 
-    if (need_refresh[CLUSTER_ACCT] && !proc_ctx_getuid(pmda->e_context))
+    if (need_refresh[CLUSTER_ACCT] && proc_ctx_getuid(pmda->e_context) >= 0)
        refresh_acct(&proc_acct);
 
     if (need_refresh[CLUSTER_PID_STAT] ||
```

Other items:
- selinux policy (I've started on this).
- man page updates for pmdaproc(1) to mention these metrics and the files used

If you want to chat directly to me about any of this (might be easier to coordinate our efforts that way?) please feel free to contact me on slack - access details on https://pcp.io/community.html

shogo-matsumoto:
@natoscott 

Thank you so much for reviewing my code and merging it! I understand the remaining items. (I'm afraid my progress may be slow but) I'll start with the easier of tasks which remains unfinished. The following items seem relatively easy for me.

- configuration variables 
- code style
- PACCT_PCP_PRIVATE_FILE
- PACCT_SYSTEM_FILE 
- file size

>Should we allow any authenticated user, I wonder? (more like the proc.* metrics access) If so, also the existing -A option to pmdaproc could be used to let a sysadmin allow all (incl. remote) access to the metrics, again consistent with proc.* metrics access.

Yes, I should have added all_access(-A). To be honest, current my implementation is halfway. It prevents non-root user from triggering refresh_acct but not from fetching values. It means any user can access acct metrics if root user accesses acct metrics at the same time.
On second thought, it may be best to adopt no-authentication, as you commented first, in order to allow pmlogger service to collect acct metrics.

>If you want to chat directly to me about any of this (might be easier to coordinate our efforts that way?) please feel free to contact me on slack - access details on https://pcp.io/community.html

Thank you for the suggestion. I consider using slack. In most situations e-mail or gihub may be preferable for me in terms of doing my day job (and in terms of my english skill). But slack will be helpful to discuss about complicated items. Thanks a lot.

natoscott:
@shogo-matsumoto sounds good - I'll focus on the QA and will merge that access control change.

natoscott:
@shogo-matsumoto I made a series of updates and additions to this new code yesterday - if you could find some time to review them, that'd be fantastic.

I have a couple of questions still, things I don't understand fully:
- why is the etime (elapsed time) metric a float and not an unsigned integer like the other time metrics (utime and stime)?  or should all three by floating point values?
- we're multiplying the time metrics by 1000 (converting seconds to milliseconds) - could we not export them directly in units of seconds?  (is this related to the floating point issue above?)

Thanks!

shogo-matsumoto:
@natoscott 

> I made a series of updates and additions to this new code yesterday - if you could find some time to review them, that'd be fantastic.

Thank you for updating the code. I’ll read it.

>why is the etime (elapsed time) metric a float and not an unsigned integer like the other time metrics (utime and stime)? or should all three by floating point values?

I used float for etime because ac_etime field of struct acct_v3 is declared as float. However, as you commented, I think all three values (etime, utime, stime) should be floating values. I found my mistake. comp_t type is encoded by encode_comp_t function in linux kernel. It should be decoded when we read acct_v3. I paste a draft of a fix in the bottom of this comment. Could you review it?

By the way, is PM_SEM_COUNTER in the following code appropriate? pmval does not show non-zero value for utime or stime if PM_SEM_COUNTER is used.

src/pmdas/linux_proc/pmda.c
```
    /* acct.psinfo.utime */
    { NULL, { PMDA_PMID(CLUSTER_ACCT,ACCT_UTIME),
      PM_TYPE_U32, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,1,0,0,PM_TIME_MSEC,0) }, },
    /* acct.psinfo.stime */
    { NULL, { PMDA_PMID(CLUSTER_ACCT,ACCT_STIME),
      PM_TYPE_U32, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,1,0,0,PM_TIME_MSEC,0) }, },
```

>we're multiplying the time metrics by 1000 (converting seconds to milliseconds) - could we not export them directly in units of seconds? (is this related to the floating point issue above?)

Yes, I should have used seconds (and float).

Thanks a lot,


draft of a fix:
```
diff --git a/src/pmdas/linux_proc/acct.c b/src/pmdas/linux_proc/acct.c
index 92ec79821..1841bec0e 100644
--- a/src/pmdas/linux_proc/acct.c
+++ b/src/pmdas/linux_proc/acct.c
@@ -82,6 +82,21 @@ get_end_time_v3(void *entry)
 	   (int)(((struct acct_v3 *)entry)->ac_etime / hz);
 }
 
+static unsigned long long 
+decode_comp_t(comp_t c)
+{
+    int exp;
+    unsigned long long val;
+
+    exp = (c >> 13) & 0x7;
+    val = c & 0x1fff;
+
+    while (exp-- > 0)
+	val <<= 3;
+
+    return val;
+}
+
 static int
 acct_fetchCallBack_v3(int item, void *p, pmAtomValue *atom)
 {
@@ -103,31 +118,31 @@ acct_fetchCallBack_v3(int item, void *p, pmAtomValue *atom)
 	atom->ul = acctp->ac_btime;
 	break;
     case ACCT_ETIME:
-	atom->f = acctp->ac_etime * 1000 / hz;
+	atom->f = acctp->ac_etime / hz;
 	break;
     case ACCT_UTIME:
-	atom->ul = acctp->ac_utime * 1000 / hz;
+	atom->f = decode_comp_t(acctp->ac_utime) * 1.0 / hz;
 	break;
     case ACCT_STIME:
-	atom->ul = acctp->ac_stime * 1000 / hz;
+	atom->f = decode_comp_t(acctp->ac_stime) * 1.0 / hz;
 	break;
     case ACCT_MEM:
-	atom->ul = acctp->ac_mem;
+	atom->ull = decode_comp_t(acctp->ac_mem);
 	break;
     case ACCT_IO:
-	atom->ul = acctp->ac_io;
+	atom->ull = decode_comp_t(acctp->ac_io);
 	break;
     case ACCT_RW:
-	atom->ul = acctp->ac_rw;
+	atom->ull = decode_comp_t(acctp->ac_rw);
 	break;
     case ACCT_MINFLT:
-	atom->ul = acctp->ac_minflt;
+	atom->ull = decode_comp_t(acctp->ac_minflt);
 	break;
     case ACCT_MAJFLT:
-	atom->ul = acctp->ac_majflt;
+	atom->ull = decode_comp_t(acctp->ac_majflt);
 	break;
     case ACCT_SWAPS:
-	atom->ul = acctp->ac_swaps;
+	atom->ull = decode_comp_t(acctp->ac_swaps);
 	break;
     case ACCT_EXITCODE:
 	atom->ul = acctp->ac_exitcode;
diff --git a/src/pmdas/linux_proc/pmda.c b/src/pmdas/linux_proc/pmda.c
index 82357243c..6b6f0185d 100644
--- a/src/pmdas/linux_proc/pmda.c
+++ b/src/pmdas/linux_proc/pmda.c
@@ -1427,31 +1427,31 @@ static pmdaMetric metrictab[] = {
       PM_TYPE_U32, ACCT_INDOM, PM_SEM_DISCRETE, PMDA_PMUNITS(0,1,0,0,PM_TIME_SEC,0) }, },
     /* acct.psinfo.etime */
     { NULL, { PMDA_PMID(CLUSTER_ACCT,ACCT_ETIME),
-      PM_TYPE_FLOAT, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,1,0,0,PM_TIME_MSEC,0) }, },
+      PM_TYPE_FLOAT, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,1,0,0,PM_TIME_SEC,0) }, },
     /* acct.psinfo.utime */
     { NULL, { PMDA_PMID(CLUSTER_ACCT,ACCT_UTIME),
-      PM_TYPE_U32, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,1,0,0,PM_TIME_MSEC,0) }, },
+      PM_TYPE_FLOAT, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,1,0,0,PM_TIME_SEC,0) }, },
     /* acct.psinfo.stime */
     { NULL, { PMDA_PMID(CLUSTER_ACCT,ACCT_STIME),
-      PM_TYPE_U32, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,1,0,0,PM_TIME_MSEC,0) }, },
+      PM_TYPE_FLOAT, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,1,0,0,PM_TIME_SEC,0) }, },
     /* acct.psinfo.mem */
     { NULL, { PMDA_PMID(CLUSTER_ACCT,ACCT_MEM),
-      PM_TYPE_U32, ACCT_INDOM, PM_SEM_INSTANT, PMDA_PMUNITS(1,0,0,PM_SPACE_KBYTE,0,0) }, },
+      PM_TYPE_U64, ACCT_INDOM, PM_SEM_INSTANT, PMDA_PMUNITS(1,0,0,PM_SPACE_KBYTE,0,0) }, },
     /* acct.psinfo.io */
     { NULL, { PMDA_PMID(CLUSTER_ACCT,ACCT_IO),
-      PM_TYPE_U32, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,0,1,0,0,PM_COUNT_ONE) } },
+      PM_TYPE_U64, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,0,1,0,0,PM_COUNT_ONE) } },
     /* acct.psinfo.rw */
     { NULL, { PMDA_PMID(CLUSTER_ACCT,ACCT_RW),
-      PM_TYPE_U32, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,0,1,0,0,PM_COUNT_ONE) } },
+      PM_TYPE_U64, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,0,1,0,0,PM_COUNT_ONE) } },
     /* acct.psinfo.minflt */
     { NULL, { PMDA_PMID(CLUSTER_ACCT,ACCT_MINFLT),
-      PM_TYPE_U32, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,0,0,0,0,0) }, },
+      PM_TYPE_U64, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,0,0,0,0,0) }, },
     /* acct.psinfo.majflt */
     { NULL, { PMDA_PMID(CLUSTER_ACCT,ACCT_MAJFLT),
-      PM_TYPE_U32, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,0,1,0,0,PM_COUNT_ONE) } },
+      PM_TYPE_U64, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,0,1,0,0,PM_COUNT_ONE) } },
     /* acct.psinfo.swaps */
     { NULL, { PMDA_PMID(CLUSTER_ACCT,ACCT_SWAPS),
-      PM_TYPE_U32, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,0,1,0,0,PM_COUNT_ONE) } },
+      PM_TYPE_U64, ACCT_INDOM, PM_SEM_COUNTER, PMDA_PMUNITS(0,0,1,0,0,PM_COUNT_ONE) } },
 
     /* acct.id.uid */
     { NULL, { PMDA_PMID(CLUSTER_ACCT,ACCT_UID),
```

natoscott:
> [...]
> I used float for etime because ac_etime field of struct acct_v3 is declared as float. However, as you commented, I think all three values (etime, utime, stime) should be floating values. I found my mistake. comp_t type is encoded by encode_comp_t function in linux kernel. It should be decoded when we read acct_v3. I paste a draft of a fix in the bottom of this comment. Could you review it?

Looks good!  I also had been looking at the v2 structure in acct(5) man page, I missed that I needed to scroll down to see the float in v3.

> By the way, is PM_SEM_COUNTER in the following code appropriate? pmval does not show non-zero value for utime or stime if PM_SEM_COUNTER is used.

I think so - this follows the equivalent metrics in the proc.* metrics (e.g. proc.psinfo.utime).  Its a cumulative count of time spent by the process executing user code.  pmval does rate convesion on counters, so I'd expect it would show non-zero only during samples where the value changes.  One can use the -r/--raw option to pmval to get see the non-rate converted value.

> [...]
> draft of a fix:

Looks good @shogo-matsumoto - I'll merge that shortly.

One additional issue I've just realised exists is that we are now enabling process accounting for every PCP installation when it may not have been enabled before.  This is not good, as it incurs an overhead for everyone - and plenty of people wont use these particular metrics, but still incur these overheads.

We'll need to take one of several approaches there I think - which do you think best?

- only enable (acct_init) process accounting once a client tool asks for one of the metrics.  we can disable accounting when all clients accessing the metrics if we track all of the connected clients accessing acct.*

- move the code into a new pmdaacct(1) which is not default installed.

- require an explicit pmStore(3) / pmstore(1) to enable and disable accounting - adding acct.control.* metrics along the lines of the proc.control.* metrics.

shogo-matsumoto:
@natoscott 

>pmval does rate convesion on counters, so I'd expect it would show non-zero only during samples where the value changes.

Ok, I understand.

>We'll need to take one of several approaches there I think - which do you think best?

I think it may be best to move the code into a new pmdaacct, but it will take some time if I do that.  I made a PR (https://github.com/performancecopilot/pcp/pull/984) as a (temporarily) workaround which has a flag acct.control.enable_acct to control enable/disable of acct. As default it will prevent pmdaproc from creating private acct file (/var/lib/pcp/tmp/pmcd/pacct) and from calling acct. Note that access to system acct file (/var/account/pacct) is not restricted by the flag because other process is responsible for managing the file. Is the fix acceptable?

natoscott:
@shogo-matsumoto that's fine - I think the store-based approach is probably the best actually.

I realised after my previous post that there's some code shared with other parts of pmdaproc that wouldn't be easy to extract (like the gid/uid name caches).

I've updated the test to exercise the new code and made a small fix to the fetch return code status.  I noticed there's no help text for the new metrics you added - could you add a little writeup for each? - e.g.

```
$ pminfo -t acct.control
acct.control.open_retry_interval [Interval of failed acct file opening check]
acct.control.check_acct_interval [Interval on which acct file size is checked]
acct.control.file_size_threshold [Size at which the acct file will be reopened]
acct.control.lifetime One-line Help: Error: One-line or help text is not available
acct.control.refresh One-line Help: Error: One-line or help text is not available
acct.control.enable_acct One-line Help: Error: One-line or help text is not available
```

shogo-matsumoto:
@natoscott 

>I noticed there's no help text for the new metrics you added - could you add a little writeup for each? - e.g.

Sorry that I forgot to add help text. I made a PR(https://github.com/performancecopilot/pcp/pull/995). Thank you for helping me with testing and so on.

natoscott:
> @natoscott
> 
> > I noticed there's no help text for the new metrics you added - could you add a little writeup for each? - e.g.
> 
> Sorry that I forgot to add help text. I made a PR(#995). Thank you for helping me with testing and so on.

No problem at all.  I think we're finished from the PMDA side of things now!

PR #981, opened 2020-07-21
pmdaproc: add cgroup.memory.current metric (cgroup v2)

andreasgerstmayr:
I wasn't sure about the numbering - in the `cgroups.h` there are cgroup v2 metrics with a lower number than some cgroup v1 metrics - how did you know 5 years back that 8 months back you will add some other cgroup v2 metrics there? :open_mouth: [1]

I gave it the number 90 now and a new block, but can change that if there's a system which I should follow.

[1] https://github.com/performancecopilot/pcp/commit/e3c7ea17b68854f7c098003bb4537e13aa65dad7#diff-23194b027d3722a739170dfbb2f9d940R117

### DIFF
--- a/qa/730.out
+++ b/qa/730.out
@@ -109,6 +109,7 @@ cgroup.io.stat.rbytes
 cgroup.io.stat.rios
 cgroup.io.stat.wbytes
 cgroup.io.stat.wios
+cgroup.memory.current
 cgroup.memory.failcnt
 cgroup.memory.id.container
 cgroup.memory.limit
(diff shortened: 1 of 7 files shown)


natoscott:
> I wasn't sure about the numbering - in the `cgroups.h` there are cgroup v2 metrics with a lower number than some cgroup v1 metrics - how did you know 5 years back that 8 months back you will add some other cgroup v2 metrics there?

Heh, mate I can barely remember what I was doing yesterday, there's no chance I can remember 5 years back.  :)

> 
> I gave it the number 90 now and a new block, but can change that if there's a system which I should follow.
> 

LGTM.

PR #998, opened 2020-08-04
pmproxy, pmsearch, libpcp_web: search functions renamed to be more alike to its series counterparts; man: updated docs of search related functions

Erbenos:


natoscott:
@Erbenos as mentioned on slack, I think we should keep the 'Text' part of the API names in anticipation of a future (string metric) Values search index.

Erbenos:
Reverted back to including "Text" segment in related function names.

PR #1054, opened 2020-09-16
build: rename python-dev to python2-dev

andreasgerstmayr:
Debian bullseye removed the python-dev package. python2-dev also exists
on Debian buster, therefore use this instead of python-dev.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,6 @@
 name: CI
 on:
   push:
-    branches:
-      - master
   pull_request:
     branches:
       - master
--- a/Makepkgs
+++ b/Makepkgs
@@ -225,23 +225,6 @@ then
     # $(configure_tools)
     #
     configopts=''
-    if [ -f /etc/lsb-release ]
-    then
-	if grep -q 'DISTRIB_ID=Ubuntu' /etc/lsb-release
-	then
-	    eval `grep DISTRIB_RELEASE= /etc/lsb-release`
-	    XDISTRIB_RELEASE=`echo $DISTRIB_RELEASE | sed -e 's/[^0-9]//g' -e 's/\.//'`
-	    [ -z "$XDISTRIB_RELEASE" ] && XDISTRIB_RELEASE=0000
-	    if [ $XDISTRIB_RELEASE -ge 2004 ]
-	    then
-		# As of Ubuntu 20.04 the Python2 dependency packages have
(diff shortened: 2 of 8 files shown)


andreasgerstmayr:
This should fix the build failure on debian bullseye.

kmcdonell:
@andreasgerstmayr we're fixing the same issue in two different ways.  For Debian, you're accommodating the package name change, but for Ubuntu I took a more autocratic decision and decided there is no longer any reason to include Python2 support into the PCP packages so I added ```--without python``` to configure in Makepkgs.
Either approach is OK, but we probably don't need both.  So some rationalisation may be needed to reduce confusion furhter down the track.

andreasgerstmayr:
> decided there is no longer any reason to include Python2 support

+1
given the fact that Python 2 is EOL since beginning of this year, I don't mind dropping it entirely

I'll update this PR tomorrow.

andreasgerstmayr:
Sorry, I just couldn't resist accommodating the package name change - it's just a tiny change, in case some people really want to use Python 2. I changed the logic to use /etc/os-release, because /etc/lsb-release only exists on Ubuntu, where /etc/os-release exists on both (also on Fedora).

I've added the same logic to the configure script, to disable Python 2 unless explicitly requested.

natoscott (commented):
Does this cleanly handle the situation where someone has python-pcp installed and is upgrading to pcp-5.2.1?  Also is the case where a 3rd party package has a dependency on pcp-python handled?  (dpkg "Provides:" keyword IIRC, in debian/control*)

andreasgerstmayr:
> Does this cleanly handle the situation where someone has python-pcp installed and is upgrading to pcp-5.2.1?

Right now it just doesn't build the python-pcp package anymore.

I started a debian bullseye container, and:

* python-pcp isn't available in the debian repos
* it is available on bintray, however:
* `apt-cache rdepends python-pcp` shows that no package depends on python-pcp

> Also is the case where a 3rd party package has a dependency on pcp-python handled? (dpkg "Provides:" keyword IIRC, in debian/control*)

Like the obsoletes keyword? i.e. specifying that python3-pcp provides python-pcp? imho that's not the case here, because Python 3 is not backwards-compatible to Python 2, so python3-pcp does not provide python-pcp (they also have their own site-packages directory, etc.).

Yes, afaics the only issue can be users packaging their Python 2 only PMDA as .deb. Should we just take a step back and continue building the Python 2 bindings (and accommodate the python2-dev name change)?

PR #1070, opened 2020-09-28
fix for EINPROGRESS in __pmConnect()

kmcdonell:
Changes committed to git@github.com:kmcdonell/pcp.git 20200928

Ken McDonell (1):
      src/libpcp/src/auxconnect.c: handle EINPROGRESS from connect() in __pmConnect

 src/libpcp/src/auxconnect.c |   53 +++++++++++++++++++++++++++++++++++++-------
 1 file changed, 45 insertions(+), 8 deletions(-)

Details ...

commit adf7ea4bb4f2853ecb07f993031ec36be52fbbfb
Author: Ken McDonell <kenj@kenj.id.au>
Date:   Mon Sep 28 14:26:32 2020 +1000

    src/libpcp/src/auxconnect.c: handle EINPROGRESS from connect() in __pmConnect
    
    Give us a chance to succeed, even when connect() returns EINPROGRESS.

### DIFF
--- a/src/libpcp/src/auxconnect.c
+++ b/src/libpcp/src/auxconnect.c
@@ -660,18 +660,55 @@ int
 __pmConnect(int fd, void *addr, __pmSockLen addrlen)
 {
     __pmSockAddr *sock = (__pmSockAddr *)addr;
+    int		sts;
     if (sock->sockaddr.raw.sa_family == AF_INET)
-        return connect(fd, &sock->sockaddr.raw, sizeof(sock->sockaddr.inet));
-    if (sock->sockaddr.raw.sa_family == AF_INET6)
-        return connect(fd, &sock->sockaddr.raw, sizeof(sock->sockaddr.ipv6));
+        sts = connect(fd, &sock->sockaddr.raw, sizeof(sock->sockaddr.inet));
+    else if (sock->sockaddr.raw.sa_family == AF_INET6)
+        sts = connect(fd, &sock->sockaddr.raw, sizeof(sock->sockaddr.ipv6));
 #if defined(HAVE_STRUCT_SOCKADDR_UN)
-    if (sock->sockaddr.raw.sa_family == AF_UNIX)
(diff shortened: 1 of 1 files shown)


kmcdonell:
@goodwinos sorry, this got merged as a by-product of merging qa stuff for tonight's CI QA run.
Please review this one still, if there are any issues I'll revert/rework.
Thanks.

goodwinos:
> @goodwinos sorry, this got merged as a by-product of merging qa stuff for tonight's CI QA run.
> Please review this one still, if there are any issues I'll revert/rework.
> Thanks.

will do - sorry have been distracted with some really weird instances labels issues

goodwinos:
Looks good Ken - the connect(2) man page seems to indicate AF_UNIX might return EAGAIN in the same way the other address families can return EINPROGRESS. It's not however very clear whether AF_UNIX/EAGAIN would need to retry the connect(), but if not, then we might need e.g.

```
--- a/src/libpcp/src/auxconnect.c
+++ b/src/libpcp/src/auxconnect.c
@@ -681,7 +681,7 @@ __pmConnect(int fd, void *addr, __pmSockLen addrlen)
 
     /* something bad happened ... */
     sts = neterror();
-    if (sts == EINPROGRESS) {
+    if (sts == EINPROGRESS || (sts == EAGAIN && sock->sockaddr.raw.sa_family == AF_UNIX)) {

```

kmcdonell:
Thanks Mark.
I think it is OK, as we use the same simple guard (sts == EINPROGRESS) in other places in the code.  If we fall into the test block, then I'd expect the select() to timeout if we don't have the guard quite correct, and we go down the same old error path.
OK?

goodwinos:
OK looks fine then. I did find it strange only writefds should be checked in select(), but that's what the docs say to do!

kmcdonell:
On 29/9/20 10:49 am, Mark Goodwin wrote:
> OK looks fine then. I did find it strange only writefds should be 
> checked in select(), but that's what the docs say to do!

select on readfds would block forever/timeout if this is an output 
socket ... 8^)

PR #1107, opened 2020-10-13
migrate CI from check-vm to list-packages

kmcdonell:
See qa/admin/package-lists/README for an explanation, but the one liner is that we're leveraging all the check-vm infrastructure to seed one file with a list of packages for each disto+version+arch combination.  Any tweaking then involves editing these files and not trying to amend the check-vm control files any more.


natoscott (commented):
LGTM

andreasgerstmayr:
looks good, thanks!

PR #1118, opened 2020-10-21
Fixing runtime and compile issues with the activemq pdma

TroyWilsonAus:
When running the pdma we found a few issues.
* Invalid Header "http://localhost"
* Unable to load new from Queue ...

This PR resolves these issues.

### DIFF
--- a/src/pmdas/activemq/ActiveMQ.pm
+++ b/src/pmdas/activemq/ActiveMQ.pm
@@ -46,7 +46,7 @@ sub queues {
 
     my @queues = @{$query_result};
     my @queue_instances = map {
-	Queue->new($_->{'objectName'}, $self->{_rest_client});
+	PCP::Queue->new($_->{'objectName'}, $self->{_rest_client});
     } @queues;
     return @queue_instances;
 }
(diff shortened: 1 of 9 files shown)


src/pmdas/activemq/ActiveMQ.pm
```diff
@@ -46,7 +46,7 @@ sub queues {
 
     my @queues = @{$query_result};
     my @queue_instances = map {
-	Queue->new($_->{'objectName'}, $self->{_rest_client});
+	PCP::Queue->new($_->{'objectName'}, $self->{_rest_client});
```
natoscott, line 49: Thanks!  Is a similar change needed in ActiveMQTest.pl I wonder?  It seems to also use Queue->new()

TroyWilsonAus: Hi @natoscott ,
I've pushed more changes that should fix all the namespacing issues for the tests.

natoscott: Perfect, thanks Troy.

PR #1138, opened 2020-11-06
Linux PMDA: fix for waitio "counters" from /proc/stat going backwards

kmcdonell:


goodwinos:
looks good to me and the QA test passes

Issue #1140, opened 2020-11-11
PCP_TMPFILE_DIR used at build time as well as run time


ddiss:
This appears to be a regression from 029add427a42344ea7129452c9e0fb286d3cf881 .

When configured with default `prefix`, `localstatedir` and `tmpdir` paths, the build phase attempts to create temporary files under a non existent `/usr/local/var/tmp/` path, resulting in a failed build.
I think usage of PCP_TMPFILE_DIR should be limited to run-time only, or alternatively `prefix` should be removed from the `--with-tmpdir` default.

kmcdonell:
Trying a build --with-tmpdir=/deadbeef and repeating until it works.
So far I've found the following places:
src/libpcp/src/mk.exports
src/libpcp/src/mk.pmdbg
src/pmieconf/xtractnames
src/pmns/Make.stdpmid

And in each, replaced the tmp=... line with a stanza like this:
if [ -n "$PCP_TMPFILE_DIR" -a -d "$PCP_TMPFILE_DIR" ]
then
    tmp=`mktemp -d "$PCP_TMPFILE_DIR/pcp-foo.XXXXXXXXX"` || exit 1
else
    # assume run during a build and /tmp is a safe bet
    #
    tmp=`mktemp -d "/tmp/pcp-foo.XXXXXXXXX"` || exit 1
fi

I plan to commit this and await feedback from @ddiss

ddiss:
@kmcdonell thanks for looking at this!
https://github.com/kmcdonell/pcp/commit/adb769a8a9a56659774bc127a0bb7bc3527e4d25 looks good to me, although it could be changed to also consider `$TMPDIR` for the build-time case. E.g.
```
if [ -n "$PCP_TMPFILE_DIR" -a -d "$PCP_TMPFILE_DIR" ]
then
        tmp=`mktemp --tmpdir="$PCP_TMPFILE_DIR" -d pmns_stdpmid.XXXXXXXXX` || exit 1
else
        # assume run during a build, use $TMPDIR or /tmp.
        #
        tmp=`mktemp --tmpdir -d pmms_stdpmid.XXXXXXXXX` || exit 1
fi
```

kmcdonell:
Thanks @ddiss, but unfortunately mktemp --tmpdir does not work at all in non-GNU land, like the *BSD bretheren,

ddiss:
[...]

PR #1141, opened 2020-11-11
pmns/Make: drop duplicate if/else

ddiss:
As of 029add427a42344ea7129452c9e0fb286d3cf881, both the if and else
branches here match. Restoration of the previous build-time logic is
tracked via https://github.com/performancecopilot/pcp/issues/1140 and
will affect more scripts than just this one.

Signed-off-by: David Disseldorp <ddiss@suse.de>

### DIFF
--- a/src/pmns/Make.stdpmid
+++ b/src/pmns/Make.stdpmid
@@ -16,15 +16,7 @@
 # source the PCP configuration environment variables
 . $PCP_DIR/etc/pcp.env
 
-if [ -d "$PCP_TMPFILE_DIR" ]
-then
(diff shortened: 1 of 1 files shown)


goodwinos:
Hi David, nice to hear from you again! This change is trivially correct, though reading Ken's comment in the else part of the conditional, he probably intended to use /tmp, instead of $PCP_TMPFILE_DIR, i.e. `tmp=`mktemp -d "/tmp/pmns_stdpmid.XXXXXXXXX"` || exit 1`

But as you mentioned in #1140, we need to audit all use of $PCP_TMPFILE_DIR for build/runtime usage, as well as fix the configure prefix defaults .. but let's discuss that in #1140 and just merge this change.  thanks

PR #1144, opened 2020-11-16
Added docker-build image (for OL7)

TroyWilsonAus:
Added docker file and script to perform build in a container with most dependencies already installed.

This doesn't include all dependencies, (mostly some Perl CPAN objects are missing).

This worked for me for building core and c++ related pdma's


natoscott:
Our strategy has been to work towards continually improving qa/admin/check-vm rather than adding Docker images for specific build cases.  As discussed on slack, if a build fails due to check-vm missing toolchain / deps requirements, lets understand and fix that.

PR #1164, opened 2020-12-01
libpcp performance improvements for interp mode with dynamic indoms

kmcdonell:


goodwinos:
Built and installed fine. I've left it running QA for the new test 1286, and the atop, pmda.proc and libpcp groups. So far 632, 1079 and 1347 failed, which I'll take a look at in the morning (may or may not be related). It's a bit late in the evening to review the libpcp changes :P

goodwinos:
QA: 
632 - dates way out into the future
1079 - extra PID (670) in atop listing
1347 - similar to 1079, extra PIDs for pmsleep

All the above are passing on a different machine running current master branch

kmcdonell:
Thanks Mark.  I guess these failing as tests are not in -g archive nor -g multi-archive ... I'm in Echuca at the moment and won't be able to look at this until the middle of next week.

goodwinos:
Hi Ken, 632 is passing now after I remade qa/src/timeshift. The previous binary was built without your patch series, so I'm not sure if that's good or bad (!), but at least it's passing now.

1347 and 1079 might just need a remake for the extra PIDs showing up in atop listings. I'll retest with this PR on a different machine. 

Still need to review the libpcp changes ...

Enjoy Echuca :)

kmcdonell:
@goodwinos agree on 632 and not unexpected as it is exposed to an internal libpcp data structure I changed. 1347 is a boundary tail condition ... I need to double check which result is correct ... the missing process name and args from atop is a bit of a concern, see pids 2956902 and 2956910, but these ones disappear near the end of the archive.  From a quick look, 1079 seems similar with pid 670 coming and going near the end of the archive and no process name or args in the atop output.
Leave this in my court ... I'll add here once I have more analysis.

goodwinos:
We agreed to merge this PR - so I'll do that later this evening or in the morning. I'll just remake 1347 and 1079. And 632 should pass in CI because it's always a fresh build.

Thanks Ken

Cheers

PR #1166, opened 2020-12-03
pmdabcc: netproc module + atop per-process network statistics

andreasgerstmayr:
first version of the BCC netproc module, which provides data for the pcp atop per-process bandwidth statistics

I'm open to changes of the name (`netproc` was the best I could come up with) and metric names

```
bcc.proc.io.net.perpid.tcp.in.packets
bcc.proc.io.net.perpid.tcp.in.bytes
bcc.proc.io.net.perpid.tcp.out.packets
bcc.proc.io.net.perpid.tcp.out.bytes
bcc.proc.io.net.perpid.udp.in.packets
bcc.proc.io.net.perpid.udp.in.bytes
bcc.proc.io.net.perpid.udp.out.packets
bcc.proc.io.net.perpid.udp.out.bytes
```

(`bcc.proc.io.net.tcp.*` is already in use by the `tcplife` BCC module)

### DIFF
--- a/qa/1286
+++ b/qa/1286
@@ -13,12 +13,14 @@ echo "QA output created by $seq"
 _pmdabcc_check
 _pmdabcc_require_kernel_version 4 6
 which curl >/dev/null 2>&1 || _notrun "No curl binary installed"
+which dig >/dev/null 2>&1 || _notrun "No dig binary installed"
 
 status=1       # failure is the default!
 signal=$PCP_BINADM_DIR/pmsignal
 $sudo rm -rf $tmp.* $seq.full
 
-target_ip=1.1.1.1
+target_http=http://1.1.1.1
+target_dns=1.1.1.1
 
 _value_filter()
 {
@@ -42,10 +44,16 @@ EOF
 _pmdabcc_wait_for_metric
 
 # Generate system activity for the BCC netproc module
-curl -s http://${target_ip} > /dev/null &
-curlpid=$!
-echo "curl pid: ${curlpid}" >> "$seq.full"
-wait ${curlpid}
+# TCP
+curl -s ${target_http} > /dev/null &
+pid_curl=$!
+echo "curl pid: ${pid_curl}" >> "$seq.full"
+wait ${pid_curl}
+
+# UDP
+# dig performs the DNS request from the first forked process
+pid_dig=$(strace -f -e 'trace=!all' dig @$target_dns -x $target_dns 2>&1 | awk '/Process .* attached/ { print $3; exit }')
+echo "dig pid: ${pid_dig}" >> "$seq.full"
 
 for metric in bcc.proc.io.net.perpid.tcp.in.packets \
               bcc.proc.io.net.perpid.tcp.in.bytes \
@@ -55,7 +63,18 @@ do
     echo "=== report metric values for $metric ==="
     pminfo -f $metric 2>&1 | tee -a $here/$seq.full \
     | tee -a $seq.full \
-    | _value_filter ${curlpid}
+    | _value_filter ${pid_curl}
+done
+
+for metric in bcc.proc.io.net.perpid.udp.in.packets \
+              bcc.proc.io.net.perpid.udp.in.bytes \
+              bcc.proc.io.net.perpid.udp.out.packets \
+              bcc.proc.io.net.perpid.udp.out.bytes
+do
+    echo "=== report metric values for $metric ==="
+    pminfo -f $metric 2>&1 | tee -a $here/$seq.full \
+    | tee -a $seq.full \
+    | _value_filter ${pid_dig}
 done
 
 _pmdabcc_remove
--- a/qa/1286
+++ b/qa/1286
@@ -1,0 +1,64 @@
+#!/bin/sh
+# PCP QA Test No. 1286
+# Exercise the BCC PMDA netproc hits module - install, remove and values.
+#
+# Copyright (c) 2020 Red Hat.
+#
+
+seq=`basename $0`
+echo "QA output created by $seq"
+
+. ./common.bcc
+
+_pmdabcc_check
+_pmdabcc_require_kernel_version 4 6
+which curl >/dev/null 2>&1 || _notrun "No curl binary installed"
+
+status=1       # failure is the default!
+signal=$PCP_BINADM_DIR/pmsignal
+$sudo rm -rf $tmp.* $seq.full
+
+target_ip=1.1.1.1
+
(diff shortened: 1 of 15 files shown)


andreasgerstmayr:
I've left all references to netatop in place in an effort to not deviate too much from upstream atop.

Currently I'm removing the network statistics of finished processes (in the PMDA).
* Pros: the list of instances doesn't grow indefinitely
* Cons: we miss short-lived processes

netatop comes with a netatop daemon (netatopd), which stores network statistics from finished processes on the disk, i.e. it catches also short-lived processes and retains the history.

The "system and process activity since boot" is misleading at the moment, need to check this. Also need to verify a few things in the eBPF script (number of packets might be incorrect). Other than that, the "n" key already shows per-process network stats :)

qa/1286
```diff
@@ -0,0 +1,83 @@
...
+# Generate system activity for the BCC netproc module
+# TCP
+curl -s ${target_http} > /dev/null &
+pid_curl=$!
+echo "curl pid: ${pid_curl}" >> "$seq.full"
+wait ${pid_curl}
+
+# UDP
+# dig performs the DNS request from the first forked process
+pid_dig=$(strace -f -e 'trace=!all' dig @$target_dns -x $target_dns 2>&1 | awk '/Process .* attached/ { print $3; exit }')
```
natoscott, line 55: This is a bashism but this is a /bin/sh script.  Also, we need a notrnun check on strace here I guess.

src/pcp/atop/GNUmakefile
```diff
@@ -90,6 +91,9 @@ procmetrics.h:	pmgenmap.sh procmetrics.map
 acctmetrics.h:	pmgenmap.sh acctmetrics.map
 	$(PMGENMAP) acctmetrics.map > acctmetrics.h
 
+netprocmetrics.h:	pmgenmap.sh netprocmetrics.map
+	$(PMGENMAP) netprocmetrics.map > netprocmetrics.h
+
```
natoscott, line 96: Just a heads up - you may have some code here affected by the 'const' (metric names) changes from yesterday - affects output from pmgenmap too.

src/pcp/atop/netproc.c
```diff
@@ -0,0 +1,79 @@
...
+
+	for (i=0; i < netproc_insts_len; i++)
+	{
+		if (pmDebugOptions.appl0)
+			fprintf(stderr, "%s: updating net info of process %d: %s\n",
+				pmGetProgname(), pids[i], insts[i]);
+
+		// for each instance of the netproc metrics, we need to find the index inside the tasks array
+		// TODO: better algorithm to avoid O(m*n) complexity?
+		for(j=0; j < taskslen; j++) {
```
natoscott, line 73: Coding style (braces, whitespace).

src/pcp/atop/netproc.c
```diff
@@ -0,0 +1,79 @@
...
+{
+	static int 	setup;
+	static pmID	pmids[TASK_NET_NMETRICS];
+	static pmDesc	descs[TASK_NET_NMETRICS];
+	pmResult	*result;
+	char		**insts;
+	int		*pids;
+	int		netproc_insts_len, i, j;
+
+	if(!setup) {
```
natoscott, line 57: Coding style (braces, whitespace).

natoscott (commented):
Looking good!  Just minor nits...

qa/1286
andreasgerstmayr: hm, I'm using shellcheck which gives you all kinds of linting warnings on shellscripts [1]
afaics this syntax is already a POSIX standard and the old syntax is not recommended anymore ("Because of these inconsistent behaviors, the backquoted variety of command substitution is not recommended for new applications that nest command substitutions or attempt to embed complex scripts.") [2] [3]

Are there shells which can't parse this new syntax? I'm fine changing it to backticks to be in sync with the other shellscripts, but this feels like a (granted, a very tiny) step backwards - shouldn't we go the other direction and embrace the new syntax everywhere instead?

[1] https://www.shellcheck.net/
[2] https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xcu_chap02.html#tag_23_02_06_03
[3] https://github.com/koalaman/shellcheck/wiki/SC2006

src/pcp/atop/GNUmakefile
andreasgerstmayr: thx for the heads up!
I'll do a rebase before the final PR

src/pcp/atop/netproc.c
andreasgerstmayr: I've seen https://github.com/performancecopilot/ansible-pcp/blob/master/.travis/runblack.sh in the Ansible roles repo the other day. Black [1] is a code formatter with almost no settings, which takes any unformatted code and replaces it with nicely formatted code. No matter which editor you use and how non-uniform you place whitespaces etc., you will always end up with exactly the same formatting (and tries to produce the smallest diffs possible). afaics it takes the AST and formats it, i.e. the tool ignores all previous formatting. The same thing is also done (and enforced) in every Grafana plugin (using a tool called prettier [2]).

Is there interest to enforce (and automatically apply) a consistent styleguide to all our C code as well? A similar tool I could find for C code is clang-format [3].

[1] https://github.com/psf/black
[2] https://prettier.io/
[3] https://clang.llvm.org/docs/ClangFormat.html

qa/1286
natoscott: I tried a couple and csh doesn't seem to accept it - I dunno if anyone uses that for /bin/sh though.  It mainly stood out cos we just don't use that syntax anywhere else and backticks are pretty well understood.

src/pcp/atop/netproc.c
natoscott: | Is there interest to enforce (and automatically apply) a consistent styleguide to all our C code as well? 

Its come up from time to time, and generally the answer is "no, let sleeping dogs lie"  :)

qa/1286
andreasgerstmayr: okay, I've updated it.

src/pcp/atop/netproc.c
andreasgerstmayr: hmm, I'd be open for some modernization of the codebase ;) (also looking at the makefiles - there are much more sophisticated tools available in 2020) - might be worth a discussion on the mailing list after the next release(s)

src/pcp/atop/netproc.c
andreasgerstmayr: done

PR #1184, opened 2020-12-24
libpcp interp.c rework, minor fixes and qa

kmcdonell:
Something bad seems to have happened in the change to "rebase rather than merge" and I'm seeing a slew of commits here that are not expected (at least by me).
I'll leave the PR open until we can sort out what's going on.


kmcdonell:
I've done a recusive diff of my tree and the official tree, and the only differences are the ones I'd expect from my small number of commits.
And my tree passes QA on bozo, bozo-vm and vm03 ...
So I don't understand why github thinks there are so many commits in this PR, but I'm going to merge away.

PR #1189, opened 2021-01-14
Linux ZFS PMDA

mac133k:
PMDA for OpenZFS on Linux was added under src/pmdas/linux_zfs.


natoscott:
@mac133k this looks great!  Here's my 'todo' notes from reading through all the changes and additions - I've fixed a few and the rest are all small items that can arrive via follow up commits (I'll merge it shortly).

- Some metrics are missing help text
  ? might be base some descriptions on:
  https://openzfs.github.io/openzfs-docs/man/1/arcstat.1.html
- ZFS_POOL InDom missing help text [fixed]
- Linker failure with -flto [fixed]
  /usr/bin/ld: pmda.o (symbol from plugin): in function `zfs_init':
  (.text+0x0): multiple definition of `ZFS_PATH'; zfs_utils.o (symbol from plugin):(.text+0x0): first defined here
- Missing .gitignore content for generated files [fixed]
- DOS line endings on some header files [fixed]
- Some unused macros in the makefile [fixed]
- Missing license headers on new source files
- Missing copyright notice on new source files
- New man page is missing content
- Optimization - some strcmp use repeats comparison of common
prefix when extracting values (e.g. "dnode_", "scatter_", "cache_",
"hash_", "metadata_", "dmu_", etc) inside file scanning loops - the
common prefixes could be tested for once, if matching skip over and
then perform shorter string comparisons on the remaining suffixes.

PR #1197, opened 2021-01-18
Conditionals were branched out to optimize string matching on fetch.

mac133k:


natoscott:
@mac133k thanks - I've made some optimizations (and other changes) above and beyond yours here - please double-check 'em for me?  Thanks!

mac133k:
@natoscott I was thinking about tokenizing metric names, but your solution is simpler and better. I checked the metrics and it seems like all the values are fetched properly, so good stuff. Thanks.

mac133k:
@natoscott In your code `value = strtoull(mval, NULL, 0)` is called before the metric name is matched which means that invalid values are passed for conversion from the first 2 lines of a stat file, ie.:

```
$ head -5 /proc/spl/kstat/zfs/arcstats 
13 1 0x01 116 31552 2235805236238192 3208739869080194
name                            type data
hits                            4    791425
misses                          4    0
demand_data_hits                4    0
```
However it did not result in any serious error, because ultimately the value remains unassigned if the string from the first column does not match a known metric.

natoscott:
@mac133k oh, that's a good point!  Do you prefer is we revert to using individual strtoull calls for each field or leave as-is?

mac133k:
No, I agree that the previous version with strtoull in every line looked ugly, so let's leave it as is for now. I have an idea how to skip the header lines. I will submit a new PR soon.

PR #1198, opened 2021-01-19
Pmdasockets

goodwinos:


goodwinos:
I have some more changes for this PR, which I'll commit tomorrow  (wed morning Melb TZ).

build/rpm/redhat.spec
```diff
@@ -270,7 +270,7 @@ BuildRequires: qt5-qtsvg-devel
 %endif
 %endif
 
-Requires: bash xz gawk sed grep findutils which %{_hostname_executable}
+Requires: bash xz gawk sed grep findutils which iproute %{_hostname_executable}
```
natoscott, line 273: Since the implementation didn't end up using the kernel interfaces directly, I think this PMDA should be non-default and in its own pcp-pmda-sockets to isolate this new package dependency (away from core pcp).

How stable is the output of ss vs using the underlying kernel interfaces I wonder?

goodwinos: > Since the implementation didn't end up using the kernel interfaces directly, I think this PMDA should be non-default and in its own pcp-pmda-sockets to isolate this new package dependency (away from core pcp).

I thought this was clear - the run-time dependency of the PMDA on ss in the iproute package pales into insignificance compared to the maintenance nightmare of providing metrics directly from the kernel interfaces that are compatible with the ss output. The RFE requested the PCP equivalent of the ss metrics, which is what they're getting. And the maintenance effort should stay in one place,  with Alexey Kuznetsov (the ss and iproute author).

> 
> How stable is the output of ss vs using the underlying kernel interfaces I wonder?

Very stable AFAICT. The output of ss is parsed by many utilities, including the monitor.sh script used by the network support folks who requested the RFE. As such, it insulates those consumers from kernel version and netlink interface changes (and assorted exotic build dependencies). 

I'll put the PMDA in it's own package if you really want, but one install dep on the widely available and stable iproute package isn't a problem as I see it.

natoscott: > maintenance nightmare of providing metrics directly from the kernel interfaces

Doesn't seem to be a problem for any other PMDA - many tools use these interfaces directly, its a shame we've gone this much-more-expensive, high-level route IMO.  But that's OK - your choice.

> I'll put the PMDA in it's own package if you really want

Yes, please - it sets a bad precedent otherwise, with domain-specific dependencies on core PCP.  We strive for fewer dependencies on core PCP, not more.

Thanks!

goodwinos:
Abandoning this PR since it has a lot of packaging and CI related commits that would need to be reverted. I'll cherry-pick what's to be kept.

PR #1206, opened 2021-01-25
systemd: remove Wants=pmcd from pmlogger and pmie

### LINKED ISSUE #1200: Remove dependency on pmcd in pmlogger and pmie service
Both `pmlogger.service` and `pmie.service` contain `Wants=pmcd.service`, i.e. a weak dependency on `pmcd.service` resulting in pmcd getting started when pmlogger or pmie is started.

However, pmcd is not required for pmlogger or pmie in every case - it's perfectly valid to connect to a remote pmcd only, without having a local pmcd running.

For context, this requirement came from having lightweight daemons in the container (https://bugzilla.redhat.com/show_bug.cgi?id=1901130). Jan discovered that it's not possible to run pmlogger without pmcd, because pmlogger always activates the pmcd service. But, as explained above, I think it's not a container issue - pmlogger and pmie shouldn't start a local pmcd, because a local pmcd may not be required (it depends on the configuration of these daemons).

andreasgerstmayr:
Rationale: pmlogger and pmie should not always start pmcd as a dependency,
because a local pmcd is not required if pmlogger or pmie is connecting to a
remote pmcd.

Resolves #1200

### DIFF
--- a/src/pmie/pmie.service.in
+++ b/src/pmie/pmie.service.in
@@ -4,7 +4,6 @@ Documentation=man:pmie(1)
 After=network-online.target pmcd.service
 Before=pmie_check.timer pmie_daily.timer
 BindsTo=pmie_check.timer pmie_daily.timer
-Wants=pmcd.service
 
 [Service]
 Type=notify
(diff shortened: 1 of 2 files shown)


natoscott (approved):
@andreasgerstmayr LGTM.

goodwinos:
agree, looks good. If pmlogger config is needed for a local pmlogger, it's an administrative issue to make sure pmlogconf can generate it (which may mean enabling local pmcd).

PR #1207, opened 2021-02-01
build: add versioned rpm spec dependencies for pcp-zeroconf to pcp.spec.in

andreasgerstmayr:
same changes and reason as in 9dc78f07ddbcff75697662ea03671a5a094e34d5


natoscott (approved):
@andreasgerstmayr LGTM.

PR #1219, opened 2021-02-21
fix no longer working link

christianhorn:
https://prometheus.io/docs/concepts/metric_types/ seems to make most sense.
https://github.com/OpenObservability/OpenMetrics/blob/master/specification/OpenMetrics.md looks like a more upstream definition, but it refers to types which openmetrics seems to not support currently.


natoscott:
The second link is the "formal" specification of the OpenMetrics format (nowadays supported by many more tools than Prometheus alone) ... it is the link we want here.

It is OK that we don't support every type (yet - we will support more over time, as this community standard evolves) - you'll see in the spec the use of words like "may support" vs "must support" - AFAIK, we do support all the "must" clauses, and over time we'll support more and more of the "may" ones.

PR #1224, opened 2021-03-03
zeroconf: remove duplicate metrics from 30-second sampling setting.

shogo-matsumoto:
There are two intervals (10 sec, 30 sec) to collect kernel.all.cpu.*
metrics if zeroconf is enabled.

In this case, the metrics can be collected twice in a very short period
of time. Due to the short sampling interval, large error values may be
output in rate calculation.

To resolve the issue, remove the metrics from 30-second sampling setting.
At the same time, we'll add kernel.all.cpu.vuser to tools/sar to make
sure we don't lose the metric.

modified: src/pmlogconf/zeroconf/pidstat
modified: src/pmlogconf/tools/sar

### DIFF
--- a/src/pmlogconf/tools/sar
+++ b/src/pmlogconf/tools/sar
@@ -71,6 +71,7 @@ probe	kernel.uname.sysname ~ Linux ? include : exclude
 	kernel.all.cpu.sys
 	kernel.all.cpu.user
 	kernel.all.cpu.wait.total
+	kernel.all.cpu.vuser
 	kernel.all.intr
 	kernel.all.load
 	kernel.all.pswitch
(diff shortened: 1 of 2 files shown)


goodwinos:
Thanks Shogo-san, your changes to avoid duplicate logging are good in general - however we're currently reviewing the zeroconf settings, in particular for per-process and per-thread metrics. I also note there are some kernel.all.cpu.* metrics not included (e.g. guest_nice and vnice). So we'll be discussing this a bit more and merging your changes next week, likely with a few additional changes. It would be good if pmlogger could detect and handle duplicate logging blocks for individual metrics (e.g. by only logging the shortest interval instead of all requested intervals), but that's a complicated change and probably not the right  solution anyway.

shogo-matsumoto:
>So we'll be discussing this a bit more and merging your changes next week, likely with a few additional changes. It would be good if pmlogger could detect and handle duplicate logging blocks for individual metrics (e.g. by only logging the shortest interval instead of all requested intervals), but that's a complicated change and probably not the right solution anyway.

Hi Mark-san, thank you for your review and the comment. My patch is an ad hoc workaround and may not be a perfect solution, so I appreciate your help.

PR #1240, opened 2021-04-07
pmdabcc: netproc: add command name to instance names

### LINKED ISSUE #1233: Consider tweaking BCC PMDA netproc instance names
BCC PMDA netproc module provides metric instances like:

```
bcc.proc.io.net.perpid.tcp.recv.calls
    inst [4328 or "4328"] value 11873
```

But elsewhere the name is fixed-length:

```
bcc.proc.io.total
    inst [0 or "vda::000441"] value 53248

proc.psinfo.oom_score
    inst [4328 or "004328 wget"] value 668
```

Otherwise this would be cosmetic but with tools like _pmrep(1)_ one can select instances with e.g. `-i '.*001234.*'` so that doesn't now work across all the above metrics.

This doesn't look like a major issue but perhaps something to consider whether this sort of consistency would be helpful.

Thanks.

andreasgerstmayr:
Mimics proc PMDA's logic to cut the cmdline.

Resolves #1233

### DIFF
--- a/src/pmdas/bcc/modules/netproc.python
+++ b/src/pmdas/bcc/modules/netproc.python
@@ -13,6 +13,7 @@
 #
 """ PCP BCC PMDA netproc module """
 
+import re
 import ctypes as ct
 from collections import namedtuple
 import os
@@ -44,6 +45,7 @@ units_count = pmUnits(0, 0, 1, 0, 0, PM_COUNT_ONE)
 units_none = pmUnits(0, 0, 0, 0, 0, 0)
 
 # use tuple here instead of a dict or class to access the fields by index in the bpfdata method
+# do not change the first 8 items
 Netstats = namedtuple(
     "Netstats",
     [
@@ -55,6 +57,7 @@ Netstats = namedtuple(
         "udp_send_bytes",
         "udp_recv_calls",
         "udp_recv_bytes",
+        "instance_name",
     ],
 )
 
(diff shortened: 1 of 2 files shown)


myllynen (approved):
LGTM, tested to work as expected.

Just two minor questions. I think get_cmdline(pid) could be called directly so would be perhaps a bit cleaner, or what do you think? Would it make sense to make get_cmdline available for all modules (although the method doesn't exactly return cmdline at the moment so could think renaming if doing so)?

andreasgerstmayr:
Thanks for the suggestions! I've moved it to the base class now and renamed it accordingly (`get_instance_name_for_pid`).

andreasgerstmayr:
> I think get_cmdline(pid) could be called directly so would be perhaps a bit cleaner, or what do you think?

Using `self.get_cmdline()` instad of `PCPBCCModule.get_cmdline()`? tbh I got confused and thought I need to use the class name for static methods, but looks like in Python I don't need to (it's required in other languages for static methods). I've updated it accordingly.

Or are you referring to something else?

myllynen:
Yes, that's what I meant but I think even plain `get_cmdline(pid)` would be enough.

andreasgerstmayr:
> Yes, that's what I meant but I think even plain `get_cmdline(pid)` would be enough.

This got me curious:
```
class Test:
    @staticmethod
    def a():
        print("a() called")

    def b(self):
        Test.a()
        self.a()
        a()

t = Test()
t.b()

$ python test.py
a() called
a() called
  File "/home/agerstmayr/redhat/dev/test.py", line 12, in <module>
    t.b()
  File "/home/agerstmayr/redhat/dev/test.py", line 9, in b
    a()
NameError: name 'a' is not defined
```

I'll merge the PR now, thanks for the review!

PR #1242, opened 2021-04-09
rename netproc metrics and add pmlogconf rule

andreasgerstmayr:
afaics I've covered all usages of the netproc PMDA and renamed the metric in all places. Can you please double-check?

I've also added a pmlogconf rule to log the metric if it's available (note: the BCC PMDA is *not* installed by default).

### DIFF
--- a/qa/1286
+++ b/qa/1286
@@ -49,8 +49,8 @@ esac
 _pid_filter()
 {
     sed \
-      -e "s/$1/SERVERPID/g" \
-      -e "s/$2/CLIENTPID/g" \
+      -e "s/0*$1/SERVERPID/g" \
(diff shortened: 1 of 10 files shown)


myllynen (approved):
LGTM, thanks!

natoscott (approved):
@andreasgerstmayr LGTM.  It would be good to also add a pmlogrewrite file that renames the metrics in existing archives (maybe as a followup commit).  See commit a10c795487 for an example - should be a pretty simple config here using the pmlogrewrite NAME keyword.

andreasgerstmayr:
Thanks for the reviews! I've added a pmlogrewrite conf file now.

PR #1250, opened 2021-04-29
Integrate hiredis and hiredis-cluster as dependencies, add Redis MMV metrics

andreasgerstmayr:
* integrated hiredis and hiredis-cluster as dependencies
* added Redis MMV metrics
* qa/1603 to test Redis cluster mode

Note: there may be more info/warning messages in the pmproxy log than before, because due to a mismatch in the return code checking most of the Redis reply sanity checks used to be ignored (`checkStatusReplyXX` in `schema.c` etc.)


qa/1602
```diff
@@ -56,7 +54,7 @@ _save_config $PCP_SYSCONF_DIR/pmproxy
 $sudo rm -f $PCP_SYSCONF_DIR/pmproxy/*
 need_restore=true
 
-echo "== start two test Redis servers"
+echo "== start a test Redis server"
```
natoscott, line 57: @andreasgerstmayr I'm not following this - can we no longer specify multiple Redis servers this way?

natoscott: There's a couple of concerns with only supporting clusters now for multi-node setups:

- https://redis.io/topics/cluster-tutorial states: "Currently Redis Cluster does not support NATted environments and in general environments where IP addresses or TCP ports are remapped." ... not sure how big a deal that is in practice (that page does seem a little out of date maybe in places).

- we lose the ability to easily test and debug certain aspects (like key / command distribution, proxying - as in this test) - setting up an entire cluster is non-trivial relative to setting up just another Redis server.  This change to this test shows the problem there - its difficult to setup a cluster so the multi-server aspect was removed entirely.  :-(  This is an unfortunate backward step.

I guess it may be just the price we have to pay here, but ... it's not ideal.

andreasgerstmayr: > There's a couple of concerns with only supporting clusters now for multi-node setups:
> 
> * https://redis.io/topics/cluster-tutorial states: "Currently Redis Cluster does not support NATted environments and in general environments where IP addresses or TCP ports are remapped." ... not sure how big a deal that is in practice (that page does seem a little out of date maybe in places).

Yep, Redis has now `cluster-announce-ip` and `cluster-announce-port` settings (see https://github.com/redis/redis/commit/5ac5e3ebd7da8f041ebd7c4d20020208a3066b2d), so I'd say this is resolved.

> * we lose the ability to easily test and debug certain aspects (like key / command distribution, proxying - as in this test) - setting up an entire cluster is non-trivial relative to setting up just another Redis server.  This change to this test shows the problem there - its difficult to setup a cluster so the multi-server aspect was removed entirely.  :-(  This is an unfortunate backward step.

Please take a look at qa/1603 - it creates a cluster, fills it with data from a PCP archive and also checks key distribution on each node (using `dbsize`). I removed the multi-node aspect from qa/1602 because multi-node cluster without cluster mode is not supported anymore, and we have qa/1603. But I can create a cluster in qa/1602 to test multi-node again if you wish.

andreasgerstmayr:
rebased, added more diagnostics + qa updates

qa/1602
natoscott: @andreasgerstmayr great!

goodwinos:
Hi Andreas, with the new vendor directory, we now have three versions of dict.[ch]. Which one should/will be used in libpcp_web? The src/external version is copyright 2012 whereas the hiredis and hiredis-cluster versions seem to be older and different (all three are different).

Should the new vendor directory live below (say) src/external instead of in the top level?

Recent Fedora has hiredis-0.13.3-14.fc33.x86_64 .. should this be used on platforms where it is available in preference to the vendored version? Note Fedora doesn't seem to have the hiredis-cluster package. Neither seems to be available on RHEL.

qa/1603 enables the cluster option with redis-server ... --cluster-enabled yes. I'm confused about which parts of the existing redis package and the new vendored hiredis package will be used to enable and use cluster mode.

I guess I'll have a bunch more questions once I've built a test cluster .. e.g. how well does the existing redisSlots stuff work with the commonly called redisSlotsRequest() function, e.g. all throughout query.c and schema.c?

andreasgerstmayr:
> Hi Andreas, with the new vendor directory, we now have three versions of dict.[ch]. Which one should/will be used in libpcp_web? The src/external version is copyright 2012 whereas the hiredis and hiredis-cluster versions seem to be older and different (all three are different).

Yep, that was a big trouble initially - I wanted to accommodate both library versions, but C doesn't support namespaces. I resolved this now by symlinking the files we want (i.e. *except* dict/sds) into src/libpcp_web/src/deps/, so we only have a single dict and sds version in use - the one from src/external. It's the same approach we're using for $(XFILES), the only difference is that it preserves the folder structure.

> Should the new vendor directory live below (say) src/external instead of in the top level?

I like the vendor/website/namespace/project directory structure - it shows where the code is coming from, that we're vendoring this project, i.e. it's external, and it is *not* source code created/maintained by PCP. Coincidentally it's the same directory structure Go uses for vendoring ;)

> Recent Fedora has hiredis-0.13.3-14.fc33.x86_64 .. should this be used on platforms where it is available in preference to the vendored version? Note Fedora doesn't seem to have the hiredis-cluster package. Neither seems to be available on RHEL.

We cannot use it because of the dict/sds mismatch. It's slightly binary incompatible in various ways, and crashes silently. imho it will work properly only if we don't use dict/sds (or use the dict/sds version of hiredis instead of our newer version).

> qa/1603 enables the cluster option with redis-server ... --cluster-enabled yes. I'm confused about which parts of the existing redis package and the new vendored hiredis package will be used to enable and use cluster mode.

Cluster support is part of Redis, the `redis-cli --cluster-enabled yes` has nothing to do with hiredis-cluster. hiredis-cluster is a client to make use of the Redis cluster functionality, it does *not* enable cluster mode, cluster mode needs to be enabled in the Redis database directly (it's a feature of Redis). You can create a Redis cluster with `redis-cli --cluster-enabled yes` and use any client library (which has cluster support) with it.

> I guess I'll have a bunch more questions once I've built a test cluster .. e.g. how well does the existing redisSlots stuff work with the commonly called redisSlotsRequest() function, e.g. all throughout query.c and schema.c?

It's a good thing that redisSlotsRequest() is used to wrap all Redis requests - that's where I handle the new MMV metrics now :). After increasing the metrics, it calls hiredis-cluster. So all the redisSlots stuff is still there, but it's using hiredis-cluster in the backend.

andreasgerstmayr:
rebased on latest main branch to resolve a merge conflict, and squashed a few intermediate commits

PR #1289, opened 2021-05-17
docs: add sizing recommendations table to the Scaling doc

natoscott:
Adds a table with recommended deployment architectures based on
the numbers of hosts being monitored.  Clarifies the individual
architecture sections, particularly around 'fully distributed'.


andreasgerstmayr:
+1 on the architecture renamings and updates.

in the `pmlogger-farm.svg` image the `Redis cluster` text is not centered anymore, and I can't see any changes except that. Was that intended?

Speaking about the `Redis cluster` cloud icon, iirc there was a discussion about changing it to just `Redis` - I'm fine with that, PCP handles clustered & standalone Redis transparently.

Regarding the table: the logging interval is an important factor. For example, logging 100 pmcds once every 10 seconds or once every 60 seconds makes a difference (the first one is unstable on a pmlogger farm, the latter option with 60s interval works fine). Not sure how to get this information into the table - maybe a new column with "metric throughput"? For example 120 pmcds logging every 60 seconds gives us the same amount of metric value updates (and processed Redis commands/second and required storage) as 20 pmcds logging every 10 seconds. Maybe we need a diagram to illustrate that (?)

natoscott:
> +1 on the architecture renamings and updates.
> 
> in the `pmlogger-farm.svg` image the `Redis cluster` text is not centered anymore, and I can't see any changes except that. Was that intended?

Hmm, I'm confused - pmlogger-farm.svg no longer contains the text 'Redis cluster' anywhere and the new text ('Redis Instance') is centered for me.  Tried using inkscape and eog to view just now to check and both look good.

> Regarding the table: the logging interval is an important factor. 

Yep - one of several important factors.  The number of metrics has a similar impact, as does the choice of metrics (large vs small indoms), as do the physical attributes of the actual machine(s) being logged.  I think these factors are beyond the scope of the initial table though, which is just trying to give an initial back-of-the-envelope kind of recommendation and not be all-encompassing.  The rest of the document covers those finer details well I think.

> Not sure how to get this information into the table

I don't think it belongs in the new table.

> - maybe a new column with "metric throughput"? For example 120 pmcds logging every 60 seconds gives us the same amount of metric value updates (and processed Redis commands/second and required storage) as 20 pmcds logging every 10 seconds.

Usually.  :)  Not all hosts are the same, so in certain situations even this column would provide incorrect information.

> Maybe we need a diagram to illustrate that (?)

Yeah, good idea - I think there's plenty of scope for more detailed explanations like this.  My intent with this change is to provide initial high level guidance though, so I'm reluctant to add anything else to the initial table due to the risk of causing confusion, or that might cause the primary message to get lost there.

andreasgerstmayr:
> > +1 on the architecture renamings and updates.
> > in the `pmlogger-farm.svg` image the `Redis cluster` text is not centered anymore, and I can't see any changes except that. Was that intended?
> 
> Hmm, I'm confused - pmlogger-farm.svg no longer contains the text 'Redis cluster' anywhere and the new text ('Redis Instance') is centered for me. Tried using inkscape and eog to view just now to check and both look good.

Me too ;) - what does https://github.com/performancecopilot/pcp/blob/aace082815e57c248842735263eee7e02d73246a/docs/HowTos/scaling/pmlogger-farm.svg (click on the document icon, next to the "Raw" button) show? I tried multiple browsers, and can only see the same image except the Redis Cluster text isn't centered anymore.

> > Regarding the table: the logging interval is an important factor.
> 
> Yep - one of several important factors. The number of metrics has a similar impact, as does the choice of metrics (large vs small indoms), as do the physical attributes of the actual machine(s) being logged. I think these factors are beyond the scope of the initial table though, which is just trying to give an initial back-of-the-envelope kind of recommendation and not be all-encompassing. The rest of the document covers those finer details well I think.

Hm yeah, that's also there ... my point was ppl shouldn't get the impression that a single pmlogger farm with 100 pmcds and a logging interval of 10 seconds works fine.

> > Not sure how to get this information into the table
> 
> I don't think it belongs in the new table.
> 
> > * maybe a new column with "metric throughput"? For example 120 pmcds logging every 60 seconds gives us the same amount of metric value updates (and processed Redis commands/second and required storage) as 20 pmcds logging every 10 seconds.
> 
> Usually. :) Not all hosts are the same, so in certain situations even this column would provide incorrect information.
> 
> > Maybe we need a diagram to illustrate that (?)
> 
> Yeah, good idea - I think there's plenty of scope for more detailed explanations like this. My intent with this change is to provide initial high level guidance though, so I'm reluctant to add anything else to the initial table due to the risk of causing confusion, or that might cause the primary message to get lost there.

+1

Let's merge it and refine as needed, wdyt @goodwinos?

goodwinos:
...
> > 
> > Yeah, good idea - I think there's plenty of scope for more detailed explanations like this. My intent with this change is to provide initial high level guidance though, so I'm reluctant to add anything else to the initial table due to the risk of causing confusion, or that might cause the primary message to get lost there.
> 
> +1
> 
> Let's merge it and refine as needed, wdyt @goodwinos?

yep +1. 

We need to run new benchmarks too (after many recent code changes for pcp-5.3.1, e.g. hiredis/cluster, discovery callback throttle and various bug fixes) but I think we can mostly expand on all of that in a v2 of the doc.

goodwinos:
Oh and also - regarding Marko's comments - when we do re-run, use more re-representtive pmcd hosts, e.g. smaller CPU and interrupt indoms, the effect of proc data, etc.

andreasgerstmayr:
Fun fact, https://raw.githubusercontent.com/performancecopilot/pcp/aace082815e57c248842735263eee7e02d73246a/docs/HowTos/scaling/pmlogger-farm.svg shows me the old version, but if I save this file to my disk and open it with the Gnome Image App it shows me a different picture (the one with "Redis Instance").

natoscott:
> Fun fact, https://raw.githubusercontent.com/performancecopilot/pcp/aace082815e57c248842735263eee7e02d73246a/docs/HowTos/scaling/pmlogger-farm.svg shows me the old version, but if I save this file to my disk and open it with the Gnome Image App it shows me a different picture (the one with "Redis Instance").

Looking into the svg file contents, I can indeed see both copies of the text - there's an XML entity named 'foreignObject' containing the Cluster text, followed immediately by a 'text' element containing the newer Instance text.  I'll edit the file directly I guess, let's see if that solves it.

PR #1298, opened 2021-05-21
pmrep: add RHEL8 compatible sar metricsets to default pmrep conf directory

shogo-matsumoto:
The output format of some metricsets has been changed to
follow the changes in sysstat v12. Leave the old sar-v11 output
format as well for RHEL8 users.


myllynen:
LGTM, thanks a lot!

PR #1301, opened 2021-05-25
redisSlotsRequest ENOMEM handling

goodwinos:
In redisSlotsRequest(), if redisSlotsReplyDataAlloc() fails and returns NULL the async redis request still goes ahead but with a NULL private data pointer passed down the call stack ... leading to badness (much later on) when the reply data is eventually dereferenced.

This PR instead returns ENOMEM to the caller for that case. Unfortunately however, none of the redisSlotsRequest() callers seem to check the return value .. there are over 50 of them and adding error handling for all of them would be a lot of work (exiting in the library call isn't an option).


andreasgerstmayr:
lgtm, thanks!

PR #1303, opened 2021-05-26
selinux: add capability2 { bpf }

andreasgerstmayr:
Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1952374

I've tested it with RHEL 7 (no BCC PMDA per default), RHEL 8 and  RHEL 9.

@natoscott Your earlier patch had special handling to check if the capabilities2 class exists, is that required? At first I thought it may be a problem to have
```
class capability2 { syslog };
class capability2 { bpf };
```
in the policy file (duplicate declaration of the capability2 class), but it doesn't seem to be a problem.


kmcdonell:
@andreasgerstmayr we do need the conditional code ... the class may not be defined or the permissions may not be defined for that class on some platforms.  We can really only use explicit classes, types and permissions if we think they are universally available.  For example permissions "syslog" (not to mention "bpf") are not defined for the "capability2" class on RHEL 6.

And as best I understand it, the following are all equivalent:
```
class capability2 { syslog };
class capability2 { bpf };
```
```
class capability2 syslog;
class capability2 bpf ;
```
```
class capability2 { syslog bpf };
```

andreasgerstmayr:
@kmcdonell: Sorry, I should have been more specific - I was referring to this part of Nathan's earlier patch:

```
+    seinfo -x --class=capability2 $seinfo_common_flag 2>/dev/null \
+    | grep '^[[ 	]][[ 	]]*(class |)capability2$' >/dev/null \
+    && pcp_selinux_capability2_class=true
```
i.e. testing for the `capability2` class itself, not testing for a specific element (bpf/syslog) of it. I was wondering if that's required, or if it's fine if we just check capability2 class + bpf element in one go (like we did with the syslog element of the capability2 class).

natoscott:
> @kmcdonell: Sorry, I should have been more specific - I was referring to this part of Nathan's earlier patch:
> 
> ```
> +    seinfo -x --class=capability2 $seinfo_common_flag 2>/dev/null \
> +    | grep '^[[ 	]][[ 	]]*(class |)capability2$' >/dev/null \
> +    && pcp_selinux_capability2_class=true
> ```
> 
> i.e. testing for the `capability2` class itself, not testing for a specific element (bpf/syslog) of it. I was wondering if that's required, or if it's fine if we just check capability2 class + bpf element in one go (like we did with the syslog element of the capability2 class).

AFAIK it's not required - just seemed like it might be the kinda thing missing on some platforms.  :P  ... but if its available everywhere there's no need for it.

kmcdonell:
@andreasgerstmayr and @natoscott I don't see this code in the tree, so I guess it was a proposed patch.  I'm not sure checking just the class alone makes much sense in this case.

kmcdonell:
ps @andreasgerstmayr I have some more selinux changes for the bcc PMDA on F33 ... I'll wait for this PR to get done before pushing them to avoid merge conflicts.

Issue #1296, opened 2021-05-20
Auto-start redis.service in pmproxy.service


martinpitt:
I am trying to show PCP metrics in Grafana. The documented way seem to be to enable pmproxy on the client side, and [the Grafana PCP plugin](https://grafana-pcp.readthedocs.io/en/latest/) on the Grafana side. This works fine by and large, but there is one step that seems a bit non-obvious: For `systemctl enable --now pmproxy` to actually work, you currently also need `systemctl enable --now redis`.

This is not part of the [plugin's Quickstart](https://grafana-pcp.readthedocs.io/en/latest/quickstart.html) (only in the [PCP redis section](https://grafana-pcp.readthedocs.io/en/latest/datasources/redis.html)), nor is it mentioned at all on the [manpage](https://linux.die.net/man/1/pmproxy).

But moreover, as pmproxy seems to pretty much depend on redis anyway, as otherwise queries hang forever (took me quite a while to figure this out, but it's already reported in #1134), could this just happen automatically?

IMHO the most elegant way would be to add

    Wants=redis.service
    After=redis.service

into pmproxy.service's `[Unit]` section. Then this will gracefully handle the case that redis is not installed at all, but start it if available ("unbreak my setup").

This also avoids a really difficult decision for both humands and UIs like Cockpit when you disable pmproxy.service again -- should you disable redis as well? It's a shared service, so it's not clear whether it was previously enabled only for pmproxy, or for some other service as well. Hence it would be best to not statically enable/disable it in the first place, but treat this as a dependency.

martinpitt:
The most important question here is: is there a real-life use case where someone would use pmproxy *without* redis? I admittedly only looked at that case, I don't know if it's useful for anything else.

natoscott:
@martinpitt yes there definitely are such cases - the grafana-pcp 'PCP Vector' and 'PCP bpftrace' datasources operate entirely without the need for Redis, accessing live data from PCP (pmcd) only.

natoscott:
Also, the OpenMetrics export mechanism offered by pmproxy does not use Redis, nor does the PCP protocol proxying functionality ... so, there are quite a few deployment scenarios where Redis is not required.

martinpitt:
Right, I'm aware of  the Vector/bpftrace data sources, but I thought they wouldn't use pmproxy? At least the documentation does not mention this at all.

However, I just saw that in principle one can configure pmproxy to use a Redis on a remote host.

[...]

PR #1304, opened 2021-05-26
pmproxy: start pmproxy after Redis (if available)

andreasgerstmayr:
If pmproxy starts before Redis, it will disable time-series
functionality. This update makes sure that pmproxy starts after Redis
(only if Redis is available, otherwise there is no change in behavior).

Related: #1296


goodwinos:
Looks good to me, merged

PR #1307, opened 2021-05-27
Add a small clarification around pmrep and instances

mbaldessari:


myllynen:
I had to think a while why this works (it's processed as a regex). The non-regex way of targeting several instances was explained under the section describing `-i` so I now added this there as well:

```
       -i instances, --instances=instances
            Retrieve and report  only  the  specified  metric  instances.   By
            default  all initially present instances are reported, except when
            writing an archive (see -o), where also all future instances  will
            be reported.
...
            instances is a comma-separated list of one or more instance filter
            specifications.  Filters containing commas or whitespace  must  be
            quoted with single (') or double (") quotes.  Note that as part of
            a metricspec on command line a list with more than one filter both
            the  list  and  each  filter must be quoted as shown below.  It is
            also possible to define a single filter with bars (|) as  instance
            separating regex in order to make quoting easier, see below.

            Multiple  -i options are allowed as an alternative way of specify‐
            ing more than one non-metric-specific instance filters.
...
            As an example, the following would report the same instances:

                 $ pmrep -i '. minute' kernel.all.load
                 $ pmrep -i '1 minute','5 minute' kernel.all.load
                 $ pmrep -i "'1 minute','5 minute'" kernel.all.load
                 $ pmrep -i '1 minute' -i '5 minute' kernel.all.load
                 $ pmrep kernel.all.load,,"'1 minute','5 minute'"
                 $ pmrep kernel.all.load,,'1 minute|5 minute'
...
```

What do you think, is this helpful enough now?

Thanks.

mbaldessari:
Aye, works for me. Thanks, Marko

I'd still add this example as well (as examples at the end are the first thing I look in a manpage), but do not feel strongly about it. I.e. we can drop this one as well.

myllynen:
You're right, nobody wants to read a dozen A4 long man page if they can get the info they need by quickly checking few examples. I'll merge (and tweak a bit) this now. Thanks for raising this issue & PR!

PR #1310, opened 2021-05-31
Improve flexibility to define new metrics and monitor replication

rmahique:
Make it more flexible to support the definition of metrics and sources by using the configuration file, other changes to make it more flexible, default metrics remain

Signed-off-by: Raul Mahiques <rmahique@redhat.com>


rmahique:
.

natoscott (approved):
LGTM - Marko? - is this suitable for Friday's pcp-5.3.1 release?
Also, is the MIME::Base64 module likely to be available by default on all platforms? (do we need any new deb / rpm spec dependencies)

myllynen:
This looks good to me, although I don't have access anymore to environment where I could test this.

Apart from a few completely unsubstantial cosmetics the only thing I'd like to see improved is a bit of documentation, few sentences either on the man page or in the example conf file or both. @rmahique do you think you could add few sentences as a follow-up before Friday if we merge this?

Also, if using only the default metrics is it so that the metrics (metadata and names) don't change before/after this patch? Thanks.

PR #1312, opened 2021-05-31
qa/1573: increase max pmproxy RSS by 1 MB

andreasgerstmayr:
on Fedora 34 pmproxy uses slightly above 10 MB RSS.

### DIFF
--- a/qa/1573
+++ b/qa/1573
@@ -2,7 +2,7 @@
 # PCP QA Test No. 1573
 # Exercise libpcp_web memory leak without a redis-server.
 #
-# Copyright (c) 2020 Red Hat.
+# Copyright (c) 2021 Red Hat.
 #
 
 seq=`basename $0`
@@ -101,7 +101,7 @@ pmproxy_rss2=`pminfo -f proc.memory.rss |
 
 echo === checking rss within tolerance
 _within_tolerance "rss" $pmproxy_rss1 $pmproxy_rss2 10%
-[ $pmproxy_rss2 -gt 10000 ] && echo "Unexpected pmproxy RSS: $pmproxy_rss2, was initially $pmproxy_rss1"
+[ $pmproxy_rss2 -gt 11000 ] && echo "Unexpected pmproxy RSS: $pmproxy_rss2, was initially $pmproxy_rss1"
 
 echo "RSS1 for PID $pmproxy_pid is $pmproxy_rss1" >> $here/$seq.full
 echo "RSS2 for PID $pmproxy_pid is $pmproxy_rss2" >> $here/$seq.full


natoscott:
@andreasgerstmayr I'd really like to understand why so much memory is being used rather than just ignoring it.  This is (meant to be) a bare-bones pmproxy - it intrigues me how it can use ~3MB more (resident) than pmcd (which is actually doing work) ... where can that memory be in use I wonder?

myllynen:
Now that you're investigating memory usage in detail in a real-world case, could you please check the current pmrep per-process memory related metricsets, are those helpful and would you keep those metricsets as-is or tweak them somehow? See /etc/pcp/pmrep/proc.conf for details. Thanks.

goodwinos:
Related: #1318 

Regarding the pmap listing, this is basically reformatted from /proc/PID/maps. And those mappings are generally libs shared between many processes. A more informative snapshot is from /proc/PID/status, e.g. 

```
$ egrep '^Name|^Vm|^Rss' /proc/$(pidof pmproxy)/status
Name:	pmproxy
VmPeak:	  157420 kB
VmSize:	  148864 kB
VmLck:	       0 kB
VmPin:	       0 kB
VmHWM:	   60744 kB
VmRSS:	   60744 kB
RssAnon:	   52604 kB
RssFile:	    8140 kB
RssShmem:	       0 kB
VmData:	   65396 kB
VmStk:	     300 kB
VmExe:	      76 kB
VmLib:	   11188 kB
VmPTE:	     188 kB
VmSwap:	       0 kB
```

And also /proc/PID/statm (which is cheaper to scrape, as proposed for issue #1318)

```
$ cat /proc/$(pidof pmproxy)/statm
37216 15186 2035 19 0 16424 0
```

natoscott:
I had a bit of a look into this issue today, found a few things but but didn't get to the bottom of it.

Comparing pmcd to pmproxy is quite enlightening.  One area that's quite different is pmproxy has at least an extra ~750K RSS from "libcrypto.so.1.1.1k" ... for no apparent reason.  If that's happening in a few libraries, it'd explain the difference - its not the only thing though, there's plenty of anon memory and other mappings.

```
$ sudo pmap -x $(pidof pmproxy) | sort -k 3 -nr | grep libcrypto
00007f1bf0446000     488     488       0 r---- libcrypto.so.1.1.1k
00007f1bf04c0000    1720     248       0 r-x-- libcrypto.so.1.1.1k
00007f1bf0700000     176     176     176 r---- libcrypto.so.1.1.1k
00007f1bf066e000     584     128       0 r---- libcrypto.so.1.1.1k
00007f1bf072c000      16      16      16 rw--- libcrypto.so.1.1.1k
```

versus

```
sudo pmap -x $(pidof pmcd) | sort -k 3 -nr | grep libcrypto
00007f308c41f000     176     176     176 r---- libcrypto.so.1.1.1k
00007f308c165000     488     128       0 r---- libcrypto.so.1.1.1k
00007f308c44b000      16      16      16 rw--- libcrypto.so.1.1.1k
00007f308c38d000     584       0       0 r---- libcrypto.so.1.1.1k
00007f308c1df000    1720       0       0 r-x-- libcrypto.so.1.1.1k
```

One bug I found - pmproxy is ignoring secure.enabled and always attempts to enable TLS (fix incoming) - but even with that fixed, it doesn't change the situation compared to pmcd.

goodwinos:
Here's some more analysis FWIW. I'd agree with Andreas, most of this is the 42 shared libraries that pmproxy is linked with. On f34, it's _also_ linked with libpcap for some reason, which pushes it over the threshold causing qa/1573 to fail.

Running on f33, with the following minimal pmproxy.conf:
```
[pmproxy]
pcp.enabled = true
http.enabled = true
redis.enabled = true
secure.enabled = false
[discover]
enabled = true
[pmseries]
enabled = false
```

```
start and break in main: RSS is 6624. After entering the main loop at the end of this analysis, RSS is 9068.

USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root     3016875  0.0  0.0  18384  6624

enter ParseOptions, no change
enter openports, no change
enter server_init, not much change

in open_request_ports:
540	    if (__pmServerHasFeature(PM_SERVER_FEATURE_DISCOVERY))
541		server->presence = __pmServerAdvertisePresence(PM_SERVER_PROXY_SPEC, port);
causes VSZ to jump from 18384 to 92144 (big jump) but RSS was unchanged so the memory was not mapped.

enter setup_secure_module:
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root     3140348  0.0  0.0  92144  6708

enter setup_redis_module:
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root     3140348  0.0  0.0  92144  6708

stepping thru setup_redis_module:

205		proxy->slots = redisSlotsConnect(proxy->config ...)
...
164	    if ((slots->acc = redisClusterAsyncContextInit()) == NULL) {

Both VSZ and RSS growth :
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root     3026934  0.0  0.0  92276  8332
...
215	    sts = redisClusterConnect2(slots->acc->cc);
   1605	    return cluster_update_route(cc);
       1507	        ret = cluster_update_route_by_addr(cc, node->host, node->port);
           1291	        c = redisConnectWithTimeout(ip, port, *cc->connect_timeout);
	       852	    return redisConnectWithOptions(&options);
	           818	        redisContextConnectBindTcp(c, options->endpoint.tcp.ip,
		       544	    return _redisContextConnectTcp(...)
		           60	      && __nss_database_lookup2 (DATABASE_NAME_STRING,...
			       190	  *fctp = __nss_lookup_function (*ni, fct_name);
			           448		  if (nss_load_library (ni) != 0)
				       351	      ni->library->lib_handle = __libc_dlopen (shlib_name);
				       (gdb) p shlib_name
                                       $10 = 0x7ffffffdc290 "libnss_files.so.2"
After loading that shared library,
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root     3139280  0.0  0.0  92644  9080

(gdb) where
#0  nss_load_library (ni=ni@entry=0x555555587460) at nsswitch.c:321
#1  0x00007ffff79ac9f9 in __GI___nss_lookup_function (ni=0x555555587460, fct_name=<optimized out>, fct_name@entry=0x7ffff7a18dd7 "gethostbyname2_r") at nsswitch.c:448
#2  0x00007ffff79acb90 in __GI___nss_lookup (ni=ni@entry=0x7ffffffdc3c8, fct_name=fct_name@entry=0x7ffff7a18dd7 "gethostbyname2_r", fct2_name=fct2_name@entry=0x0, fctp=fctp@entry=0x7ffffffdc3d0) at nsswitch.c:190
#3  0x00007ffff79ade27 in __GI___nss_hosts_lookup2 (ni=ni@entry=0x7ffffffdc3c8, fct_name=fct_name@entry=0x7ffff7a18dd7 "gethostbyname2_r", fct2_name=fct2_name@entry=0x0, fctp=fctp@entry=0x7ffffffdc3d0) at XXX-lookup.c:66
#4  0x00007ffff799a238 in __gethostbyname2_r (name=name@entry=0x555555585f30 "localhost", af=af@entry=2, resbuf=resbuf@entry=0x7ffffffdc520, buffer=<optimized out>, buflen=1024, result=result@entry=0x7ffffffdc518, h_errnop=0x7ffff6e4b87c) at ../nss/getXXbyYY_r.c:269
#5  0x00007ffff79722d7 in gaih_inet (name=name@entry=0x555555585f30 "localhost", service=service@entry=0x7ffffffdc720, req=req@entry=0x7ffffffdcd30, pai=pai@entry=0x7ffffffdc708, naddrs=naddrs@entry=0x7ffffffdc704, tmpbuf=tmpbuf@entry=0x7ffffffdc800) at ../sysdeps/posix/getaddrinfo.c:559
#6  0x00007ffff7972e69 in __GI_getaddrinfo (name=<optimized out>, service=<optimized out>, hints=0x7ffffffdcd30, pai=0x7ffffffdcd28) at ../sysdeps/posix/getaddrinfo.c:2244
#7  0x00007ffff7bcceb8 in _redisContextConnectTcp (c=0x555555585850, addr=0x555555580861 "localhost", port=6379, timeout=0x7ffffffdce20, source_addr=0x0) at deps/hiredis/net.c:425
#8  0x00007ffff7bcd3dd in redisContextConnectBindTcp (c=0x555555585850, addr=0x555555580861 "localhost", port=6379, timeout=0x7ffffffdce20, source_addr=0x0) at deps/hiredis/net.c:544
#9  0x00007ffff7bcb0cb in redisConnectWithOptions (options=0x7ffffffdce40) at deps/hiredis/hiredis.c:818
#10 0x00007ffff7bcb29a in redisConnectWithTimeout (ip=0x555555580861 "localhost", port=6379, tv=...) at deps/hiredis/hiredis.c:852
#11 0x00007ffff7bd57dc in cluster_update_route_by_addr (cc=0x7ffff7fa8010, ip=0x555555580861 "localhost", port=6379) at deps/hiredis-cluster/hircluster.c:1291
#12 0x00007ffff7bd5f8b in cluster_update_route (cc=0x7ffff7fa8010) at deps/hiredis-cluster/hircluster.c:1507
#13 0x00007ffff7bd62e1 in _redisClusterConnect2 (cc=0x7ffff7fa8010) at deps/hiredis-cluster/hircluster.c:1605
#14 0x00007ffff7bd6e0f in redisClusterConnect2 (cc=0x7ffff7fa8010) at deps/hiredis-cluster/hircluster.c:2011
#15 0x00007ffff7bbc301 in redisSlotsInit (config=0x55555557a900, events=0x7ffff7b810e0 <default_loop_struct>) at slots.c:215
#16 0x00007ffff7bb0b0f in redisSlotsConnect (config=0x55555557a900, flags=SLOTS_KEYMAP, info=0x55555555d70d <proxylog>, done=0x555555563f8e <on_redis_connected>, userdata=0x55555557fcc0, events=0x7ffff7b810e0 <default_loop_struct>, arg=0x55555557fcc0) at schema.c:1452
#17 0x0000555555564457 in setup_redis_module (proxy=0x55555557fcc0) at redis.c:205
#18 0x0000555555564564 in redis_reconnect_worker (arg=0x55555557fcc0) at redis.c:227
#19 0x00007ffff7be83ef in pmWebTimerRegister (callback=0x555555564466 <redis_reconnect_worker>, data=0x55555557fcc0) at timer.c:79
#20 0x000055555555f6c6 in setup_proxy (arg=0x7fffffffd340) at server.c:786
#21 0x00007ffff7b5c076 in uv__run_timers (loop=loop@entry=0x7ffff7b810e0 <default_loop_struct>) at src/timer.c:178
#22 0x00007ffff7b638da in uv_run (loop=0x7ffff7b810e0 <default_loop_struct>, mode=UV_RUN_DEFAULT) at src/unix/core.c:376
#23 0x000055555555f881 in main_loop (arg=0x55555557fcc0) at server.c:836
#24 0x000055555555cfcb in main (argc=5, argv=0x7fffffffe578) at pmproxy.c:467

```

So redis context and the 16384 slots are using additional memory. And the nss library, which may be being dlopen'd unnecessarily if we're not connecting to redis over ssl.

natoscott:
> start and break in main: RSS is 6624. After entering the main loop at the end of this analysis, RSS is 9068.
> 

OK, so that's the mysterious 3MB diff with pmcd ... good progress but root cause is still not really understood though.  Some more notes and questions to help progress further:

> __pmServerAdvertisePresence(PM_SERVER_PROXY_SPEC, port);
> causes VSZ to jump from 18384 to 92144 (big jump) but RSS was unchanged so the memory was not mapped.
> 

Good to know - this part should be improved with your (separate, planned) Avahi change?

> ...
> 164	    if ((slots->acc = redisClusterAsyncContextInit()) == NULL) {
> 
> Both VSZ and RSS growth :
> USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
> root     3026934  0.0  0.0  92276  8332
> ...

OK, there's a ton of bloat somewhere in there for sure.  This is a change since we moved to hiredis-cluster I think.  Previously our slots.c code used a binary tree populated with only elements for cluster nodes that actually exist (hence, small for the default one-node, non-cluster case).  I wonder if the hiredis-cluster library is particularly inefficient in memory use for the non-cluster case?

> 215	    sts = redisClusterConnect2(slots->acc->cc);

(IOW, until recently we didn't make the above call at all)

>                                        $10 = 0x7ffffffdc290 "libnss_files.so.2"
> After loading that shared library,
> USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
> root     3139280  0.0  0.0  92644  9080

This nss library looks like noise to me? - just a small number of Kb.  It seems that the new cluster code is the most likely root cause from these notes.

> So redis context and the 16384 slots are using additional memory. And the nss library, which may be being dlopen'd unnecessarily if we're not connecting to redis over ssl.

The redis context structure itself is also noise...

```
(gdb) p sizeof(redisAsyncContext)
$1 = 448
```

Did you find a data structure expanded across all 16384 slots?  That could be an interesting one - wouldn't need to be super large to consume alot of the missing 2-3MB RSS.

cheers.

goodwinos:
> > start and break in main: RSS is 6624. After entering the main loop at the end of this analysis, RSS is 9068.
> 
> OK, so that's the mysterious 3MB diff with pmcd ... good progress but root cause is still not really understood though. Some more notes and questions to help progress further:
> 
> > __pmServerAdvertisePresence(PM_SERVER_PROXY_SPEC, port);
> > causes VSZ to jump from 18384 to 92144 (big jump) but RSS was unchanged so the memory was not mapped.
> 
> Good to know - this part should be improved with your (separate, planned) Avahi change?

yes - I've updated BZ 1899625 with some notes - both pmcd and pmproxy will benefit not just from reduces VSZ but also by dropping the unwanted systemd Before= setting.

> > ...
> > 164	    if ((slots->acc = redisClusterAsyncContextInit()) == NULL) {
> > Both VSZ and RSS growth :
> > USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
> > root     3026934  0.0  0.0  92276  8332
> > ...
> 
> OK, there's a ton of bloat somewhere in there for sure. ... I wonder if the hiredis-cluster library is particularly inefficient in memory use for the non-cluster case?

Seems to be - effectively treated as cluster of only one node, but I don't think there's much we can do about it without a lot of effort for not much gain.

> > ```
> >                                    $10 = 0x7ffffffdc290 "libnss_files.so.2"
> > ```
> > After loading that shared library,
> > USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
> > root     3139280  0.0  0.0  92644  9080
> 
> This nss library looks like noise to me? - just a small number of Kb. It seems that the new cluster code is the most likely root cause from these notes.

it's clearly not huge. my comment was more along the lines of why does it need to be loaded at all?

> 
> > So redis context and the 16384 slots are using additional memory. And the nss library, which may be being dlopen'd unnecessarily if we're not connecting to redis over ssl.
> 
> The redis context structure itself is also noise...
> 
> ```
> (gdb) p sizeof(redisAsyncContext)
> $1 = 448
> ```

yes it's small, but the redisClusterContext structure is much bigger
(gdb) p sizeof(redisClusterContext)
$1 = 131792

> 
> Did you find a data structure expanded across all 16384 slots? That could be an interesting one - wouldn't need to be super large to consume alot of the missing 2-3MB RSS.

yes. It's the redisClusterContext.table field. The structure containing that field is calloc'd in redisClusterContextInit(). The table array is REDIS_CLUSTER_SLOTS = 16384 pointers, which is 128K. It's initialized by calloc (and again later the table[] array is zeroed with a memset, unnecessarily). Even though it's only 128K, after that call RSS went up by 1648K. Not sure why, guess it's something related to heap management. Following that are numerous smaller RSS bumps involving binding the socket and actual redis connection, e.g. redisContextConnectBindTcp(), redisConnectWithTimeout(),  and getaddrinfo(). 

Other than the avahi change (which will reduce VSZ but not RSS), it doesn't look like there's much we can really do for this .. and we're only talking about less than 10MB anyway.

natoscott:
> > 
> > Good to know - this part should be improved with your (separate, planned) Avahi change?
> 
> yes - I've updated BZ 1899625 with some notes - both pmcd and pmproxy will benefit not just from reduces VSZ but also by dropping the unwanted systemd Before= setting.
> 

Good stuff.

> > > ...
> > > 164	    if ((slots->acc = redisClusterAsyncContextInit()) == NULL) {
> > 
> > OK, there's a ton of bloat somewhere in there for sure. ... I wonder if the hiredis-cluster library is particularly inefficient in memory use for the non-cluster case?
> 
> Seems to be - effectively treated as cluster of only one node, but I don't think there's much we can do about it without a lot of effort for not much gain.
> 

Let's see what we can do here, especially since this is a regression and we know the cause now.

It seems to me there's no need to be calling that function at all for most people (most people don't use clusters) - maybe we can auto-detect cluster mode like we do with redisearch, and just use a regular async, non-cluster context in the default case?  That doesn't seem like alot of effort.

While you're away I'll look into it some more with @andreasgerstmayr and see what we can come up with.

andreasgerstmayr:
> It seems to me there's no need to be calling that function at all for most people (most people don't use clusters) - maybe we can auto-detect cluster mode like we do with redisearch, and just use a regular async, non-cluster context in the default case? That doesn't seem like alot of effort.
> 
> While you're away I'll look into it some more with @andreasgerstmayr and see what we can come up with.

afaics we have two options here: either update the memory allocation strategy in hiredis-cluster upstream, or use either hiredis or hiredis-cluster in PCP depending if cluster mode is enabled.

The main annoyance of the second solution are the callbacks - for hiredis-cluster the redis response callbacks get `redisClusterAsyncContext`, for hiredis they get `redisAsyncContext`. So we'd need a new type (possibly a struct containing a union of both, and a field whether cluster mode is enabled or not), update all response handlers to use the new struct, and also update https://github.com/performancecopilot/pcp/blob/f59d5194a0d82225762e7b561af2615c310ca8b5/src/libpcp_web/src/slots.c#L286-L303 to allocate this new struct, run the callback, and free the struct.

natoscott:
I made some upstream changes over in https://github.com/Nordix/hiredis-cluster/pull/64

This is a handy improvement, but having spent time on this now I see it's not the biggest improvement we can make here after all.  I found the pmSetProcessIdentity call to initgroups(3), in libpcp, of all things, to be problematic with current Fedora/glibc.  Optimizing this (trivially) gives back a whopping single-line-change return of >1MB RSS in pmproxy, and makes this test pass reliably again.  Better yet, this is a win for every PCP daemon, including PMDAs.  Even pmcd is lighter weight than before (sub-7MB RSS there now).

See https://github.com/performancecopilot/pcp/pull/1411 for details.  With these changes the test is reliably passing once more and pmproxy memory use is well below the test limit, so I'm closing out this issue finally. 🎉

natoscott:
| Running on f33, with the following minimal pmproxy.conf:
^^^^^^^^^^^^^^^^^

@goodwinos argh!  This is why your analysis and results were not reproducible for me on f34 where this test is failing.  See the very first comment from @andreasgerstmayr in this issue.

andreasgerstmayr:
>  I found the pmSetProcessIdentity call to initgroups(3), in libpcp, of all things, to be problematic with current Fedora/glibc. Optimizing this (trivially) gives back a whopping single-line-change return of >1MB RSS in pmproxy, and makes this test pass reliably again.

Interesting, I wonder why this call is using 1MB+? The `/etc/group` file has 1166 bytes on my system, and I have 83 groups in that file. Or maybe it's allocating that much memory, and free'ing it after, but glibc doesn't return the memory back to the system, and PCP doesn't allocate more memory after that, so we see the increased RSS usage.

Anyway, I'm glad this issue is resolved. Thanks for your investigations!

natoscott:
@andreasgerstmayr stab in the dark without looking at the code, I'd punt there's memory being allocated for a cache of identity information from NIS (even if its not in use).

PR #1313, opened 2021-06-01
selinux: use new logging_watch_all_log_dirs_path macro

natoscott:
Allow PMDAs that access log files to do so when latest versions
of the kernel and selinux-policy is installed.

This resolves a new qa/652 (pmdasystemd) failure amongst others.

### DIFF
--- a/configure
+++ b/configure
@@ -713,6 +713,7 @@ fmt_uint64
 fmt_int64
 fmt_pid
 rdynamic_flag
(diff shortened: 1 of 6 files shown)


natoscott:
This fix is held up by a problem (I think) in the base selinux-policy definitions:
https://bugzilla.redhat.com/show_bug.cgi?id=1966397

Issue #1318, opened 2021-06-03
add pmproxy resource usage metrics and a new uv timer for periodic stats updates in libpcp_web


goodwinos:
We need metrics for monitoring pmproxy dynamic memory usage. Both globally (i.e. pmproxy process vsize and rss from /proc/PID/statm) and per module where appropriate via new mmv_stats instrumentation - webgroup, series, redis, discover. These should have instant semantics (increase on alloc, decrease on free). Free can be problematic because the size of the free'd block might not always be readily available.

In any case, global pmproxy.mem.rss and vsize metrics will be immediately useful for correlating process size with various workloads (especially on large logger farms) and for tracking down leaks and dict map blowouts.

PR #1319, opened 2021-06-03
pmproxy: add pmproxy.mem.rss and pmproxy.mem.vsz metrics

goodwinos:
Adds pmproxy metrics for virtual size and resident set size.
This will be useful for correlating pmproxy memory growth
with other metrics.

Also updates pmrep config for ':pmproxy' to include these in
the first two columns.

QA is still todo (would just verify against ps listing).

Related: https://github.com/performancecopilot/pcp/issues/1318

### DIFF
--- a/src/pmproxy/src/server.c
+++ b/src/pmproxy/src/server.c
@@ -60,6 +60,42 @@ proxylog(pmLogLevel level, sds message, void *arg)
     pmNotifyErr(priority, "%s%s", state, message);
 }
 
+static void
+refresh_proxy_mem_metrics(struct proxy *proxy)
+{
+    pmAtomValue		*value;
+    uint64_t		vsz, rss;
+    int			fd;
+    char		buf[MAXPATHLEN];
+    static		pid_t pid = 0;
+    static		int pagesz = 0;
+
+    if (!pid)
+    	pid = getpid();
+    sprintf(buf, "/proc/%u/statm", pid);
+    if ((fd = open(buf, O_RDONLY)) >= 0) {
+    	if (read(fd, buf, sizeof(buf)) > 0) {
+	    if (sscanf(buf, "%lu %lu", &vsz, &rss) != 2) {
+		close(fd);
+	    	return;
+	    }
+	}
+    }
+    close(fd);
+
+    /* proc data is in units of pages */
+    if (!pagesz)
+    	pagesz = getpagesize();
+    vsz = vsz * pagesz / 1024;
+    rss = rss * pagesz / 1024;
+
+    if ((value = mmv_lookup_value_desc(proxy->metrics_handle, "mem.vsz", NULL)) != NULL)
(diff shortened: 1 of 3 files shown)


src/pmproxy/src/server.c
```diff
@@ -60,6 +60,42 @@ proxylog(pmLogLevel level, sds message, void *arg)
...
+    pmAtomValue		*value;
+    uint64_t		vsz, rss;
+    int			fd;
+    char		buf[MAXPATHLEN];
+    static		pid_t pid = 0;
+    static		int pagesz = 0;
+
+    if (!pid)
+    	pid = getpid();
+    sprintf(buf, "/proc/%u/statm", pid);
```
natoscott, line 75: This is Linux-specific so I guess this whole function should be conditionally compiled depending on platform.

src/pmproxy/src/server.c
```diff
@@ -828,6 +877,7 @@ check_proxy(uv_check_t *arg)
     struct proxy	*proxy = (struct proxy *)handle->data;
 
     flush_secure_module(proxy);
+    refresh_proxy_mem_metrics(proxy);
```
natoscott, line 880: This looks like its going to get updated once per loop iteration - that seems like *alot* ... ?
How about a stats timer instead, perhaps updating this (and similar in future) values once
every few seconds instead?

natoscott (changes requested):

src/pmproxy/src/server.c
goodwinos: I was just going with minimal changes if we want to take this for pcp-5.3.1. But yes it would be frequently updated. For this sort of thing it would be better if libpcp_mmv could register a fetch callback, so the value would only be computed on demand (and always be current rather than up to a few seconds old)

natoscott: > I was just going with minimal changes if we want to take this for pcp-5.3.1.

There's no rush on this AFAICS - maybe merge with those other pmproxy fixes (post-release)?

>  But yes it would be frequently updated.

Well, that's not good - definitely not appropriate for tomorrows release then.

> For this sort of thing it would be better if libpcp_mmv could register a fetch callback [...]

That's not feasible - libpcp_mmv has no knowledge of when fetch happens (by design).

A timer is the only viable option here and its just fine as an approach - you could even piggyback it onto the libpcp_mmv garbage collection timer (or perhaps add an interface to libpcp_web for registering timer-driven callbacks - an interface used by separate GC and metrics callbacks).

This code could live more neatly in libpcp_web then, instead of in the midst of not-really-related code in pmproxy.  We could trivially extend this timer-driven mechanism to also call getrusage and add accurate CPU utilization metrics too.  Hmm, in fact ... getrusage(2) seems to have rss stats, and it's platform-independent, so perhaps the one call could suffice (no vsize though - perhaps just that part becomes Linux-specific).

goodwinos:
abandoning this PR. The work will be done as part of #1318

PR #1334, opened 2021-06-24
Automatic cpu core detection for pmda-denki

christianhorn:
- remove command line option for cpus
- autodetect for up to 4096 cores, current linux kernel limit
- "intel rapl" -> "rapl", as also newer AMD cpus support it

### DIFF
--- a/qa/1653
+++ b/qa/1653
@@ -47,7 +47,7 @@ echo -e "\n### Installing test dataset 1...\n"
 cd $PCP_PMDAS_DIR/denki
 ROOTPATH="$here/denki/Thinkpad_L480_bat_busy"
 $sudo sed -ie \
-	"s,^args.*,args=\"-U root -c 4 -r ${ROOTPATH}\","  \
+	"s,^args.*,args=\"-U root -r ${ROOTPATH}\","  \
 	Install
 $sudo ./Install 2>&1
 pminfo -fF denki
@@ -55,7 +55,7 @@ pminfo -fF denki
 echo -e "\n### Installing test dataset 2...\n"
 ROOTPATH="$here/denki/bigsystem"
 $sudo sed -ie \
-	"s,^args.*,args=\"-U root -c 32 -r ${ROOTPATH}\","  \
+	"s,^args.*,args=\"-U root -r ${ROOTPATH}\","  \
 	Install
 $sudo ./Install 2>&1
 pminfo -fF denki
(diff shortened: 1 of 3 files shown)


natoscott:
Hmm, this seems like a step backward - why hard code the number of CPUs instead of just asking for the correct value from the  system?  It breaks our code in subtle ways when the kernel improves and any CPU count limit is increased.  It also means we have to traverse loops thousands of times on systems with small CPU counts.

The command line option is for QA only (like -r) - its not something regular users have to use.

christianhorn:
The suggested implementation is not hardcoded, it sets 4096 as absolute maximum (and we can change that easily, even now while the kernel does not yet support it).

> It breaks our code in subtle ways when the kernel improves and any CPU count limit is increased.
I think we could replace it with the maximum value for integer without issues. Ageed, this pull request would not deal with >4096 cores.

> It also means we have to traverse loops thousands of times on systems with small CPU counts.
The detection function runs exactly one time, it has a loop which runs potentially from 0 to MAX_CPUS, but the first cpu-number which is not found leads to the end of the loop. I confirmed with a debug statement inside the loop.

> The command line option is for QA only (like -r) - its not something regular users have to use.
I did only look at the current build with the -r and test data, not on a real 32core system. But I think it would also on the real 32core system need the -c 32.

Let me confirm behavior of the current code on a multi-numa-zone/32+ core system, will report.

christianhorn:
Some news:
- our core limits are higher as I had looked up, theoretically ~8k cores
- current code works on live systems also with >1 rapl packages. The system I tested on is AMD EPYC, I also verified that these provide RAPL on rhell8 (so kernel 4 with ports from newer upstream). Impressive to see 2 packages humming away with 60W consumption each, compared to laptops.
- But I still feel bad to use sysconf(_SC_NPROCESSORS_CONF) for live systems and command line cores for QA, having as much code shared for both usages would be good.

I consider 2 pathes:
- You used "static int ncpus; ncpus =  sysconf(_SC_NPROCESSORS_CONF)", so that would cover 2^32, which we could set for MAX_CPUS. That would allow maximum code sharing reallife and QA code, and not decrease the limit of cpus. It seems like sysconf() returns long int.. but with all future proofness, assuming 2^128 might be to much :)
- I also considered to read the directory /sys/devices/system/cpu/ and evaluate the biggest cpu* entry.. but just setting MAX_CPUS is simpler.

christianhorn:
This commit should remove the downsides.
- It has no regression regarding number of supported cpus, ncpus was "int" and MAX_CPUS is unsigned int
- No loops are wasted, the detection loop starts to count with 0 and tests one-by-one if the next cpu exists, and leaves the loop with the first cpu which is not found.
- The commit allows now common cpu detection for both realworld and QA cases

PR #1337, opened 2021-07-01
pmdalinux: take buddyinfo's order into account in calculating total size

shogo-matsumoto:


natoscott:
@shogo-matsumoto thanks!  I found an issue while testing this relating to sign extension when moving from a signed 32 bit value to the unsigned long long used in the pmAtomValue for the total (bytes) metric.  This resulted in regression test qa/821 failing with these changes.  I went ahead and fixed it as best I could - please review my follow-up changes here?  Thanks!

```
$ ./check 821
821 26s ... - output mismatch (see 821.out.bad)
[...]
<     inst [21 or "DMA32::order10::node0"] value 2724
---
>     inst [21 or "DMA32::order10::node0"] value 18014398508077056
[...]
Failures: 821
Failed 1 of 1 tests
```

I believe the correct value there should be 2789376, but sign extension happened, i.e.

```
gdb -q
(gdb) p /t 18014398508077056
$1 = 111111111111111111111111111111111010101001000000000000
(gdb) 
```

shogo-matsumoto:
@natoscott Thank you for merging the PR. I didn't notice the overflow problem in my low memory environment. Your recent commit looks good to me. Thank you for your help!

PR #1342, opened 2021-07-08
Improve pmproxy and libpcp_web scalability

goodwinos:
5e3b792d3 libpcp_web: plug mem leak in redisMapInsert during daily log-rolling
2a00a90b0 libpcp_web/discovery: improve lock handling and scalability

earlier related commits are already in main branch

### DIFF
--- a/src/libpcp_web/src/maps.c
+++ b/src/libpcp_web/src/maps.c
@@ -160,6 +160,12 @@ redisMapLookup(redisMap *map, sds key)
 void
 redisMapInsert(redisMap *map, sds key, sds value)
 {
+    redisMapEntry *entry = redisMapLookup(map, key);
+
+    if (entry) {
+	/* fix for Coverity CID323605 Resource Leak */
+	dictDelete(map, key);
+    }
     dictAdd(map, key, value);
 }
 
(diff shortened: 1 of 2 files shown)


src/libpcp_web/src/maps.c
```diff
@@ -160,6 +160,12 @@ redisMapLookup(redisMap *map, sds key)
 void
 redisMapInsert(redisMap *map, sds key, sds value)
 {
+    redisMapEntry *entry = redisMapLookup(map, key);
+
+    if (entry) {
+	/* fix for Coverity CID323605 Resource Leak */
+	dictDelete(map, key);
+    }
```
natoscott, line 168: Looks like it should be using dictReplace instead of dictAdd, then no need for the extra lookup.

goodwinos: thanks for the review - I considered that, but it ment the caller would have to sdsfree the key if the entry already exists (since dictReplace updates the value and frees the old value, but doesn't replace the key) .. if the caller doesn't then we have a new leak that's just as bad as the existing one. And also redisMapInsert is void, so the caller doesn't get to find out if the key exists. So I went with dictDelete, which is a sure free of both the key and the value if the key already exists.

natoscott: Sounds good.

PR #1344, opened 2021-07-09
docs: update scaling doc with federated setup and troubleshooting chapter

andreasgerstmayr:


natoscott:
LGTM.

goodwinos:
Looks good Andreas - would it be worth adding a comparison between federated / non-federated results?

andreasgerstmayr:
> Looks good Andreas - would it be worth adding a comparison between federated / non-federated results?

Yep, I was about to write a sentence like "The observed values of each individual pmlogger farm of the federated deployment are almost identical to a single pmlogger farm, rendering the PCP deployment horizontally scalable" but it sounds too cheesy - do you have any better suggestions? :smile: 

The observed values of a non-federated farm with 100 loggers vs each farm of the federated results are almost identical, except of the network bandwidth (noted with "The network bandwidth is higher due to the inter-node communication of the Redis cluster.").

andreasgerstmayr:
I'll merge it for now, let's add any updates on a later commit.

PR #1346, opened 2021-07-12
qa: remove valgrind requirements from statsd tests that don't actually use it

Erbenos:
See title, was trying to run tests on a clean system, that's how I found out.


kmcdonell:
Thanks @Erbenos.  I think qa/1720 needs the same love and care, I'll fix that one.

kmcdonell:
Mea cupla, qa/1720 does use valgrind, but it is hidden in qa/statsd/cases/15.py.  So that one's OK.

PR #1348, opened 2021-07-14
docs: document hostspec parameter functionality for openmetrics endpoint

andreasgerstmayr:
Added (almost) the same doc text as in the PMAPI section.


natoscott (approved):
LGTM.

PR #1349, opened 2021-07-14
build: optionally enable sanitizers

andreasgerstmayr:
Very handy for data race detection and much more, see
https://github.com/google/sanitizers/wiki

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -37,6 +37,12 @@ AC_ARG_WITH([gperftools],
     [use_gperftools=$withval; PACKAGE_CONFIGURE="$PACKAGE_CONFIGURE --with-gperftools=$withval"],
     [use_gperftools=no])
 
+AC_ARG_WITH([sanitizer],
+    [AS_HELP_STRING([--with-sanitizer],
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -730,6 +736,15 @@ AS_IF([test "x$use_gperftools" != "xno"], [
...
 ])
 
+dnl Check for sanitizers
+if test "$use_sanitizer" = no
+then
+    sanitizer=""
+else
+    sanitizer="-fsanitize=$use_sanitizer"
+fi
+AC_SUBST(sanitizer)
```
natoscott, line 746: Cosmetic only, but maybe write this in fewer lines using this configure.ac pattern ...

```
sanitizer=""
if test "$use_sanitizer" = no; then
   sanitizer="-fsanitize=$use_sanitizer"
fi
AC_SUBST(sanitizer)
```

natoscott (approved):
LGTM - minor, very optional suggestion only. :)

configure.ac
andreasgerstmayr: done. thanks!

PR #1351, opened 2021-07-15
Add bpf pmda to read ELF-based bpf modules

jasonk000:
These provide an alternative implementation to the pmdabcc python module.

This results in a significant improvement in startup time and memory footprint for these tracing modules. Specifically this now runs in <6MB RSS instead of ~100+MB. The tradeoff is it is less dynamically configurable, however writing BCC and probes with python requires quite a bit of expertise, and is not significantly more complex to build a BPF-based module. In any event, this is extremely useful for us as it is a free 100MB of memory on every machine, and eliminates the need to ship bcc and LLVM / dependencies to every machine too.

Help needed / known issues
- I haven't spent much (any) time looking at cross platform or cross OS and how to make sure the makefiles will work correctly. I'm not sure what's needed yet here.
- This requires a recent-ish version of clang / llvm toolchain in order to compile the BPF to ELF. GCC has partial but incomplete support for BPF CO-RE and to produce the correct ELF files. If there is a way to work with GCC too I think that would be good to include.
- The compilation assumes `-lelf` and `-lbpf`, I haven't recorded this in any makefiles
- C is not a first language for me, and I haven't yet done a pass to check what `free()` etc I might have missed.
- I have not yet tested this in the context of a DSO / htop environment but I think this would be valuable.

I will of course do a squash before continuing!


src/pmdas/bpf/bpf.c
```diff
@@ -0,0 +1,405 @@
...
+}
+
+/**
+ * load all known modules
+ */
+void
+bpf_load_modules()
+{
+    module_count = sizeof(all_modules)/sizeof(all_modules[0]);
+    modules = (module**) malloc(2 * sizeof(module*));
```
andreasgerstmayr, line 184: hardcoded `2` modules

src/pmdas/bpf/bpf.c
```diff
@@ -0,0 +1,405 @@
...
+static int
+bpf_fetchCallBack(pmdaMetric *mdesc, unsigned int inst, pmAtomValue *atom)
+{
+    unsigned int	cluster = pmID_cluster(mdesc->m_desc.pmid);
+    unsigned int	item = pmID_item(mdesc->m_desc.pmid);
+
+    // find module for this cluster, and issue fetch to module
+    // if somehow this bypasses PMCD PMNS filtering it is possible that
+    // a call to an uninitialised module occurs; the module is to handle this
+    for (int i = 0; i < module_count; i++)
```
andreasgerstmayr, line 72: could be a dict/hash, but the real-world performance gain is probably negligible as there won't be 1000s of modules

andreasgerstmayr (commented):
Looks very good, thanks for this PR Jason!

I've added a few comments inline (not all are equally important).

I'm trying to grasp the need for the `vmlinux.h` file - doesn't this file lock us in with a specific kernel version and architecture (exactly what we want to avoid with BPF CO-RE)? Could the requirement of this file be avoided when BTF is enabled?

src/pmdas/bpf/modules/GNUmakefile
```diff
@@ -0,0 +1,18 @@
...
+# needs to match libbpf v0.4+
+LIBBPF_HEADERS_PATH = /home/jkoch/code/libbpf/src
+
+MODULES = biolatency.so runqlat.so
+BPF_MODULES = biolatency.bpf.o runqlat.bpf.o 
+
+default: $(MODULES) $(BPF_MODULES)
+
+# Use the clang-11 pipeline to emit LLVM to LLD and emit BPF straight to an ELF .o.
+# clang-10 also works. The GCC pipeline has parts of this but not completely
```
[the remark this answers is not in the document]
jasonk000: Yes, I think it's worth mentioning that this PMDA would only be applicable to recent enough versions of Linux that support BTF output anyway, for which clang/llvm will be available.

src/pmdas/bpf/tail-logs.sh
```diff
@@ -0,0 +1 @@
+tail -F /var/log/pcp/pmcd/*
```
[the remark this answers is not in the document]
jasonk000: hah, oops!, thanks =)

src/pmdas/bpf/bpf.c
jasonk000: good catch

src/pmdas/bpf/modules/biolatency.c
```diff
@@ -0,0 +1,158 @@
...
+    int ret = bpf_map_lookup_elem(biolatency_fd, &key, &value);
+    if (ret == -1) {
+        return PMDA_FETCH_NOVALUES;
+    }
+
+    atom->ull = value;
+    return PMDA_FETCH_STATIC;
+}
+
+module* load_module()
```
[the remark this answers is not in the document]
jasonk000: Good suggestion. I was thinking primarily about flexibility & abillity for module to make dynamic decisions, but that is achievable in other ways too.

src/pmdas/bpf/modules/module.h
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * the pmid cluster assigned to each module, unique for each module.
+ */
+#define RUNQLAT_CLUSTER 0
+#define BIOLATENCY_CLUSTER 1
+
+/**
+ * List of all modules defined
+ *
+ * TODO make this dynamic
```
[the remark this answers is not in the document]
jasonk000: I'll remove this todo as I agree

src/pmdas/bpf/modules/module.h
```diff
@@ -0,0 +1,136 @@
...
+ */
+char *all_modules[] = {
+    "runqlat",
+    "biolatency"
+};
+
+/**
+ * List of modules that will be loaded and served as metrics
+ *
+ * TODO make this dynamic
```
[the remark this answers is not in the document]
jasonk000: oh this is dead code; the [module loading decision is already dynamic](https://github.com/performancecopilot/pcp/pull/1351/files/5665f1a12cb307fdb9d20f97e4a0ce4f6a82b492#diff-8c86d251fc99260decdf5947a95ed931c23ee5b8f446cc69a8079c5a486ea9d7R285)

src/pmdas/bpf/bpf.c
jasonk000: I initially had this mapped as a direct cluster id -> module mapping, which made it a linear lookup, but changed this when introducing dynamic loading. What I will do is add a post-configuration cluster id -> module setup. Thanks!

src/pmdas/bpf/bpf.c
jasonk000: fixed in d1fa2a27b

src/pmdas/bpf/modules/module.h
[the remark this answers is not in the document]
jasonk000: fixed in eced58ca3

src/pmdas/bpf/modules/module.h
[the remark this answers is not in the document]
jasonk000: fixed in eced58ca3

src/pmdas/bpf/modules/biolatency.c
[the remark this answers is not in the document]
jasonk000: fixed in 5b82207bb

src/pmdas/bpf/tail-logs.sh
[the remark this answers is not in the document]
jasonk000: fixed in bcd4994cd64

src/pmdas/bpf/bpf.c
jasonk000: fixed in 8d708ceb7

jasonk000:
> Looks very good, thanks for this PR Jason!
> 
> I've added a few comments inline (not all are equally important).
> 

Thank you. I've addressed all of them (pending your input) and will have a look at how to specify clang as a dependency for `configure`.

> I'm trying to grasp the need for the `vmlinux.h` file - doesn't this file lock us in with a specific kernel version and architecture (exactly what we want to avoid with BPF CO-RE)?

Good question. `vmlinux.h` provides the compilation process with the required structs/type information so that compile & link stage has the information to validate types and set up the relocation information in ELF file. It does _not_ need to match the running kernel version as libbpf performs the relocations at runtime to match the running kernel.

In the scenario that a new feature is added to the kernel that you want to trace with BPF CO-RE, if it has new struct/type information that is not available in your version of vmlinux.h, only then does it need to be regenerated.

This article ([BPF Portability and CO-RE](https://nakryiko.com/posts/bpf-portability-and-co-re/)) - especially the section "Compiler support" - provides an excellent explanation of this process.

> Could the requirement of this file be avoided when BTF is enabled?

There are three options
1. Embed a full vmlinux.h in the repo (aka current)
2. Generate vmlinux.h at build time (see below for build time requirements)
3. Provide a cutdown / custom header file with just the structs and fields we need (trivial, introduces work to development process)

To generate dynamically on the  build-server, requirements are:
- `CONFIG_DEBUG_INFO_BTF=y` must be compiled into kernel - most recent distributions will have this.
- `bpftool` installation (on ubuntu this is provided by `linux-tools-common`).
- `pahole >= 1.16` is required (on ubuntu this is provided by `dwarves`).
- Then we can add `bpftool btf dump file /sys/kernel/btf/vmlinux format c` to create it.

To provide only the structs we can ship a cut down vmlinux.h and provide _only_ the struct and fields that we need for our program. This requires manual maintenance if we need to access extra kernel-side structs; they need to be introduced manually. I don't think this is a big headache as we only need a few fields from `task_struct` currently; it is probably only 5-10 lines. Looking at the `BPF_CORE_READ` calls, we need only `struct task_struct { pid, tgid, state }`. This is a very viable solution too, it only makes it a module maintainer responsibility.

jasonk000:
For review of the cutdown `vmlinux.h` I have included a new file in 45d9cccae called `modules/vmlinux_cut.h`. This file is sufficient to replace vmlinux.h. It requires manual maintenance by copying the relevant definitions from a bpftool-generated vmlinux.h.

natoscott:
> These provide an alternative implementation to the pmdabcc python module.

Awesome.

>     * I haven't spent much (any) time looking at cross platform or cross OS and how to make sure the makefiles will work correctly. I'm not sure what's needed yet here.
> [...]
>     * The compilation assumes `-lelf` and `-lbpf`, I haven't recorded this in any makefiles

I've added configure macros for these (see commit c3fb7c371389021fe8e26e3301e1cf9db94300d3)

For clang, something like this in configure.ac will get you going...

```
AC_CHECK_PROGS(CLANG, clang)
AC_SUBST(CLANG)
```

and add matching changes in src/include/builddefs.in - then make use of $(CLANG) in the makefile.

It'd be really good to know if we need to bring in that entire toolchain or whether GCC is going to be able to do this for us anytime soon... but it sounds like we're going to need both.

natoscott:
> For review of the cutdown `vmlinux.h` I have included a new file in [45d9ccc](https://github.com/performancecopilot/pcp/commit/45d9cccae673721b9177c4aad6dac45653316deb) called `modules/vmlinux_cut.h`. This file is sufficient to replace vmlinux.h. It requires manual maintenance by copying the relevant definitions from a bpftool-generated vmlinux.h.

This looks to me like the way to go - fewer toolchain dependencies is ideal, and at least in its current form this stripped down header looks readily maintainable.

andreasgerstmayr:
Sorry, forgot to mention - I'm on PTO the next two weeks, meanwhile @natoscott will look over this PR.

natoscott:
@jasonk000 I've gone through the code in much more detail now, so can give some deeper feedback.

In general it's looking really good.  I recommend we remove the bpf.conf concept entirely (hear me out) and focus on two things:

- making module loading work 'automatically' - i.e. scan for modules in known locations (that only root can write to), load all .so files discovered there automatically on startup
- make writing the PCP BPF modules as simple as possible, and move all logic dealing with identifier assignment into the PMDA core code (src/pmdas/bpf/bpf.c) - this will help folks writing new modules, and make maintenance easier.

To achieve these goals, there's two PMDA concepts to embrace:

- dynamic metric names
- instance domain caches

To use "dynamic metric names", you'll need to change the 'pmns' file to look more like src/pmdas/mmv/src/pmns for example (i.e. use wildcards, no hard-coded names there).  Then, you need to add some new callbacks into the PMDA to handle the name <-> pmID lookups on behalf of pmcd(1).  The 3 new callbacks needed are 'name', 'pmid' and 'children'.  You might find the MMV PMDA code to be a useful reference there, but other C agents do similar things (sample/src/sample.c too, and a few other more complex ones like pmdalinux, pmdaproc).

In terms of "instance domain caches", I highly recommend switching to the pmdaCacheOp(3) interfaces instead of using the array-based instance domains currently being used.  These interfaces guarantee certain rules are satisfied for you and if used right, in addition to indoms for modules, we can also use these APIs to manage (dynamic, persistent) identifier assignment across the entire PMDA.  Read the rules outlined in the second paragraph of pmdaCacheOp(3) carefully.  Also note there's the concept of PMDA_CACHE_STRINGS in that man page - we can use that for persistent identifier management for things beyond indoms (pmID cluster identifiers, for example).

Using these two concepts (dynamic metric names, indom caches) we can completely get rid of all static identifier assignments.  Here's how it would work overall.

At PMDA startup, in the bpf_init() routine for the agent, reserve two 'special' persistent indom caches for strings.  The first one we'll use for pmID cluster numbers, and the second for pmInDom serial numbers - let's call them CLUSTERS and SERIALS one...

```
#define CLUSTERS 0
#define SERIAL 0
[...]
bpf_init()
{
    [...]
    pmdaCacheResize(CLUSTERS, (1<<12 - 1));  /* see libpcp.h __pmID_int */
    pmdaCacheOp(CLUSTERS, PMDA_CACHE_STRINGS);
    pmdaCacheOp(CLUSTERS, PMDA_CACHE_LOAD);
    pmdaCacheResize(SERIALS, (1<<22 - 1));  /* see libpcp.h __pmInDom_int */
    pmdaCacheOp(SERIALS, PMDA_CACHE_STRINGS);
    pmdaCacheOp(SERIALS, PMDA_CACHE_LOAD);
    bpf_load_modules();
```

in bpf_load_modules, we now iterate the directory ($PCP_PMDAS_DIR/bpf/modules perhaps) and pick out any file ending with .so - we dlopen that file (just like the code already does) and keep the rest of the filename (without .so suffix) as the main identifier for the module.  Perhaps we go so far as to mandate this is the module 'name' and drop that part of the current module interface - one less callback function for module writers to worry about.

Still inside bpf_load_module, using that 'name', for every module we dlopen call:

```
   int cluster = pmdaCacheStore(CLUSTERS, PMDA_CACHE_ADD, name, module);
   module->cluster = cluster;
```

At the end of loading all the modules, before returning from bpf_load_modules(), call

```
   pmdaCacheOp(CLUSTERS, PMDA_CACHE_SAVE);
```

In this way we have persistent identifiers for metric names - the same cluster number will be assigned on every iteration of the PMDA (see mention of 'rules' earlier as to why that's important - this affects a number of client tools).

Rinse, repeat this concept to assign instance domain identifiers for those modules that need them.  We probably need to keep the existing pmdabpf module API for a module to say how many indoms it has, 
e.g.

```
    int cluster = pmdaCacheStore(CLUSTERS, PMDA_CACHE_ADD, name, module);
    module->cluster = cluster;
    int count = module->get_indom_count();
    module->indoms = count;
    for (int i = 0; count; i++) {
        char indom[64];
        pmsprintf(indom, sizeof(indom), "%s/%d", name, i);
        int serial = pmdaCacheStore(SERIALS, PMDA_CACHE_ADD, indom, NULL);
        module->set_indom_serial(i, serial);
    }
```

Now we have persistent, dynamic instance domain numbers also, and those indom identifiers cannot conflict across individual modules.

I'll stop here to break the review discussion up.  Next step though is for the modules to tell the PMDA each of the metrics (names) they support, and to construct a 'global' pmdaNamespace with names from across all the modules with that information, used in the new namespace callbacks I mentioned.

We may also want to provide a helper library for modules (module_helpers.c suggests we're going to need this anyway) - which provides helper glue code between the PMDA and modules, again to make the modules easy to write.

jasonk000:
I've added dynamic cluster ID, dynamic indom, and dynamic namespace setup, and it all works ok.

I have the two tests cases working on f34 but have a makefile issue:

Build completes, test fails on first run:
```

Passed all 1 tests

Virtual environment setup done, took 19m.

Please run:

    sudo -u pcpqa -i ./check XXX

to run a QA test. PCP is already installed, from sources located in './pcp'.
Starting a shell in the new virtual environment...

[pcpbuild@003241a157dc ~]$ sudo -u pcpqa -i
[pcpqa@003241a157dc ~]$ ./1898
QA output created by 1898

=== bpf agent installation ===
Updating the Performance Metrics Name Space (PMNS) ...
Terminate PMDA if already installed ...
[...install files, make output...]
Updating the PMCD control file, and notifying PMCD ...
Check bpf metrics have appeared ... X metrics and X values

PMDA was not ready in 10 seconds, test failed
[pcpqa@003241a157dc ~]$ 
logout
```
(Checking the log shows it's an issue with the ELF file not matching what libbpf expects it to look like).

After that, I recompile and install the `bpf/modules` files
```
[pcpbuild@003241a157dc ~]$ cd pcp/pcp-5.3.2/src/pmdas/bpf/modules/
[pcpbuild@003241a157dc modules]$ make && make install
make: Nothing to be done for 'check'.
gcc -I../../../../src/include -I../../../../src/include/pcp -DPCP_VERSION=\"5.3.2\" -fPIC -fno-strict-aliasing -D_GNU_SOURCE  -Wall -O2 -g    -c -o biolatency.o biolatency.c
gcc -I../../../../src/include -I../../../../src/include/pcp -DPCP_VERSION=\"5.3.2\" -fPIC -fno-strict-aliasing -D_GNU_SOURCE  -Wall -O2 -g    -c -o module_helpers.o module_helpers.c
gcc -I../../../../src/include -I../../../../src/include/pcp -DPCP_VERSION=\"5.3.2\" -fPIC -fno-strict-aliasing -D_GNU_SOURCE  -Wall -O2 -g    -Wall -L../../../../src/libpcp/src -L../../../../src/libpcp_web/src -L../../../../src/libpcp_pmda/src -L../../../../src/libpcp_mmv/src   -lbpf -shared biolatency.o module_helpers.o -o biolatency.so
gcc -I../../../../src/include -I../../../../src/include/pcp -DPCP_VERSION=\"5.3.2\" -fPIC -fno-strict-aliasing -D_GNU_SOURCE  -Wall -O2 -g    -c -o runqlat.o runqlat.c
gcc -I../../../../src/include -I../../../../src/include/pcp -DPCP_VERSION=\"5.3.2\" -fPIC -fno-strict-aliasing -D_GNU_SOURCE  -Wall -O2 -g    -Wall -L../../../../src/libpcp/src -L../../../../src/libpcp_web/src -L../../../../src/libpcp_pmda/src -L../../../../src/libpcp_mmv/src   -lbpf -shared runqlat.o module_helpers.o -o runqlat.so
clang-12 -emit-llvm -O2 -Wno-unknown-attributes -Xclang -disable-llvm-passes -g -c biolatency.bpf.c -o - | opt -O2 -mtriple=bpf-pc-linux | llvm-dis | llc -march=bpf -filetype=obj -o biolatency.bpf.o
clang-12 -emit-llvm -O2 -Wno-unknown-attributes -Xclang -disable-llvm-passes -g -c runqlat.bpf.c -o - | opt -O2 -mtriple=bpf-pc-linux | llvm-dis | llc -march=bpf -filetype=obj -o runqlat.bpf.o
../../../../install-sh -o root -g root -m 755 -d /usr/libexec/pcp/pmdas/bpf/modules
../../../../install-sh -o root -g root -m 755 -d /var/lib/pcp/pmdas/bpf/modules
../../../../install-sh -o root -g root -m 644 -t /var/lib/pcp/pmdas/bpf/modules biolatency.so runqlat.so biolatency.bpf.o runqlat.bpf.o  /usr/libexec/pcp/pmdas/bpf/modules
rm biolatency.o runqlat.o module_helpers.o
```
And now, the tests pass:
```
[pcpbuild@003241a157dc modules]$ sudo -u pcpqa -i
[pcpqa@003241a157dc ~]$ ./1898
QA output created by 1898

=== bpf agent installation ===
Updating the Performance Metrics Name Space (PMNS) ...
Terminate PMDA if already installed ...
[...install files, make output...]
Updating the PMCD control file, and notifying PMCD ...
Check bpf metrics have appeared ... X metrics and X values

=== report metric values ===
OK
=== check metric labels ===

=== remove bpf agent ===
Culling the Performance Metrics Name Space ...
bpf ... done
Updating the PMCD control file, and notifying PMCD ...
[...removing files...]
Check bpf metrics have gone away ... OK
[pcpqa@003241a157dc ~]$ ./1895
QA output created by 1895

=== bpf agent installation ===
Updating the Performance Metrics Name Space (PMNS) ...
Terminate PMDA if already installed ...
[...install files, make output...]
Updating the PMCD control file, and notifying PMCD ...
Check bpf metrics have appeared ... X metrics and X values

=== report metric values ===
OK
=== check metric labels ===

=== remove bpf agent ===
Culling the Performance Metrics Name Space ...
bpf ... done
Updating the PMCD control file, and notifying PMCD ...
[...removing files...]
Check bpf metrics have gone away ... OK
```

@natoscott would love it if you have any pointers.

natoscott:
> I've added dynamic cluster ID, dynamic indom, and dynamic namespace setup, and it all works ok.
> 
> I have the two tests cases working on f34 but have a makefile issue:

It's lookin' pretty awesome.

> @natoscott would love it if you have any pointers.

Yeah - I'm able to reproduce the same.  It seems to be something 'special' that's happening as part of the rpm build process (I know there's binary/library rewriting that happens in the final stages of the rpmbuild(1) - but not something we have control over.  I haven't been able to pinpoint exactly which change by rpmbuild is causing this though.

I did notice we seem to be taking SIGSEGV though (pmcd.log shows this), and I believe that's the root cause of the lack of metrics showing up in the test.  I *think* its inside the call to bpf_object__name (in runqlat.c for example).  So my working theory is something that's being changed in the layout of the shared library is causing libbpf code to crash... but that's not 100% proven yet.

HTH.  Here's a few small, unrelated cleanups...

```
diff --git a/src/pmdas/bpf/.gitignore b/src/pmdas/bpf/.gitignore
new file mode 100644
index 0000000000..da89d1095c
--- /dev/null
+++ b/src/pmdas/bpf/.gitignore
@@ -0,0 +1,5 @@
+domain.h
+exports
+pmdabpf
+pmda_bpf.so
+modules/*.so
diff --git a/src/pmdas/bpf/Install b/src/pmdas/bpf/Install
index 4f9b256667..2379c16749 100755
--- a/src/pmdas/bpf/Install
+++ b/src/pmdas/bpf/Install
@@ -13,7 +13,7 @@
 # or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
 # for more details.
 # 
-# Install the simple PMDA and/or PMNS
+# Install the BPF PMDA and/or PMNS
 #
 
 . $PCP_DIR/etc/pcp.env
diff --git a/src/pmdas/bpf/Remove b/src/pmdas/bpf/Remove
index 9f7272b446..1768907104 100755
--- a/src/pmdas/bpf/Remove
+++ b/src/pmdas/bpf/Remove
@@ -12,7 +12,7 @@
 # or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
 # for more details.
 # 
-# Remove the simple PMDA
+# Remove the BPF PMDA
 #
 
 . $PCP_DIR/etc/pcp.env
diff --git a/src/pmdas/bpf/modules/GNUmakefile b/src/pmdas/bpf/modules/GNUmakefile
index 771cffa74b..250061cbed 100644
--- a/src/pmdas/bpf/modules/GNUmakefile
+++ b/src/pmdas/bpf/modules/GNUmakefile
@@ -10,8 +10,6 @@ MODULEDIR = $(PCP_PMDASADM_DIR)/$(IAM)/modules
 $(eval $(shell grep ^VERSION_ID= /etc/os-release))
 $(eval $(shell grep ^ID= /etc/os-release))
 
-include $(BUILDRULES)
-
 # Use the clang pipeline to emit LLVM to LLD and emit BPF straight to an ELF .o.
 # The GCC pipeline has parts of this but not completely, and in any case, likely
 # requires version/feature detection too.
@@ -47,5 +45,7 @@ install_pcp install:	default
 	$(INSTALL) -m 755 -d $(MODULETMP)
 	$(INSTALL) -m 644 -t $(MODULETMP) $(MODULES) $(BPF_MODULES) $(MODULEDIR)
 
+include $(BUILDRULES)
+
 clean::
 	rm -f $(MODULES) $(BPF_MODULES)

```

jasonk000:
Thanks for the pointer. It looks like `rpmbuild` in fact calls `strip` against the binaries, and as it does this, it breaks the relocation information included in the eBPF ELF file. This looks OK but is no longer usable by `libbpf`.

Based on this, there are two options:
- Introduce a custom strip approach that filters out the relevant bpf modules on the list. I've introduced a basic wrapper here that works to prove the point.
- Disable symbol stripping altogether. The list of files is actually quite short and I doubt will have any practical negative impact on the produced packages to not strip. By my count there are only 7 files affected: `libpcp_pmda.a`, `libpcp_trace.a`, `libpcp_gui.a`, `libpcp.a`, `libpcp_web.a`, `libpcp_import.a`, `libpcp_mmv.a`

build/rpm/custom-strip
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+echo "STRIPPING IN $*"
+
```
natoscott, line 4: This script could do with a little comment describing why it exists.

src/pmdas/bpf/README
```diff
@@ -0,0 +1,40 @@
...
+===========
+
+To develop additional modules:
+- Follow the examples provided (runqlat and biolatency) to create a new module (which will output a .so). The entry
+  point in your .so will be a load_module() call that should return a newly allocated `struct module` object.
+- Create your bpf code (this will become a .bpf.o). Use the various `_helpers` headers.
+- Ensure `module.h` has correct unique setup for your cluster and metric ids.
+- Add details to `pmns` and `help` files to ensure they match the `module.h` changes.
+
+TODO
```
natoscott, line 36: TODO section is out-of-date now.

src/pmdas/bpf/bpf.conf
```diff
@@ -0,0 +1,3 @@
+# list the modules to start here
```
natoscott, line 1: We gonna stick with this config file or make the agent completely dynamic?  (this modules list can be built via a directory scan - easier install / maintenance)

If module configuration becomes necessary for some modules (in future) module-specific configs might be simpler.

src/pmdas/bpf/modules/GNUmakefile
```diff
@@ -0,0 +1,54 @@
...
+# Some distributions place commands at different names (llvm-dis vs llvm-dis-12),
+# so detect distro directly.
+#
+# clang-12 special notes: https://www.spinics.net/lists/netdev/msg689901.html
+#     nb: opt and llvm-dis appear as opt-12 and llvm-dis-12 on some OSs
+
+.PHONY: .FORCE
+FORCE:
+
+ifeq "$(ID)$(VERSION_ID)" "fedora34"
```
natoscott, line 26: Instead of these distribution-specific checks (maintenance nightmare) how about checking for the different clang versions in configure.ac?  (and passing the correct flags for each down from that top level)

src/pmdas/bpf/modules/biolatency.c
```diff
@@ -0,0 +1,163 @@
...
+struct module bpf_module = {
+    .init               = biolatency_init,
+    .register_metrics   = biolatency_register,
+    .metric_count       = biolatency_metric_count,
+    .indom_count        = biolatency_indom_count,
+    .set_indom_serial   = biolatency_set_indom_serial,
+    .shutdown           = biolatency_shutdown,
+    .refresh            = biolatency_refresh,
+    .fetch_to_atom      = biolatency_fetch_to_atom,
+	.metric_name        = biolatency_metric_name,
```
natoscott, line 162: Extra whitespace.

src/pmdas/bpf/modules/module.h
```diff
@@ -0,0 +1,108 @@
...
+     * Initialise a module.
+     *
+     * @return 0 if no issues, non-zero to indicate errors.
+     */
+	init_fn_t init;
+
+    /**
+     * Release any resources associated with the module.
+     *
+     * TODO Currently never called
```
natoscott, line 74: Can this be added?  (else lets drop the callback before merge)

src/pmdas/bpf/modules/module_helpers.c
```diff
@@ -0,0 +1,25 @@
+#include "module_helpers.h"
```
natoscott, line 1: Since there's the only one small function here, we could simplify the build by making fill_instids_log2 a static inline in module.h and getting rid of module_helpers.[ch] entirely.

This would make things slightly easier for folks building modules.

natoscott (changes requested):

src/pmdas/bpf/bpf.conf
andreasgerstmayr: +1

I'd go with `mods-enabled` and `mods-available` directories, where users can move (or symlink) scripts from the `mods-available` directory to the `mods-enabled` directory. The `mods-enabled` directory and all files inside must be exclusively writeable by root, and all modules inside will be loaded on startup.

Regarding module configuration: Users could write sections like
```
[tracing.so]
process_name = firefox
```
in `bpf.conf`, to overwrite any default module configuration values.

I'm not sure if we need the flexibility of running two instances of the same module (with different configs) - for that we'd need another solution (the bcc PMDA can do that for example).

andreasgerstmayr:
I tested the PMDA today on my Fedora 34 notebook:

I couldn't get it working with libbpf 0.3.0 (getting error -22 while loading /sys/kernel/btf/vmlinux, no idea why), so I downloaded libbpf 0.4.0 (https://bodhi.fedoraproject.org/updates/FEDORA-2021-990e9573f6). @natoscott, did it work for you with libbpf 0.3 from the Fedora repos?

Now biolatency works, but runqlat doesn't show any values.
Afaics that's because the symbol/kprobe got renamed in GCC 10.2, the Python based tools fixed it in https://github.com/iovisor/bcc/pull/3315 by using a regex when attaching to that probe. The libbpf tool uses `SEC("tp_btf/sched_switch")` (https://lwn.net/Articles/838168/).

Things like this will keep happening (same for our current bcc-based tools), not sure if there is a better solution other than cluttering the code with many conditionals to check for different versions etc. (well, switching to tracepoints solves it, but in many cases there is no tracepoint available)? - Anyway, that's a thing for the maintainers to worry about, just wanted to bring it up.

Some minor things:
* help texts don't work
* afaics we can drop `vmlinux.h` now, as we have `vmlinux_cut.h` (and let's put the instructions how to generate it in the `README.md`)

I think the PR is in a good shape and ready for merging after addressing Nathan's comments and the two issues above.

build/rpm/custom-strip
```diff
@@ -0,0 +1,21 @@
...
+skip=(
+  "${RPM_BUILD_ROOT}/usr/libexec/pcp/pmdas/bpf/modules/runqlat.bpf.o"
```
andreasgerstmayr, line 6: let's put a wildcard there instead of hard coding the module names, otherwise it will bit us for future modules

andreasgerstmayr (changes requested):

src/pmdas/bpf/README
jasonk000: fixed in 7b04b5b0a

src/pmdas/bpf/bpf.conf
jasonk000: Agree generally a configuration approach is needed.

I'm not sure I understand the benefit to the approach of enumerating directory to load, feels like we are splitting configuration across two places (module directory and `bpf.conf`. How do you see the benefit vs listing the modules to load directly in the config file as bcc does now .. maybe something learned in bcc development that I'm missing?

jasonk000: > We gonna stick with this config file or make the agent completely dynamic? (this modules list can be built via a directory scan - easier install / maintenance)
> If module configuration becomes necessary for some modules (in future) module-specific configs might be simpler.

Yeah, my initial thinking was let's start with a basic configuration file parser. I'm not clear on the benefit of enumerating directories over listing in the file, as a user likely has to configure the file anyway?

One alternative would be the module is responsible for its own configuration somehow.

src/pmdas/bpf/modules/GNUmakefile
jasonk000: Agreed, I'm a bit more comfortable with the different OS environments so I think I can make this work now.

build/rpm/custom-strip
jasonk000: fixed in cb8b03215

build/rpm/custom-strip
jasonk000: fixed in cb8b03215

src/pmdas/bpf/modules/module_helpers.c
jasonk000: Sure. I've done this in a06b3a403.

build/rpm/custom-strip
jasonk000: fixed in cb8b03215b1be9fd45dfde55ba6720a2d3969e9f

natoscott:
> @natoscott, did it work for you with libbpf 0.3 from the Fedora repos?

No, I've had to use a libbpf git build.

> I think the PR is in a good shape and ready for merging after addressing Nathan's comments and the two issues above.
> And the memory footprint is impressive, using just 7 MB RSS +1

+1

src/pmdas/bpf/modules/GNUmakefile
jasonk000: fixed in f6d963a5c

jasonk000:
OK!
- `iniparser` included to parse ini files
- configuration set up to allow override of inbuilt clang and llvm-strip path, and pass configuration settings to module
- found a way to avoid customisation of BPF compilation based on version
- set up help text
- use of iovisor/bcc GPL codebase

Ideally we could actually auto-import all iovisor/bcc/libbpf-tools tools?

src/pmdas/bpf/bpf.conf
jasonk000: I've introduced a configuration file and use of `iniparser`. Nothing _uses_ it yet, but the capability is there and a module can pick it up and read it.

andreasgerstmayr:
> OK!
> 
>     * `iniparser` included to parse ini files
> 

We have a vendored ini library (`src/external/ini.c`), and wrappers around it (`src/libpcp_web/src/config.c`), could you use this one instead of the new `iniparser` dependency?
(Please don't shoot the messenger :smiley:  )

>     * configuration set up to allow override of inbuilt clang and llvm-strip path, and pass configuration settings to module
> 
>     * found a way to avoid customisation of BPF compilation based on version
> 
>     * set up help text
> 
>     * use of iovisor/bcc GPL codebase

Awesome!

> 
> 
> Ideally we could actually auto-import all iovisor/bcc/libbpf-tools tools?

Sounds great! We can import all `.bpf.c` files, but we still need to translate the command-line args to config files, and put everything in the new `module` struct, right?

jasonk000:
> We have a vendored ini library (src/external/ini.c), and wrappers around it (src/libpcp_web/src/config.c), could you use this one instead of the new iniparser dependency?

I didn't realise it was included, I looked in src/vendor only. I picked iniparser as it is lightweight and available everywhere we are going to build bpf.

Regardless, I've made the swap in 35c04e9. In my view, inih is less clean as it requires extracting pieces of libpcp_web. I'm not sure we want to refactor libpcp_web as part of this PR, but that could be a logical step.

> Sounds great! We can import all .bpf.c files, but we still need to translate the command-line args to config files, and put everything in the new module struct, right?

Yes exactly, I've only ported the minimum to get the BPF framework running. There will need to be subsequent PR for any new scripts to import:
- map each of them to configuration values
- set up the pmns hierarchy for the module
- work out how to model the outputs in PM data format (eg: how to handle string-like outputs)

It's a little bit of work for each one, but across all of them at once would be quite a lot. I'd suggest they come in as separate changes rather than trying to bring the whole lot in a single pass.

andreasgerstmayr (approved):

andreasgerstmayr:
> In my view, inih is less clean as it requires extracting pieces of libpcp_web. I'm not sure we want to refactor libpcp_web as part of this PR, but that could be a logical step.

+1, I'll take care of that


> 
> > Sounds great! We can import all .bpf.c files, but we still need to translate the command-line args to config files, and put everything in the new module struct, right?
> 
> Yes exactly, I've only ported the minimum to get the BPF framework running. There will need to be subsequent PR for any new scripts to import:
> 
>     * map each of them to configuration values
> 
>     * set up the pmns hierarchy for the module
> 
>     * work out how to model the outputs in PM data format (eg: how to handle string-like outputs)
> 
> 
> It's a little bit of work for each one, but across all of them at once would be quite a lot. I'd suggest they come in as separate changes rather than trying to bring the whole lot in a single pass.

+1, can be done one module per PR

Thanks for your contributions Jason, this new PMDA is really awesome!

PR #1400, opened 2021-09-06
selinux: Use interface call instead of a simple rule

vmojzis:
The rule does not cover access to directories along the way the log
file. Also, interface already contains a list of all the required types.

This change is needed for the policy to work on systems where glusterd
module was removed.

Signed-off-by: Vit Mojzis <vmojzis@redhat.com>


vmojzis:
The whole policy should be rewritten, please get in touch if that is something you'd be willing to work on.

vmojzis:
It seems you already "fixed" the issue by removing the policy rule, but for future it would be great if you could at least put the whole policy into an "optional_policy" block (same as I did with the new macro). Or even better, assign each rule/group of rules it's own optional_policy block containing the appropriate requirements (using macros would simplify this quite a bit -- you can search for appropriate macros using "audit2allow -R"). That way whenever some dependency is missing, the policy block using it will be automatically disabled (as opposed to the policy breaking).

andreasgerstmayr:
> That way whenever some dependency is missing, the policy block using it will be automatically disabled (as opposed to the policy breaking).

That sounds great!

We'd like to tackle this, and also move the PCP policy out of the selinux-policy package (i.e. https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/contrib/pcp.te) into pcp-selinux (see Red Hat BZ https://bugzilla.redhat.com/show_bug.cgi?id=1941873) in a future release. We'll get back to you once we're at it, afaics we will definitely need some help with this change. Thanks!

portante (approved):
FWIW, verified this change by confirming the interface definition from https://github.com/fedora-selinux/selinux-policy/blob/7e50553feb19abeab49911db46c15b50b6bda47e/policy/modules/contrib/glusterd.if#L147.

portante:
And I have also built and installed this PR on a RHEL 8.5 system and the installed module works.

natoscott:
This was part of the series from @portante recently merged.

Issue #1305, opened 2021-05-26
encoding of special characters in hostspecs


andreasgerstmayr:
How are special characters processed in hostspecs?

For authentication with pmproxy, pmproxy uses the following hostspec to connect to a local pmcd: `pcp://127.0.0.1?username=<username>&password=<password>`. With a password like `a=b` the following hostspec is generated: `pcp://127.0.0.1?username=metrics&password=a=b`, which fails when connecting to pmcd (the authentication mechanism finds an empty password, and is prompting for the password in the terminal).

Afaics the hostspec is read in: https://github.com/performancecopilot/pcp/blob/b877590f82cd1ef8113aaf288bafb2b0c31835d8/src/libpcp/src/spec.c#L830-L893 but I can't see any unescaping of special characters like `&`, `=`, `?` (which can be part of the parameter values).

If the hostspec should resemble a valid URI, we need a function like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent to encode query values and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent to decode query values.

Afaics `http_decode`/`http_url_decode` from `src/pmproxy/src/http.c` performs this (de)coding and parsing of URIs already.

(This is the root issue of https://bugzilla.redhat.com/show_bug.cgi?id=1964255)

PR #1413, opened 2021-09-17
libpcp: use urlencoding for hostspec parameter values

andreasgerstmayr:
Resolves: #1305

The new urldecoding function in `decode_uri_component` in libpcp/strings.c is effectively the same as `http_decode` in pmproxy/http.c, except using `char*` instead of `sds` (because `sds` is not available in libpcp).

To avoid having two functions doing the same, I'd suggest using the urldecoding function of libpcp also for pmproxy. Should I export this function in libpcp/exports? It's a simple, standalone, utility function (and currently has no `__pm` prefix). What do you suggest @natoscott?


natoscott:
> [...]
> To avoid having two functions doing the same, I'd suggest using the urldecoding function of libpcp also for pmproxy. Should I export this function in libpcp/exports? It's a simple, standalone, utility function (and currently has no `__pm` prefix). What do you suggest @natoscott?

Yep, I agree having a single shared implementation in libpcp would be a good thing.  The naming would be best if tweaked to match libpcp conventions - something simple like pmstrencode and pmstrdecode I guess.  Or even simpler, without the the 'str' perhaps?  Any of those is fine IMO.  If its necessary to explain the intended use (C string URI component encoding/decoding), a brief man page might be best rather than going for long, segmented function names (like pmencode_uri_component and pmdecode_uri_component).

In sds.c the encoding routine uses a cryptic 'repr' name (sdscatrepr) - IMO that's taking short names too far :) but there's a happy medium to be found.

andreasgerstmayr:
> > [...]
> > To avoid having two functions doing the same, I'd suggest using the urldecoding function of libpcp also for pmproxy. Should I export this function in libpcp/exports? It's a simple, standalone, utility function (and currently has no `__pm` prefix). What do you suggest @natoscott?
> 
> Yep, I agree having a single shared implementation in libpcp would be a good thing. The naming would be best if tweaked to match libpcp conventions - something simple like pmstrencode and pmstrdecode I guess. Or even simpler, without the the 'str' perhaps? Any of those is fine IMO.

Ok, I've renamed them to `__pmStrEncode` / `__pmStrDecode`.

> If its necessary to explain the intended use (C string URI component encoding/decoding), a brief man page might be best rather than going for long, segmented function names (like pmencode_uri_component and pmdecode_uri_component).

Regarding documentation: What about annotating the function, and generating the man page with doxygen (or alternatives)? In many editors/IDEs you can hover over function names (or struct members) and it shows the comment/documentation, without leaving the editor (that would be a separate discussion/issue, just wanted to mention it).

> In sds.c the encoding routine uses a cryptic 'repr' name (sdscatrepr) - IMO that's taking short names too far :) but there's a happy medium to be found.

I guess that's related to the `repr()` function in Python:
```
>>> print(repr("\t\r\n"))
'\t\r\n'
```
It comes handy when you have a string with special/non-printable characters like carriage returns or tabs and want to see them, afaics same what sdscatrepr is doing.

natoscott (approved):
LGTM.

natoscott:
> > > [...]
> Regarding documentation: What about annotating the function, and generating the man page with doxygen (or alternatives)?

For little helper functions like this, I have no problem with that (and maybe no man page at all is fine here too - its such a small, specialised pair of routines).  For most of the PMAPI though, this isn't going to work - there's too much content, formatting, and other man page specific details for the pages to become generated without substantial lost information (and quality).

> I guess that's related to the `repr()` function in Python:

Ah, interesting - I didn't know that!  Thanks.

andreasgerstmayr:
> > > > Regarding documentation: What about annotating the function, and generating the man page with doxygen (or alternatives)?
> 
> For little helper functions like this, I have no problem with that (and maybe no man page at all is fine here too - its such a small, specialised pair of routines). For most of the PMAPI though, this isn't going to work - there's too much content, formatting, and other man page specific details for the pages to become generated without substantial lost information (and quality).

I agree, it'll be difficult to get the same level of formatting in code comments :|


I noticed an issue with some username/password combinations - turns out `base64_decode` returns a `sds` string where `sdslen(x)` != `strlen(x)` when the base64 string is padded (e.g. `SQ==`). I've fixed this case now and added cmocka tests for it.

While at it, I though the name `pmStrEncode` is not specific enough - is it base64 encoding? url encoding, or some other encoding? So I went with `pmUrlEncode` finally. I'll merge it after CI if there are no objections.

natoscott:
> While at it, I though the name `pmStrEncode` is not specific enough - is it base64 encoding? url encoding, or some other encoding? So I went with `pmUrlEncode` finally. I'll merge it after CI if there are no objections.

No objections here - thanks, great work!

Issue #896, opened 2020-04-23
pmlogctl/pmiectl for managing pmlogger and pmie instances and pmfind for discovering them


goodwinos:
On platforms supporting systemd, use a template service unit configuration for managing multiple pmlogger instances. Similarly for pmie.

goodwinos:
**Design notes for templated service units for pmlogger.** Similar (though simpler) concepts apply for pmie services.

**Service units**
We keep the pmlogger.service config as-is (for the primary logger) and add a new config named "pmlogger@.service". The latter is templated (see below) and shared by all pmloggers (if any) except the primary.

For the primary logger, there are pretty much no changes. The rc scripting for start-systemd will be called without the (new) optional host argument. The primary logger config in the control file will still be used. Same pidfile etc. On non-systemd platforms there is no change; the pmlogger service manages all pmloggers configured in the control files (primary and all remote loggers, if any) and the control files continue to be manually edited / configured.

For remote pmlogger instances, the pmlogger@HOST service manages the pmlogger for each HOST. In the template, %i substitutes HOST. Each templated service instance is individually managed as a systemd service.

```
    [Unit]
    Description=Performance Metrics Archive Remote Logger for %i
    Documentation=man:pmlogger(1)
    After=network-online.target

    [Service]
    Type=notify
    NotifyAccess=all
    TimeoutSec=120
    Restart=always
    ExecStart=@PCP_RC_DIR@/pmlogger start-systemd %i
    ExecStop=@PCP_RC_DIR@/pmlogger stop-systemd %i
    PIDFile=@PCP_RUN_DIR@/pmlogger-%i.pid

    [Install]
    WantedBy=multi-user.target
```

**Advantages of service unit templating include:**

- individual pmloggers can be managed (started / stopped etc), rather than the entire farm. Regexs can be used to manage multiple loggers by matching on the first field in each control file entry (i.e. hostname).

- all pmloggers can start in parallel rather than serially (=> better boot times) and each will notify systemd when they're started and ready (or stopping)

- new pmloggers can be configured and enabled dynamically without manually editing the control file, e.g. systemctl enable pmlogger@host1 pmlogger@host2 ...

A disadvantage is that "systemctl start pmlogger" only starts the primary logger, not the whole farm. A systemd 'target' config such as "pmlogger-farm.target" could be used to start or stop all enabled pmloggers if we want. This is how Redis v6 handles management of all Redis instances. Or we can use regex, e.g. systemctl start pmlogger@*

[...]

PR #1428, opened 2021-10-01
Add pmlogger_check --only-primary option flag

goodwinos:
Logical opposite of pmlogger_check --skip-primary. Will be used (in subsequent commits) by pmlogger.service to start only the primary logger. Man page for pmlogger_check(1) is also updated.

### DIFF
--- a/man/man1/pmlogger_check.1
+++ b/man/man1/pmlogger_check.1
@@ -19,7 +19,7 @@
 \f3pmlogger_daily\f1 \- administration of Performance Co-Pilot archive log files
 .SH SYNOPSIS
 .B $PCP_BINADM_DIR/pmlogger_check
-[\f3\-CNpqsTV?\f1]
+[\f3\-CNPpqsTV?\f1]
 [\f3\-c\f1 \f2control\f1]
 [\f3\-l\f1 \f2logfile\f1]
 .br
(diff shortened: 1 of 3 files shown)


kmcdonell:
Hi Mark ... I was also expecting a change to pmlogger.service along the lines
```
Environment="PMLOGGER_CHECK_PARAMS=-C --only-primary"
```
to match what is done for pmlogger_check.service and then some additional glue to get this from the systemd service definition down thru the pmlogger RC script and onto pmlogger_check. 
Without this, I think the status quo would be unchanged.

goodwinos:
> Hi Mark ... I was also expecting a change to pmlogger.service along the lines
> 
> ```
> Environment="PMLOGGER_CHECK_PARAMS=-C --only-primary"
> ```
> 
> to match what is done for pmlogger_check.service and then some additional glue to get this from the systemd service definition down thru the pmlogger RC script and onto pmlogger_check. Without this, I think the status quo would be unchanged.

yes - I hadn't done that bit yet - wanted to discuss the hooks. I'll add those changes

thanks

goodwinos:
have been testing on the lab farm, seems to be mostly working but some of the pmloggers are not staying up for some reason so I'l have to continue with this in the morning

kmcdonell:
Hi Mark, this looks good and I've merged it into my tree and kicked the tyres a bit locally.
Only question I have is this bit in the pmlogger.service.in file:
```
# no start timeout - for logger farms
TimeoutStartSec=infinity
```
Specifically, I believe the logger farm management is delegated to the pmlogger_check service, so I don't follow this comment and the setting. 
However this is benign enough and we can revisit it next week.
So in the interim I'm going to approve/merge this PR as I have other stuff in the same general area that I'd like to avoid merge conflicts over.

goodwinos:
Hi Ken, agree that timeout shouldn't be needed any more (it was originally there to stop systemd killing farm nodes after the timeout). But now the primary under pmlogger.service and the farm (if any) under pmlogger_check.service are separated, the killing has stopped.

I have the extra glue needed to activate this - was testing over the week-end on a high latency cluster and also on a low latency, but didn't get to merge it (too busy with other stuff). Seems to be fine now - to start the primary and the farm immediately we need systemctl start pmlogger pmlogger_check. Or just start pmlogger and then pmlogger_check timer will trigger to start the farm in <= 30 mins. So will I push the remaining glue changes straight to main or will this upset your merge (not sure what other changes you have pending - is that parser latency improvements?)

Once this settles down, I'd like to consider making every farm node it's own service (using templates - as an evolved version of https://github.com/performancecopilot/pcp/issues/896#issuecomment-627698970).

thanks

PR #1447, opened 2021-10-15
pmdabpf valgrind tests

andreasgerstmayr:
depends on #1445

### DIFF
--- a/qa/1900
+++ b/qa/1900
@@ -1,0 +1,48 @@
+#!/bin/sh
+# PCP QA Test No. 1900
+# Test bpf PMDA with valgrind.
+#
+# Copyright (c) 2022 Red Hat.
+#
+
+seq=`basename $0`
+echo "QA output created by $seq"
+
+# get standard environment, filters and checks
+. ./common.product
+. ./common.check
+. ./common.bpf
+
+_check_valgrind
+_pmdabpf_check
+valgrind --version | sed -e 's/valgrind-//' | awk -F. '$1 < 3 || ($1 == 3 && $2 < 19) {exit 1}' || \
+    _notrun "this test requires valgrind 3.19+"
+
+status=1
+$sudo rm -rf $tmp.* $seq.full
+trap "cd $here; $sudo rm -rf $tmp.*; exit \$status" 0 1 2 3 15
+
+_filter_logs()
+{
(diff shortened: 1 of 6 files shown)


andreasgerstmayr:
blocked on https://bugs.kde.org/show_bug.cgi?id=451626

There's one other issue: The `pminfo -L -K clear -K add,157,$PCP_PMDAS_DIR/bpf/pmda_bpf,bpf_init` magic runs `bpf_init()` and then immediately fetches the metric values, right? Sometimes the biolatency module (which collects the latency of block I/O requests) didn't gather any requests in that short timeframe. Is there any way to run some command (`dd` to get some I/O activity) or wait a bit before fetching the values?

natoscott:
> There's one other issue: The `pminfo -L -K clear -K add,157,$PCP_PMDAS_DIR/bpf/pmda_bpf,bpf_init` magic runs `bpf_init()` and then immediately fetches the metric values, right?

Yep.  pmNewContext(3) does the bpf_init from memory, and pmFetch(3) does the sampling - but all within pminfo.

> Sometimes the biolatency module (which collects the latency of block I/O requests) didn't gather any requests in that short timeframe. Is there any way to run some command (`dd` to get some I/O activity) or wait a bit before fetching the values?

Not using pminfo - I think you'd need a custom tool that uses local context with a delay.  Or maybe use pmval?  It will still see the uninitialized metrics on first fetch, but subsequent samples should come good.  Doesn't do all of what pminfo does though, so maybe a little qa/src utility is needed, not sure.

| Ideally still keep exercising the DSO part of the bpf PMDA under valgrind.

+1

andreasgerstmayr:
> Or maybe use pmval? It will still see the uninitialized metrics on first fetch, but subsequent samples should come good. Doesn't do all of what pminfo does though, so maybe a little qa/src utility is needed, not sure.

Ah, great idea! I didn't realize pmval also supports `-K`. Yep, that works. Perfect :)

andreasgerstmayr:
I've created `qa/src/wait_for_values.c` now, because it's too messy to filter pmval output which is interweaved with valgrind output.
This tool waits up to X milliseconds to fetch any non-zero value, and sleeps 100ms if no value was found.

PR #1449, opened 2021-10-18
build: enable bpf PMDA in Debian packaging

andreasgerstmayr:
Does Debian automatically add dependencies on the libraries used for building? E.g. at runtime we need `libbpf0` and `libelf1`, but in the `control.pcp` and `fixcontrol` I only see the development dependencies (`lib*-dev`).

Another problem is the `linux-tools-$(uname -r)` package, which contains the current running kernel version in the name. This package contains `bpftool`, which is required for building the bpf PMDA, but not required at runtime.


andreasgerstmayr:
I was curious how Debian is packaging the libbpf-tools, and they're including "bpftool" as a package name in the build dependencies:
https://salsa.debian.org/debian/bpfcc/-/blob/c38a724df294a1b86a2c6cff315d0e351f7ea3c6/debian/control#L34

jasonk000:
With regard to libbpf0 and libelf1, libbpf0 depends directly on libelf1. libbpf deps has been listed in `qa/`, although it's probable I have misunderstood the build and packaging process when putting it together.

andreasgerstmayr:
> With regard to libbpf0 and libelf1, libbpf0 depends directly on libelf1.

Good to know, I've dropped the explicit dependency on libelf now.

> libbpf deps has been listed in qa/, although it's probable I have misunderstood the build and packaging process when putting it together.

The configure script checks which dependencies are installed, and enables (compiles) PMDAs if the requirements are met.
To set the dependencies in the package metadata, we have `debian/control.pcp` and `debian/fixcontrol`, where the `debian/fixcontrol` runs `sed` on the `control.pcp` and updates it based on the installed software on the build system. The files you updated in qa/ are lists of packages which should be installed at the QA systems before running the configure script, in order to enable/include all optional PCP components (like the bpf PMDA etc.).

andreasgerstmayr:
Debian has a standalone `bpftool` package now: https://packages.debian.org/bullseye/bpftool (afaics this package didn't exist when creating this PR, as the `bpftool` program was included in the `linux-tools` package previously).

However Ubuntu doesn't have `bpftool` as a standalone package (yet?): https://packages.ubuntu.com/search?suite=jammy&arch=any&searchon=contents&keywords=bpftool

So the current PR would leave Ubuntu users with impossible build requirements if the bpf PMDA gets enabled in the configure script. I assume Ubuntu will follow Debian and have it as a standalone package soon.. I guess? So maybe let's just wait.

andreasgerstmayr:
related to bpftool issue #1672, resolved by #1673

andreasgerstmayr:
With the vendored bpftool in #1673 this PR is ready to be merged now.

andreasgerstmayr:
@natoscott does this PR look good to you? Now we only need clang, llvm and libbpf-dev, because we vendor bpftool now.

natoscott:
@andreasgerstmayr LGTM, thanks!

natoscott (approved):

PR #1460, opened 2021-10-29
services: disable service advertising by default in pmcd and pmproxy

goodwinos:
Tweak pmcd.options and pmproxy.options to set -A by default. The -A
flag disables service advertising, which is otherwise enabled by
default. With hindsight, -A should have had the opposite semantics,
i.e. -A should have enabled service advertising .. but we can't
change that retrospectfully. Since pmcd.options and pmproxy.options
are config files, the existing settings (if altered from default)
will be retained on upgrade.

Also tweak systemd unit configs to remove Wants=avahi-daemon.service.
This was causing pmcd's systemd deps to trigger Insights warnings,
see BZ1899625 - "Enabling pmcd leads to Insights warning about avahi".

We retain the After=avahi-daemon.service setting in pmcd and add it
to pmproxy service units to ensure avahi starts before the PCP daemons.
This is not a hard dependency, it just specifies the service start
order.

Resolves: RHBZ#1899625

### DIFF
--- a/src/pmcd/pmcd.options
+++ b/src/pmcd/pmcd.options
@@ -19,6 +19,9 @@
 # run in the foreground (not as a daemon)
 # -f
 
+# disable service advertising (comment out to enable)
+-A
+
 # maximum incoming PDU size (default 64KB)
 # -L 16384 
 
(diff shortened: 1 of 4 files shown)


natoscott (approved):
LGTM

PR #1464, opened 2021-11-03
pmproxy: transparent Redis reconnect

andreasgerstmayr:
* switch from a boolean 'setup' to a state (connecting, connected,
  ready, ...) for Redis connection state
* handle Redis disconnects in the async reply handler
* in a timer, check if the current state is disconnected and issue a
  reconnect

### DIFF
--- a/qa/src/redis_proxy.python
+++ b/qa/src/redis_proxy.python
@@ -1,0 +1,73 @@
+#!/usr/bin/env pmpython
+#
+# simple TCP proxy, based on https://stackoverflow.com/a/46422554 (CC BY-SA 3.0)
+#
+
+import time
+import argparse
+import asyncio
+
+
+async def pipe(reader, writer, args, local_writer):
+    is_local_reader = not writer == local_writer
+
+    try:
+        while not reader.at_eof():
+            data = await reader.read(2048)
+            if not data:
+                break
+
+            print(">" if is_local_reader else "<", data)
+
+            # fake loading error (if enabled)
+            if (
+                is_local_reader
+                and args.loading_delay
+                and time.time() < args.loading_delay
+            ):
+                # https://github.com/redis/redis/blob/6.2.6/src/server.c#L2508-L2509
+                data = b"-LOADING Redis is loading the dataset in memory\r\n"
+                print("<", data)
+                local_writer.write(data)
+            else:
+                writer.write(data)
+    except Exception:
+        pass
+    finally:
+        writer.close()
+
+
+async def handle_client(local_reader, local_writer, args):
+    remote_host, remote_port = args.remote.split(":")
+    try:
+        remote_reader, remote_writer = await asyncio.open_connection(
+            remote_host, int(remote_port)
+        )
+        pipe1 = pipe(local_reader, remote_writer, args, local_writer)
+        pipe2 = pipe(remote_reader, local_writer, args, local_writer)
+        await asyncio.gather(pipe1, pipe2)
+    finally:
+        local_writer.close()
+
+
+async def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--loading_delay", type=int)
+    parser.add_argument("local")
+    parser.add_argument("remote")
+    args = parser.parse_args()
+
+    local_host, local_port = args.local.split(":")
+    if args.loading_delay:
+        args.loading_delay += time.time()
+
+    server = await asyncio.start_server(
+        lambda r, w: handle_client(r, w, args),
+        local_host,
+        int(local_port),
+    )
+    async with server:
+        await server.serve_forever()
+
+
+asyncio.run(main())
--- a/src/pmproxy/src/redis.c
+++ b/src/pmproxy/src/redis.c
@@ -14,6 +14,8 @@
 #include "server.h"
 #include "discover.h"
 
+#define REDIS_RECONNECT_INTERVAL 60
+
 static int search_queries;
 static int series_queries;
 static int redis_protocol;
@@ -38,6 +40,8 @@ static pmDiscoverCallBacks redis_search = {
 static pmDiscoverSettings redis_discover = {
     .module.on_info	= proxylog,
 };
+
+static void redis_reconnect_worker(void *);
 
 static sds
 redisfmt(redisReply *reply)
@@ -149,6 +153,10 @@ on_redis_connected(void *arg)
     pmNotifyErr(LOG_INFO, "%s setup\n", message);
     sdsfree(message);
 
+    /* Redis was already connected before */
+    if (proxy->redisetup == 1)
+	return;
+
     if (series_queries) {
 	if (search_queries)
 	    redis_series.next = &redis_search;
@@ -168,6 +176,21 @@ on_redis_connected(void *arg)
     }
 
     proxy->redisetup = 1;
+}
+
+static redisSlotsFlags
+get_redis_slots_flags()
+{
+    redisSlotsFlags	flags = SLOTS_NONE;
+
+    if (redis_protocol)
+	flags |= SLOTS_KEYMAP;
+    if (series_queries)
+	flags |= SLOTS_VERSION;
+    if (search_queries)
+	flags |= SLOTS_SEARCH;
+
+    return flags;
 }
 
 /*
@@ -191,20 +214,42 @@ setup_redis_module(struct proxy *proxy)
 
     if (proxy->slots == NULL) {
 	mmv_registry_t	*registry = proxymetrics(proxy, METRICS_REDIS);
-	redisSlotsFlags	flags = SLOTS_NONE;
+	redisSlotsFlags	flags = get_redis_slots_flags();
 
-	if (redis_protocol)
-	    flags |= SLOTS_KEYMAP;
-	if (series_queries)
-	    flags |= SLOTS_VERSION;
-	if (search_queries)
-	    flags |= SLOTS_SEARCH;
 	proxy->slots = redisSlotsConnect(proxy->config,
 			flags, proxylog, on_redis_connected,
 			proxy, proxy->events, proxy);
 	redisSlotsSetMetricRegistry(proxy->slots, registry);
 	redisSlotsSetupMetrics(proxy->slots);
+	pmWebTimerRegister(redis_reconnect_worker, proxy);
     }
+}
+
+static void
+redis_reconnect_worker(void *arg)
+{
+    struct proxy	*proxy = (struct proxy *)arg;
+    static unsigned int	wait_sec = REDIS_RECONNECT_INTERVAL;
+
+    /* wait X seconds, because this timer callback is called every second */
+    if (wait_sec > 0) {
+	wait_sec--;
+	return;
+    }
+    wait_sec = REDIS_RECONNECT_INTERVAL;
+
+    /*
+     * skip if Redis is disabled, state is SLOTS_READY or SLOTS_ERR_FATAL
+     * however: also perform a reconnect if the state is stuck in connecting
+     * or some other state
+     */
+    if (!proxy->slots || proxy->slots->state == SLOTS_READY || proxy->slots->state == SLOTS_ERR_FATAL)
+	return;
+
+    proxylog(PMLOG_INFO, "Trying to connect to Redis ...", arg);
+    redisSlotsFlags	flags = get_redis_slots_flags();
+    redisSlotsReconnect(proxy->slots, flags, proxylog, on_redis_connected,
+			proxy, proxy->events, proxy);
 }
 
 void
--- a/qa/1355
+++ b/qa/1355
@@ -1,0 +1,103 @@
+#!/bin/sh
+# PCP QA Test No. 1355
+# Exercise pmproxy Redis reconnect.
+#
+# Note: This test takes about 3 minutes because it examines
+# the Redis reconnect timer, which runs only once per minute.
+#
+# Copyright (c) 2021 Red Hat.  All Rights Reserved.
+#
+
+seq=`basename $0`
+echo "QA output created by $seq"
+
+# get standard environment, filters and checks
+. ./common.product
+. ./common.filter
+. ./common.check
+
+_check_redis_server_version_offline
+
+_cleanup()
+{
+    cd $here
+    [ -n "$tail_pid" ] && $signal -s TERM $tail_pid
+    [ -n "$pmproxy_pid" ] && $signal -s TERM $pmproxy_pid
+    [ -n "$redisproxy_pid" ] && $signal -s TERM $redisproxy_pid
+    [ -n "$redis_pid" ] && $signal -s TERM $redis_pid
+    $sudo rm -rf $tmp $tmp.*
+}
+
+status=0	# success is the default!
+signal=$PCP_BINADM_DIR/pmsignal
+
+username=`id -u -n`
+
+$sudo rm -rf $tmp $tmp.* $seq.full
+trap "_cleanup; exit \$status" 0 1 2 3 15
+
+_filter_pmproxy_log()
+{
+    grep -oE "Lost connection.+|Trying to connect to Redis.+|Cannot connect to Redis.+|Redis .+ setup"
+}
+
+# real QA test starts here
+echo "Start Redis server"
+redisport=`_find_free_port`
+redis-server --port $redisport --save "" > $tmp.redis.log 2>&1 &
+redis_pid=$!
+pmsleep 0.125
+
+echo "Start Redis proxy to fake loading error"
+redisproxyport=`_find_free_port`
+$python src/redis_proxy.py --loading_delay 30 127.0.0.1:$redisproxyport 127.0.0.1:$redisport > $tmp.redisproxy.log &
+redisproxy_pid=$!
+pmsleep 0.125
+
+echo "Start pmproxy"
+pmproxyport=`_find_free_port`
+pmproxy -f -U $username -x $tmp.err -l $tmp.pmproxy.log -p $pmproxyport -r $redisproxyport -t &
+pmproxy_pid=$!
+pmsleep 0.125
+
+echo
+echo "== Observe a loading error"
+sleep 30
+lineno=1
+tail -n +$lineno $tmp.pmproxy.log | _filter_pmproxy_log
+lineno=$((`wc -l < $tmp.pmproxy.log` + 1))
+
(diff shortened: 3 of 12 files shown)


andreasgerstmayr:
(still needs QA + more testing)

goodwinos:
@andreasgerstmayr I'll review and test this in the morning

andreasgerstmayr:
* replaced exponential backoff with simple 1-minute interval of the redis reconnect worker for simplicity sake
* added `qa/src/redis_proxy.python` script to simulate `LOADING Redis is loading the dataset in memory` error
* handle above error correctly

open TODOs: automated QA test

qa/src/redis_proxy.python
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python3
```
natoscott, line 1: pmpython?

src/pmproxy/src/redis.c
```diff
@@ -170,6 +176,20 @@ on_redis_connected(void *arg)
     proxy->redisetup = 1;
 }
 
+redisSlotsFlags get_redis_slots_flags()
```
natoscott, line 179: Could be declared 'static' and return type is on a separate line usually.

src/pmproxy/src/server.c
```diff
@@ -793,6 +793,7 @@ setup_proxy(uv_timer_t *arg)
     setup_redis_module(proxy);
     setup_http_module(proxy);
     setup_pcp_module(proxy);
+    pmWebTimerRegister(redis_reconnect_worker, proxy);
```
natoscott, line 796: Can this call not be made in setup_redis_module?  Its exposing internal details about Redis outside of redis.c ... be nicer if it moved in there, then redis_reconnect_worker could be made static also.

src/pmproxy/src/server.h
```diff
@@ -219,6 +219,7 @@ extern void close_secure_module(struct proxy *);
 #endif
 
 extern void setup_redis_module(struct proxy *);
+extern void redis_reconnect_worker(void *);
```
natoscott, line 222: Hopefully not necessary.

natoscott (changes requested):

goodwinos:
Hi Andreas, 

in the original patch, redis_reconnect_worker() calls redisSlotsReconnect() which calls redisClusterConnect2(), which wasn't exported or something - I can't really remember but maybe that was why the worker had to be in pmproxy redis.c. But anyway I've built and tested this and it's looking good. I made one small change which was to decrease REDIS_RECONNECT_INTERVAL from 60 to 2. This behaved much better and doesn't have much overhead in the timer worker for the common case where redis is connected. With it set to 60 seconds, clients (e.g. curl requests) would appear to hang when redis was actually up but we were waiting for the reconnect. At 2 seconds, it basically wasn't noticeable.

I tested the following scenarios with pmlogger disabled (so no discovery ingest going on) -
- redis down, start pmproxy, curl request (failed), redis up, pmproxy reconnects, curl request succeeds.
- redis up, start pmproxy, curl req (success), redis down, curl req (fails), redis up, pmproxy reconnect, curl success
- redis up, start pmproxy, redis down and then redis up (with no intervening curl req), curl req succeeds

So that's all good it seems (IIRC the last test above didn't used to work). The error from a failed client request about expecting an array response isn't exactly user friendly (and this was visible in the grafana-pcp dashboard, which will confuse punters looking for the cause)

Remaining testing might involve the above as a unit test, but with: 
- pmlogger enabled with discovery active
- pmlogger_farm enabled with lots of discovery ingest
- multiple parallel http client requests (e.g. from a busy grafana-pcp datasource) in a stress test with redis bouncing up/down periodically
- redis in cluster mode
- remote redis over high latency network

Cheers

qa/src/redis_proxy.python
andreasgerstmayr: :+1:

src/pmproxy/src/server.c
andreasgerstmayr: Yep, good idea.

src/pmproxy/src/server.h
andreasgerstmayr: :+1:

andreasgerstmayr:
Thanks for the reviews!

> Hi Andreas,
> 
> in the original patch, redis_reconnect_worker() calls redisSlotsReconnect() which calls redisClusterConnect2(), which wasn't exported or something - I can't really remember but maybe that was why the worker had to be in pmproxy redis.c.

I think the worker should be in pmproxy/src/redis.c and not libpcp_web, because for the pmseries or pmsearch CLI tools I don't want the reconnect worker - they should just fail (afaics that's what UNIX CLI tools do, there's no retry by default).

> But anyway I've built and tested this and it's looking good. I made one small change which was to decrease REDIS_RECONNECT_INTERVAL from 60 to 2.

Sounds trivial, but what should we do about the log spam? If Redis is unavailable for some time, we spam the logs with
```
[Fri Nov  5 14:43:49] pmproxy(745612) Info: Trying to connect to Redis ...
[Fri Nov  5 14:43:49] pmproxy(745612) Info: Cannot connect to Redis: Connection refused
```
every two seconds. Should I just not display any log messages about reconnects by default? I assume there are configurations out there with Redis configured but not available, and until now there's only one connection attempt at pmproxy startup, but then there will be millions of failed connection attempts to Redis over the lifetime of the pmproxy process.

Another issue with that small interval is that after establishing the connection, we need to get the current keymap and Redis server version, only after that we can consider Redis 'ready'. If this takes too long, it will do another reconnect while one reconnect is running.

Now, one option would be to only attempt to reconnect when Redis is not currently in the progress of reconnecting. However, in case Redis is stuck in the connection attempt (some one-off network/firewall/TCP issue), then we're indefinitely stuck in the connecting state. That's why I try to reconnect always if Redis is not 'ready' (ready == keymap + Redis version loaded).

> This behaved much better and doesn't have much overhead in the timer worker for the common case where redis is connected. With it set to 60 seconds, clients (e.g. curl requests) would appear to hang when redis was actually up but we were waiting for the reconnect. At 2 seconds, it basically wasn't noticeable.

I think the real fix to stuck curl requests is that pmproxy should respond with an error code like HTTP 500, Redis is not available in case Redis is not connected.

> 
> I tested the following scenarios with pmlogger disabled (so no discovery ingest going on) -
> 
>     * redis down, start pmproxy, curl request (failed), redis up, pmproxy reconnects, curl request succeeds.
> 
>     * redis up, start pmproxy, curl req (success), redis down, curl req (fails), redis up, pmproxy reconnect, curl success
> 
>     * redis up, start pmproxy, redis down and then redis up (with no intervening curl req), curl req succeeds

Thanks for testing!

> 
> 
> So that's all good it seems (IIRC the last test above didn't used to work). The error from a failed client request about expecting an array response isn't exactly user friendly (and this was visible in the grafana-pcp dashboard, which will confuse punters looking for the cause)
> 
> Remaining testing might involve the above as a unit test, but with:
> 
>     * pmlogger enabled with discovery active
> 
>     * pmlogger_farm enabled with lots of discovery ingest
> 
>     * multiple parallel http client requests (e.g. from a busy grafana-pcp datasource) in a stress test with redis bouncing up/down periodically
> 
>     * redis in cluster mode
> 
>     * remote redis over high latency network

I currently have a test which is basically a combination of 1 and 3. I'm just thinking about the case of many in-flight Redis queries while a reconnect is happening, where after a successful reconnect some stale callback returns with an outdated "Connection refused" error, triggering another reconnect. Maybe we need to compare timestamps of last reconnect and compare them with the time when the Redis request was sent out, so we can ignore errors pre-dating the last successful Redis reconnect?

andreasgerstmayr:
I've rebased it and ran a full QA run: https://github.com/andreasgerstmayr/pcp/actions/runs/1426620244 (you can download the 'qareport' artifact, which contains the graphical test report in the same design as https://performancecopilot.github.io/qa-reports/reports/latest/).

The new qa/1355 works everywhere except centos8 due to an old Python version (I've added a small TCP proxy written in Python, which fakes the Redis loading error, i.e. for the first X seconds the proxy returns this error, after that it proxies all requests to a real Redis server).

I think it's good to merge as-is, as it doesn't break anything previously working, and fixes the cases where Redis was loading at startup, Redis not available at pmproxy startup, or Redis bouncing during pmproxy runtime.
There might be some edge cases not working yet - see Marks and my comment above - but that'll need to wait.

goodwinos:
Andreas wrote:

> Thanks for the reviews!
> 
> > Hi Andreas,
> > in the original patch, redis_reconnect_worker() calls redisSlotsReconnect() which calls redisClusterConnect2(), which wasn't exported or something - I can't really remember but maybe that was why the worker had to be in pmproxy redis.c.
> 
> I think the worker should be in pmproxy/src/redis.c and not libpcp_web, because for the pmseries or pmsearch CLI tools I don't want the reconnect worker - they should just fail (afaics that's what UNIX CLI tools do, there's no retry by default).

yep makes complete sense

> 
> > But anyway I've built and tested this and it's looking good. I made one small change which was to decrease REDIS_RECONNECT_INTERVAL from 60 to 2.
> 
> Sounds trivial, but what should we do about the log spam? If Redis is unavailable for some time, we spam the logs with
> 
> ```
> [Fri Nov  5 14:43:49] pmproxy(745612) Info: Trying to connect to Redis ...
> [Fri Nov  5 14:43:49] pmproxy(745612) Info: Cannot connect to Redis: Connection refused
> ```
> 
> every two seconds. Should I just not display any log messages about reconnects by default? I assume there are configurations out there with Redis configured but not available, and until now there's only one connection attempt at pmproxy startup, but then there will be millions of failed connection attempts to Redis over the lifetime of the pmproxy process.

Discovery had a similar spamming issue when a pmlogger lock was found (at which time it needs to stop ingesting until the lock clears). I ended up just wrapping the whole lot in pmDebug.discovery because it was just too verbose. However it might be useful in the pmproxy case to notify once that redis is down and the reconnection protocol has been triggered. Then stay silent until pmproxcy connects to redis again, and announce that it's back. That way admins can check pmproxy.log and will know to go off and start redis (if that's actually what they want).

> 
> Another issue with that small interval is that after establishing the connection, we need to get the current keymap and Redis server version, only after that we can consider Redis 'ready'. If this takes too long, it will do another reconnect while one reconnect is running.

that's a bit messy. I assume re-authentication would be needed too, if enabled.

> 
> Now, one option would be to only attempt to reconnect when Redis is not currently in the progress of reconnecting. 

guess so. 

> However, in case Redis is stuck in the connection attempt (some one-off network/firewall/TCP issue), then we're indefinitely stuck in the connecting state. That's why I try to reconnect always if Redis is not 'ready' (ready == keymap + Redis version loaded).
> 
> > This behaved much better and doesn't have much overhead in the timer worker for the common case where redis is connected. With it set to 60 seconds, clients (e.g. curl requests) would appear to hang when redis was actually up but we were waiting for the reconnect. At 2 seconds, it basically wasn't noticeable.

I think Nathan has some comments on the default reconnect interval - in my testing it just seemed to work better than having to wait up to a full minute before anything was working again (after bouncing redis server)

> 
> I think the real fix to stuck curl requests is that pmproxy should respond with an error code like HTTP 500, Redis is not available in case Redis is not connected.

sounds much better than "Error, Array expected" !

> 
> > I tested the following scenarios with pmlogger disabled (so no discovery ingest going on) -
> > ```
> > * redis down, start pmproxy, curl request (failed), redis up, pmproxy reconnects, curl request succeeds.
> > 
> > * redis up, start pmproxy, curl req (success), redis down, curl req (fails), redis up, pmproxy reconnect, curl success
> > 
> > * redis up, start pmproxy, redis down and then redis up (with no intervening curl req), curl req succeeds
> > ```
> 
> Thanks for testing!
> 
> > So that's all good it seems (IIRC the last test above didn't used to work). The error from a failed client request about expecting an array response isn't exactly user friendly (and this was visible in the grafana-pcp dashboard, which will confuse punters looking for the cause)
> > Remaining testing might involve the above as a unit test, but with:
> > ```
> > * pmlogger enabled with discovery active
> > 
> > * pmlogger_farm enabled with lots of discovery ingest
> > 
> > * multiple parallel http client requests (e.g. from a busy grafana-pcp datasource) in a stress test with redis bouncing up/down periodically
> > 
> > * redis in cluster mode
> > 
> > * remote redis over high latency network
> > ```
> 
> I currently have a test which is basically a combination of 1 and 3. I'm just thinking about the case of many in-flight Redis queries while a reconnect is happening, where after a successful reconnect some stale callback returns with an outdated "Connection refused" error, triggering another reconnect. Maybe we need to compare timestamps of last reconnect and compare them with the time when the Redis request was sent out, so we can ignore errors pre-dating the last successful Redis reconnect?

yes - we need to keep state and/or a timer to avoid the re-re-connect scenario.

natoscott:
> > every two seconds. Should I just not display any log messages about reconnects by default?

Other PCP client daemons that do reconnect (pmie, pmlogger) take the approach of logging once on connection loss, then once again when the connection is successfully re-established.  This is what the user/admin needs to know - nothing else.

> I assume there are configurations out there with Redis configured but not available, and until now there's only one connection attempt at pmproxy startup, but then there will be millions of failed connection attempts to Redis over the lifetime of the pmproxy process.

We need to change that IMO (even on once-per-minute) - that's just too verbose, and is not end-user-useful information for the most part.
 
> > Another issue with that small interval is that after establishing the connection, we need to get the current keymap and Redis server version, only after that we can consider Redis 'ready'. If this takes too long, it will do another reconnect while one reconnect is running.

There are multiple independent states there I guess - the initial socket-connection-established state (which takes very little time, usually, even for remote redis-servers) is different to "fully ready" to service requests state.

> > Now, one option would be to only attempt to reconnect when Redis is not currently in the progress of reconnecting.
> 
> guess so.

No question - this is an issue (race) even with long intervals, we're just hiding the underlying problem by introducing a large (unfortunately user-visible) delay.
 
> > However, in case Redis is stuck in the connection attempt (some one-off network/firewall/TCP issue), then we're indefinitely stuck in the connecting state.

This is not a common scenario FWIW.  We need to handle it (timeouts?) but doing so shouldn't cause a poor user experience everywhere else, especially in the common cases.

> > That's why I try to reconnect always if Redis is not 'ready' (ready == keymap + Redis version loaded).

Perhaps separating out socket-level connection establishment to these other high-order aspects of Redis 'readiness' could help reduce/remove the overlapping-reconnection race conditions.

> > > This behaved much better and doesn't have much overhead in the timer worker for the common case where redis is connected. With it set to 60 seconds, clients (e.g. curl requests) would appear to hang when redis was actually up but we were waiting for the reconnect. At 2 seconds, it basically wasn't noticeable.
> 
> I think Nathan has some comments on the default reconnect interval - in my testing it just seemed to work better than having to wait up to a full minute before anything was working again (after bouncing redis server)

Oh, simply that the 60 seconds is too long to delay for retry - its going to be very visible to end users who will be confused and have to spend time diagnosing it (sounds like it was the first thing Mark noticed when he tried this out).

The very common scenario of 'service restart redis' is one to optimize for - this should "always" (unless under adverse conditions, etc) be near-instant to provide a transparent user experience, and to avoid us continuing to get complaints.

> > I think the real fix to stuck curl requests is that pmproxy should respond with an error code like HTTP 500, Redis is not available in case Redis is not connected.
> 
> sounds much better than "Error, Array expected" !

+1

> > I currently have a test which is basically a combination of 1 and 3. I'm just thinking about the case of many in-flight Redis queries while a reconnect is happening, where after a successful reconnect some stale callback returns with an outdated "Connection refused" error, triggering another reconnect. Maybe we need to compare timestamps of last reconnect and compare them with the time when the Redis request was sent out, so we can ignore errors pre-dating the last successful Redis reconnect?
> 

Or (more simple?) sequence numbers perhaps - i.e. every request gets an associated connection seq# and each successful reconnect increments that number.  Using timestamps may prove error prone when the system clock is adjusted.

> yes - we need to keep state and/or a timer to avoid the re-re-connect scenario.

That hard-coded delay of 60 seconds should probably be configurable too BTW.

andreasgerstmayr:
I've reduced log spam now as you both suggested. It was actually quite simple, we just need to check to *not* log multiple failed attempts in a row, because multiple successful attempts won't happen anyway in a row (for a successful connection, there is no re-connect of course). The changes are quite small and low risk: #1471.

I've implemented the other changes in PR #1470.


> > Another issue with that small interval is that after establishing the connection, we need to get the current keymap and Redis server version, only after that we can consider Redis 'ready'. If this takes too long, it will do another reconnect while one reconnect is running.

> that's a bit messy. I assume re-authentication would be needed too, if enabled.

Fortunately the auth is handled by hiredis-cluster already for every new connection.

> Perhaps separating out socket-level connection establishment to these other high-order aspects of Redis 'readiness' could help reduce/remove the overlapping-reconnection race conditions.

We already have that, first is `SLOTS_CONNECTED`, second is `SLOTS_READY`.

> > However, in case Redis is stuck in the connection attempt (some one-off network/firewall/TCP issue), then we're indefinitely stuck in the connecting state.

> This is not a common scenario FWIW. We need to handle it (timeouts?) but doing so shouldn't cause a poor user experience everywhere else, especially in the common cases.

I think stale/broken connections due to faulty loadbalancers, routers, switches, SDN, or any other network gear are not that uncommon. You're right, the configured timeouts should handle this. I think reconnecting in all cases if Redis is not ready would be more robust, but I'm not sure how to prevent re-re-connects with a small interval like 2s. I've changed it now as you suggested to only attempt to reconnect in the `SLOTS_DISCONNECTED` state. I made some changes in the response callback, to track errors in all states, so we should also catch errors in the phase between Redis connection established and Redis is ready.

> Or (more simple?) sequence numbers perhaps - i.e. every request gets an associated connection seq# and each successful reconnect increments that number. Using timestamps may prove error prone when the system clock is adjusted.

Good idea, I've implemented it that way now. It was actually quite a small & clean patch, as we already have the required data structures and callbacks in place (the `redisSlotsReplyData` struct and related functions/callbacks). See #1470.

As it's literally the last hours before the release and thus not as well tested as it should be, I'm not very sure if we should include it in the upcoming release or not. The first PR #1471 however (re log spam) is low-risk and can definitely be included imho.

natoscott:
> [...] I'm not very sure if we should include it in the upcoming release or not.

It's sufficiently important and user-affecting that I think its worth pulling in after close review - its surprisingly little code change.
We can delay the release a little while today too so that we can test this further.

PR #1468, opened 2021-11-08
pmproxy: don't connect to Redis if it's not required

andreasgerstmayr:
only connect to Redis if one of the following is enabled
* Redis proxying
* pmseries REST API
* pmsearch REST API
* archive discovery


natoscott (approved):
LGTM

PR #1479, opened 2021-11-19
pmdabpf: support multiple arches

andreasgerstmayr:
* pmdabpf: added support for arm64 and powerpc
* reinstate full vmlinux.h file, from upstream iovisor/bcc. let's consider this file as a blob (as it's a generated file) and not manually mess with it - sorry for criticising the exact same thing on the initial review of Jason's bpf PMDA PR :|
* additionally, fix runqlat bpf module for kernels 5.14 (rename of state field in task struct)


andreasgerstmayr:
as we're already requiring bpftool, we could also generate the vmlinux.h on-the-fly during the build? need to verify if BTF is enabled on all distros which we want to support for this PMDA.

andreasgerstmayr:
@jasonk000 what do you think, could we generate the `vmlinux.h` file on-the-fly during build time? It'd have the additional advantage of now having to worry about different architectures, not having a big blob in the git sources and/or not having to cut/extract the important parts of the vmlinux.h every time we update some bpf PMDA module.

The disadvantage is that it will effectively disable the bpf PMDA on distribution releases which don't have BTF enabled yet by default, for example Ubuntu 20.04 LTS. If you recompile PCP on a BTF-enabled Ubuntu 20.04 however, it will work and include the bpf PMDA.

jasonk000:
As long as user can provide / override / default the path to some location for vmlinux or the BTF information then a solution should be easy for user to self-provide.

Generating the file at build time to support setup makes sense to me, especially across multiple architectures. It does have the complication you mentioned. In our deployment we provide vmlinux BTF information out of band -- ie: we still use a kernel that does not provide BTF, and have a separate packaging process to deliver BTF.

So, the reason we'd need to provide the path is because we have the BTF available outside the kernel:
```
bpftool btf dump file /sys/kernel/btf/vmlinux format c
```
vs
```
bpftool btf dump file /boot/vmlinux-5.4.0-1057-aws format c
```

jasonk000:
@andreasgerstmayr this works great, `--with-pmdabpf-btf=` and `--with-pmdabpf-bpftool=` do allow me to continue custom building against kernels that do not have BTF enabled at build time. Thank you!

PR #1488, opened 2021-12-02
pmproxy: fix race on teardown of pcp protocol proxy clients

natoscott:


kmcdonell:
I've pulled these changes to vm39, and qa/294 is passing there now ... so the "corrupted double-linked list" issue I saw earlier (that was a hard failure) has been resolved.
And check -g pmproxy passes all tests, so no regressions that the QA suite can expose.
Looks good, thanks @natoscott.

PR #1489, opened 2021-12-05
logutil: add path handler for pmlogger_farm_check service

goodwinos:
The pmlogger_farm_check service is responsible for checking pmlogger
farm processes, restarting any pmloggers that have failed and for
starting any new pmlogger farm processes after control file changes.

The service is driven by pmlogger_farm_check.timer, which starts
the service every 5 minutes. When the service is triggered by the
timer, it simply runs 'pmlogctl -m check' and then exits. See
pmlogctl(1) for details. This works fine but it does mean there may
be up to a five minute delay before any pmlogger control file changes
will be automatically applied to the pmlogger farm.

This commit adds a systemd path handler to the pmlogger_farm_check
service to also trigger the service when there are any changes to
$PCP_SYSCONF_DIR/pmlogger/control or to the control file directory
$PCP_SYSCONF_DIR/pmlogger/control.d (i.e. new control files are added
or changes have been made to any existing control files). The trigger
is applied only when an fd for the file or directory is closed (not
when it is opened or written to).

The existing timer is retained - to periodically check/migrate/restart
any dead loggers, regardless of whether there have been any control
file changes.

This has been manually tested but will need a new QA test - TODO.

### DIFF
--- a/build/rpm/pcp.spec.in
+++ b/build/rpm/pcp.spec.in
@@ -2861,6 +2861,7 @@ then
 	%systemd_preun pmlogger_daily.timer
 	%systemd_preun pmlogger_check.timer
 	%systemd_preun pmlogger_farm_check.timer
+	%systemd_preun pmlogger_farm_check.path
 	%systemd_preun pmie_farm_check.timer
 
 	systemctl stop pmlogger.service >/dev/null 2>&1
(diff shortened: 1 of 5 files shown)


natoscott:
We did this once before for the pmlogger/pmie services and it was an unmitigated disaster of race conditions and systemd cyclic dependencies.  What's changed that we'd revisit this again now?

natoscott (changes requested):
Before a path-based solution is proposed again, the circular service dependency loop that affected pmlogger and networking a year or two back would need to be resolved - pmlogger_farm has the same After=network-online setup that was implicated in that original conundrum.

natoscott:
Superceded by https://github.com/performancecopilot/pcp/pull/1495

PR #1491, opened 2021-12-07
pmproxy: when secure.enabled is false, do not attempt to init ssl context even for https requests

### LINKED ISSUE #1490: https request to pmproxy with secure.enabled = false triggers segfault
### Summary

Setting `secure.enabled = false` segfaults when it receives a HTTPS incoming request. Plaintext HTTP is ok.

### Reproduce

This can be reproduced by changing the test for 1475 to set `secure.enabled = false` and re-running the test; the test results in a segfault.

```
=== checking serial https/TLS operation ===
./1457: line 165: 19860 Segmentation fault      (core dumped) $_valgrind_clean_assert pmproxy -f -l- --timeseries -c $tmp.conf -p $port -U $username > $tmp.valout 2> $tmp.valerr

...

=== valgrind stdout ===
Invalid write of size 4
at 0x52F6C74: SSL_set_accept_state (in /usr/lib/x86_64-linux-gnu/libssl.so.1.1)
by 0x11DAB7: ??? (in /usr/lib/pcp/bin/pmproxy)
by 0x112BD3: ??? (in /usr/lib/pcp/bin/pmproxy)
by 0x5A2A5CE: ??? (in /usr/lib/x86_64-linux-gnu/libuv.so.1.0.0)
by 0x5A2B33B: ??? (in /usr/lib/x86_64-linux-gnu/libuv.so.1.0.0)
by 0x5A3033F: uv__io_poll (in /usr/lib/x86_64-linux-gnu/libuv.so.1.0.0)
by 0x5A20CC7: uv_run (in /usr/lib/x86_64-linux-gnu/libuv.so.1.0.0)
by 0x1112FD: ??? (in /usr/lib/pcp/bin/pmproxy)
by 0x110715: ??? (in /usr/lib/pcp/bin/pmproxy)
by 0x6130C0E: (below main) (libc-start.c:310)
Address 0x38 is not stack'd, malloc'd or (recently) free'd
```

### Issue

In pmproxy initialisation, ordinarily, the `proxy->ssl` value is set to the SSL context, filled with certs etc.

However, if `secure.enabled = false` is read by `setup_secure_module` to [decide whether to proceed with SSL](https://github.com/performancecopilot/pcp/blob/6e941d444d2fc19ceb3213974129bd1b59cd15fd/src/pmproxy/src/secure.c#L293-L296) then context setup does not complete; the value `proxy->ssl` is not set.
```
if ((option = pmIniFileLookup(config, "pmproxy", "secure.enabled"))) {
    if (strncmp(option, "true", sdslen(option)) != 0)
        return;
}

.... line 375 never reached ....

proxy->ssl = context;
```

- Later, during request processing, the [initial characters of the request are read to determine the protocol type](https://github.com/performancecopilot/pcp/blob/6e941d444d2fc19ceb3213974129bd1b59cd15fd/src/pmproxy/src/server.c#L389-L394).
- The read path then takes the code through [here](https://github.com/performancecopilot/pcp/blob/6e941d444d2fc19ceb3213974129bd1b59cd15fd/src/pmproxy/src/server.c#L437-L438) .. 
- .. where it attempts to [perform first-time setup](https://github.com/performancecopilot/pcp/blob/6e941d444d2fc19ceb3213974129bd1b59cd15fd/src/pmproxy/src/secure.c#L116-L120).
- This fails because of the [missing proxy->ssl value](https://github.com/performancecopilot/pcp/blob/6e941d444d2fc19ceb3213974129bd1b59cd15fd/src/pmproxy/src/secure.c#L84-L89) and .. 
- lands us at the `SSL_set_accept_state` segfault described earlier.

### Logs

With debug for http and auth, the following shows:
```
on_secure_client_read: client 0x55833529cc90
setup_secure_client: SSL/TLS connection initiated by client 0x55833529cc90
(EOF)
```

jasonk000:
fixes #1490

### DIFF
--- a/qa/1458
+++ b/qa/1458
@@ -1,0 +1,218 @@
+#!/bin/sh
+# PCP QA Test No. 1458
+# Exercise access pmproxy with secure.enabled = false
+#
+# The main purpose of this is to test that the component works correctly
+# when secure.enabled = false; we can expect the https URLs to fail.
+#
+# See https://github.com/performancecopilot/pcp/issues/1490
+
+# Copyright (c) 2019,2021 Red Hat
+# Modified by Netflix, Inc.
+#
+
+seq=`basename $0`
+echo "QA output created by $seq"
+
+# get standard environment, filters and checks
+. ./common.product
+. ./common.filter
+. ./common.check
+
+_check_valgrind
+openssl help 2>/dev/null || _notrun "No openssl binary found"
+
+if [ -f /etc/lsb-release ]
+then
+    . /etc/lsb-release
+    if [ "$DISTRIB_ID" = Ubuntu ]
+    then
+	# This test fails for Ubuntu 19.10 with a myriad of errors involving
+	# the use of uninitialized values.  The code paths very but typically
+	# involve libuv -> libssl ->  libcrypto
+	#
+	case "$DISTRIB_RELEASE"
+	in
+	    19.10)
+		_notrun "problems with libuv, libssl, libcrypto and valgrind on Ubuntu $DISTRIB_RELEASE"
+		;;
+	esac
+    fi
+fi
+
+_cleanup()
+{
+    cd $here
+    if $need_restore
+    then
+	need_restore=false
+	_restore_config $PCP_SYSCONF_DIR/labels
+	_sighup_pmcd
+    fi
+    $sudo rm -rf $tmp $tmp.*
+}
+
+status=1	# failure is the default!
+need_restore=false
+username=`id -u -n`
+$sudo rm -rf $tmp $tmp.* $seq.full
+trap "_cleanup; exit \$status" 0 1 2 3 15
+
+_check_empty()
+{
+    tee -a $seq.full > $tmp.unfiltered
+    if [ -s $tmp.unfiltered ]
+    then
+	echo "Botch: got output from curl"
+    else
+	echo "Good!, empty output from curl"
+    fi
+}
+
+_filter_json()
+{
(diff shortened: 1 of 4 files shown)


kmcdonell:
@jasonk000 Could you please add qa/1458.out? Thanks.

natoscott:
> @jasonk000 Could you please add qa/1458.out? Thanks.

I've pushed this now so that its in place for tonight's full CI run - @jasonk000 please double-check?  Thanks.

kmcdonell:
I suspect there is more work to be done here (for qa/1458) ... I have 3 different failure signatures in the QA Farm (after qa/1458.out was available).

jasonk000:
Hi @natoscott, I see Ken's note, but I'm not sure how to progress it further from here - could you share the details from CI?

natoscott:
@jasonk000 see Ken's private mail from earlier today.  Summary follows...

There's basically two main types of failure (across the five slightly different failing platforms).  In one case, curl consistently gives what looks like an empty response - I'm not sure about root cause there yet but possibly this is on platforms without libuv (I'll add a check for that).  In the other case, we see spurious pmproxy.log messages about failed attempts to connect to Redis.  This is a known issue Andreas and I have been discussing recently, where libpcp_web will fallback to attempting creation of a Redis connection even when pmproxy is configured not to use Redis ... we can filter the message out for now, its unrelated to this fix, or I might tackle the underlying libpcp_web problem if I get time today.

kmcdonell:
Small correction, I've seen 5 failure cases across 9 different failing platforms.

PR #1509, opened 2021-12-31
bugfix: if total_metrics ==0 and calloc() may return un-NULL address. pmdaInit() will failed!

jinqinghui:


natoscott:
Merged with minor additions in commit message.

PR #1512, opened 2022-01-09
Fix elasticsearch pmda

### LINKED ISSUE #1511: pmdaelasticsearch does not scale with large number of shards
We have a V1 cluster with around 22,000+ active shards (roughly 7,000 indices).  This results in the `_cluster/state` API returning 68 MB of JSON.  With the API being called potentially twice (once just to get the name of the master node, and again for indices metrics (gateway snapshot interval, number of replicas, number of shards, version created), it can severely limit the operation of the PMDA.

portante:
Fixes #1511 in part.

This corrects a bug with the `_cluster/state` API being invoked at the wrong time (broken `if` condition), uses the `_nodes` API instead of the `_cluster/state` API to determine the master node, and also uses the names of the nodes instead of their IDs.

Further, we perform a bit of refactoring based on `pylint` feedback, and apply some helpful formatting and tableization which made it easier to find problems in the metric defintions.

Finally we added a number of additional metrics in four separate commits to make it easier to see the progression of additions.

It would be great to find a way to only call the `_nodes` and `_stats/search` APIs once, instead of twice, though.

### DIFF
--- a/src/pmdas/elasticsearch/pmdaelasticsearch.python
+++ b/src/pmdas/elasticsearch/pmdaelasticsearch.python
@@ -232,6 +232,38 @@ class ElasticsearchPmda(PMDA):
             [ 'nodes.indices.indexing.is_throttled',                                   PM_TYPE_U32,    PM_SEM_INSTANT, pmUnits(0,0,0,0,0,0),             '' ], # 94
             [ 'nodes.indices.indexing.throttle_time_in_millis',                        PM_TYPE_U64,    PM_SEM_COUNTER, pmUnits(0,1,0,0,PM_TIME_MSEC,0),  '' ], # 95
             [ 'nodes.indices.store.throttle_time_in_millis',                           PM_TYPE_U64,    PM_SEM_COUNTER, pmUnits(0,1,0,0,PM_TIME_MSEC,0),  '' ], # 96
+            [ 'nodes.indices.warmer.current',                                          PM_TYPE_U64,    PM_SEM_INSTANT, pmUnits(0,0,0,0,0,0),             '' ], # 97
+            [ 'nodes.indices.warmer.total',                                            PM_TYPE_U64,    PM_SEM_COUNTER, pmUnits(0,0,1,0,0,PM_COUNT_ONE),  '' ], # 98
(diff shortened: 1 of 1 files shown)


portante:
I also see debug output from the Elasticsearch PMDA (the ranges are a way to summarize all the lines emitted, 3,800+):

```
[Mon Jan 17 05:13:50] pmdaelasticsearch(24500) Debug: pmdaLabel: InDom 108.2[(0-3884)]: Missing metric value(s)
[Mon Jan 17 05:13:50] pmdaelasticsearch(24500) Debug: pmdaLabel: InDom 108.1[(11-21)]: Missing metric value(s)
[Mon Jan 17 05:13:50] pmdaelasticsearch(24500) Debug: pmdaLabel: InDom 108.1[(11-21)]: Missing metric value(s)
```

natoscott:
> It would be great to find a way to only call the _nodes and _stats/search APIs once, instead of twice, though.

This is possible.  There is a python fetch callback - use set_fetch() method to add your own - which is called once per fetch PDU, and importantly its called before the refresh() method used by this PMDA.  If a custom fetch callback method just clears a boolean class variable (named like self.search_called perhaps), we can test and set that same boolean in the refresh method to ensure we only call the search API once.

The rest of this change is looking really good (I've got only one small local tweak so far) - just running tests on it now and will merge shortly, perhaps tackle the above optimization via a followup PR?  Let me know if its not clear & I'll send sample code.

portante:
@natoscott I'd be happy to work on the `set_fetch()` method.  Let me know how you want to approach it.

natoscott:
@portante something like this should do it (untested) - 
https://gist.github.com/natoscott/0fffd13d5f8d752b5ab7084c40607dbf

natoscott:
| [Mon Jan 17 05:13:50] pmdaelasticsearch(24500) Debug: pmdaLabel: InDom 108.2[(0-3884)]: Missing metric value(s)

Regarding those debug lines, I'm not sure where that's originating from actually.  The closest I can find is these two diagnostics from libpcp_pmda:
libpcp_pmda/src/callback.c:	    fprintf(stderr, "pmdaLabel: InDom %s labels request\n",
libpcp_pmda/src/callback.c:	    fprintf(stderr, "pmdaLabel: InDom %s %d instance labels request\n",

... similar, but doesn't quite match with what you see.  Very odd.

PR #1534, opened 2022-02-23
Remove "lockdown" selinux class

cjeanner:
With newer kernel, this class doesn't exist at all; selinux-policy also
removed it lately. Keeping this lockdown call makes the package
impossible to install on newer systems such as centos-9 or rhel-9, and
this further breaks other selinux related packages in such a way we
can't easily recover.

### DIFF
--- a/configure
+++ b/configure
@@ -738,7 +738,6 @@ pcp_selinux_security
 pcp_selinux_sudo_exec
 pcp_selinux_initrc_tmp
 pcp_selinux_glusterd_log
-pcp_selinux_lockdown_class
 pcp_selinux_icmp_socket_class
 pcp_selinux_capability2_bpf
 pcp_selinux_capability2_syslog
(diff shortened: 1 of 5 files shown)


kmcdonell (changes requested):
Sorry, I'm not following this.  The PCP build already has this as a conditional (we build for older distros as well), and as soon as seinfo(1) stops reporting that the class "lockdown" exists, we won't include the PCP_LOCKDOWN_CLASS and PCP_LOCKDOWN_RULE fragments in pcpupstream.te.

If you have a system where the kernel does not support the lockdown class, but seinfo(1) reports that it exists, then that would appear to be an issue with seinfo(1).

Please advise if this is the case.

Culling this stuff from configure.ac, src/builddefs.in and src/selinux is unlikely to be the correct solution here.

cjeanner:
Hello,

Well, apparently there's something fishy. We're hitting the issue while testing the newer rhel9; apparently there was some internal manipulations in order to actually remove the references from the code. There's this BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=2041503

apparently, even rebuilding with the newer kernel/selinux-policy didn't drop the references to lockdown :/.

I can't really say more, I'm just the "end-user" in this chain, unfortunately. But if you feel my patch is wrong, just discard it.

Cheers,

C.

natoscott:
> [](https://bugzilla.redhat.com/show_bug.cgi?id=2041503)
> 
> apparently, even rebuilding with the newer kernel/selinux-policy didn't drop the references to lockdown :/.

What is the full version of the pcp-selinux package you have installed there?  The rebuild from that BZ that picked up the changes for selinux lockdown rules was verified by our QE folk so I'm a bit surprised you're still seeing this issue.

It's also possible some other selinux-policy change happened, causing the "fixed" pcp-selinux policy to not apply on your system... we see that issue from time to time also.

Unfortunately, there are pcp + kernel version combinations out there that need the lockdown policy, so it has to stay now.

natoscott:
Closing as per last comment.

PR #1552, opened 2022-03-07
libpcp: support high resolution interval time parsing

natoscott:
Add a new pmParseHighResInterval(3) interface into libpcp.

Resolves https://github.com/performancecopilot/pcp/issues/1375


kmcdonell (approved):
Looks good ... libpcp and qa apps build OK, man page reads OK, qa/316 passes

PR #1556, opened 2022-03-10
libpcp_web: fix memleak: free 'metric' member of seriesGetSID struct

andreasgerstmayr:


src/libpcp_web/src/query.c
```diff
@@ -190,6 +190,8 @@ freeSeriesGetSID(seriesGetSID *sid)
 
     seriesBatonCheckMagic(sid, MAGIC_SID, "freeSeriesGetSID");
     sdsfree(sid->name);
+    if (sid->metric)
```
natoscott, line 193: No need to check for null here - sdsfree() does this.  I'll tweak it when merging.

natoscott (approved):

PR #1558, opened 2022-03-14
vendor: vendor libbpf (required by bpf PMDA)

andreasgerstmayr:
The bpf PMDA requires an up-to-date version of libbpf (v0.6+) which isn't yet available in the main (stable) distribution repositories.


andreasgerstmayr:
Let's check first if it's feasible to implement workarounds to use the older libbpf version (libbpf 0.4 is in Ubuntu 21.10 and Fedora 35), and only if it's too much work or not possible, vendor libbpf.

Issue #1377, opened 2021-08-30
libpcp: pmOptions has 4 struct timevals buried in it (start, finish, origin and interval)


natoscott:
We do have a "version" field here, maybe that can be used to allow NEW and optional struct timespec "HighRes" alternates at the end of the struct?  ... we really DO NOT want to replicate all of pmgetopt_r(), pmGetOptions(), pmGetContextOptions(), pmUsageMessage(), pmFreeOptions() and the override method type.

Issue #1372, opened 2021-08-30
libpcp: pmGetHighResArchiveLabel()


natoscott:
Need pmHighResResult as arg (typical use case is pmFetch (becomes pmFetchHighRes), fiddle with pmValue in pmResult (becomes fiddle with pmValue in pmHighResResult) and pmStore (becomes pmStoreHighRes)

kmcdonell:
I'm not sure we really need this.  The V2-V3 differences are all factored away behind __pmLogLoadLabel() which is callable for libpcp.h users, not pmapi.h users.  The returned __pmLogLabel data structure contains the start time as a __pmTimestamp, so Y2038 safe and nsec precision.
Potential users of pmGetHighResArchiveLabel() like pmdumplog, pmlogcheck, pmlogextract, pmloglabel and pmlogsize are already using __pmLogLoadLabel().
If we do need to add pmGetHighResArchiveLabel() to the PMAPI, then it is one more data structures with a struct timespec field and a thin shim around __pmLogLoadLabel().

natoscott:
There's two aspects to this that led me to thinking we should add this at the general PMAPI level.

- it'd be most excellent if at the end of this exercise, we can write client tools using just one structure (and common units) to represent time.  So, where people choose to use struct timespec and nanosecond precision, I'd like to see all functionality available to such tools, and not see a need for code to convert between timeval / timespec for this special case.  But, yeah, this is a relatively uncommon API so maybe this is not good enough a reason in this case.

- we need an interface to expose the zoneinfo to client tools (and probably the enabled/disabled feature bits too) as we do with the original squashed timezone, in case people want to use this in a user interface or some other situation outside of libpcp.  If we accept we need that, then we may as well export it via a new HighRes API I guess.

kmcdonell:
OK, but this means we end up with THREE variants of the pmLogLabel data structure (sigh).  I thought about possibly prompting __pmLogLabel to the PMAPI, but it would need struct timespec and we'd then need {struct timespec} <-> __pmTimestamp conversions in all the places where we use __pmLogLabel at the moment, or perhaps change to struct timespec internally (this would mean redoing a lot of what's been done already, along with changes on the archive i/o paths to convert between tv_nsec (64-bit) and nsec (32-bit)).
Not sure there is an elegant option here, but I think I'd vote for THREE pmLogLabel data structures (2 @ PMAPI + 1 internal) as the least evil of the options.

natoscott:
[...]

PR #1560, opened 2022-03-15
libpcp: high resolution log label and time window APIs

natoscott:
Provide new APIs to access the v3 log labels start timestamp
in high resolution form, as well as optional zoneinfo field,
in the form of pmGetHighResArchiveLabel(3).  The context end
of archive timestamp is made available in high resolution too
via a new pmGetHighResArchiveEnd routine.

Update the pmGetOptions routine to support a v3 ABI such that
PMAPI client tools can opt-in to working with nanosecond time
throughout.  pmval is converted as a proof of concept.

Resolves https://github.com/performancecopilot/pcp/issues/1372
Resolves https://github.com/performancecopilot/pcp/issues/1374
Resolves https://github.com/performancecopilot/pcp/issues/1377

### DIFF
--- a/man/man1/pmval.1
+++ b/man/man1/pmval.1
@@ -1,7 +1,7 @@
 '\"! tbl | mmdoc
 '\"macro stdmacro
 .\"
-.\" Copyright (c) 2015-2018 Red Hat.
+.\" Copyright (c) 2015-2018,2022 Red Hat.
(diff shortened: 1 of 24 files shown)


kmcdonell (approved):
There is a bucket load of code here 8^)
I have not done a line-by-line review, but checked the API changes and these are all as expected.
We have good QA coverage of all this functionality, so I vote for merging this ASAP and letting the QA suite do what it was designed to do.

PR #1564, opened 2022-03-17
pmdabpf: resolve memory leaks, destroy bpf skeleton object at shutdown

andreasgerstmayr:
Every bpf PMDA module gets two parameters passed in the init function: `init(dict *cfg, char *module_name)`
`cfg` is the global configuration object (with contents of `bpf.conf`) and `module_name` is for example `biolatency.so`.

This PR `free`s the configuration object at shutdown, and the `module_name` after calling `init()` of the module.

imho that's inconsistent, and I'm now very happy with it - however, what's an elegant way to keep references to the `module_name` of each module and `free` it in `bpf_shutdown`? One option would be to create a linked list of these pointers, but I'm sure there is a better option ... Maybe just pass the memory allocated by `pmdaCacheStore` to the `init()` function? Doesn't seem very neat either.

### DIFF
--- a/src/pmdas/bpf/bpf.c
+++ b/src/pmdas/bpf/bpf.c
@@ -61,6 +61,8 @@ static pmdaOptions opts = {
     .long_options = longopts,
 };
 
+dict *pmda_config;
+
 /*
  * callback provided to pmdaFetch
  */
@@ -192,6 +194,7 @@ bpf_load_modules(dict *cfg)
     sds *entry_key_split;
     sds sds_enabled;
     sds sds_true;
+    unsigned int cluster_id;
 
     sds_enabled = (sdsnew("enabled"));
     sds_true = (sdsnew("true"));
@@ -235,6 +238,15 @@ bpf_load_modules(dict *cfg)
             continue;
         }
 
+        cluster_id = pmdaCacheStore(CACHE_CLUSTER_IDS, PMDA_CACHE_ADD, module_name, bpf_module);
+        free(module_name);
+        // replace module_name with a pointer to memory allocated by pmdaCacheStore
+        ret = pmdaCacheLookup(CACHE_CLUSTER_IDS, cluster_id, &module_name, NULL);
+        if (ret < 0) {
+            pmNotifyErr(LOG_ERR, "failed to load module name from cache: %s\n", pmErrStr(ret));
+            continue;
+        }
+
         pmNotifyErr(LOG_INFO, "initialising %s", module_name);
 
         // module_name is passed so that the recipient can prepend the string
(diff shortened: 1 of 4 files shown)


src/pmdas/bpf/bpf.c
```diff
@@ -61,6 +61,8 @@ static pmdaOptions opts = {
     .long_options = longopts,
 };
 
+dict *pmda_config = NULL;
```
natoscott, line 64: No need to initialize globals to 0 or NULL, the compiler mandates this.

src/pmdas/bpf/modules/biolatency.c
```diff
@@ -7,6 +7,7 @@
 #define NUM_LATENCY_SLOTS 63
 pmdaInstid biolatency_instances[NUM_LATENCY_SLOTS];
 
+struct biolatency_bpf *bpf_obj = NULL;
```
natoscott, line 10: As above.

src/pmdas/bpf/modules/runqlat.c
```diff
@@ -7,6 +7,7 @@
 #define NUM_LATENCY_SLOTS 63
 pmdaInstid runqlat_instances[NUM_LATENCY_SLOTS];
 
+struct runqlat_bpf *bpf_obj = NULL;
```
natoscott, line 10: As above.

natoscott (changes requested):

src/pmdas/bpf/bpf.c
andreasgerstmayr: Updated.

andreasgerstmayr:
Thanks for the review! Updated the PR with your changes.

What do you recommend about the lifetime of the `module_name` string?

natoscott:
Yeah, I think using the copy from the CACHE_CLUSTER_IDS indom would be good, avoiding duplication and refcounting complexity - we should be always able to look that up using the cluster_id if we need to...?

andreasgerstmayr:
> Yeah, I think using the copy from the CACHE_CLUSTER_IDS indom would be good, avoiding duplication and refcounting complexity - we should be always able to look that up using the cluster_id if we need to...?

+1
I've moved `pmdaCacheStore` a few lines up now & am using the copy of pmdaCache for the `module_name`.

src/pmdas/bpf/bpf.c
```diff
@@ -243,13 +255,14 @@ bpf_load_modules(dict *cfg)
         if (ret != 0) {
             libbpf_strerror(ret, errorstring, 1023);
             pmNotifyErr(LOG_ERR, "module initialisation failed: %s, %d, %s", module_name, ret, errorstring);
-            free(module_name);
+            ret = pmdaCacheStore(CACHE_CLUSTER_IDS, PMDA_CACHE_HIDE, module_name, bpf_module);
```
natoscott, line 258: I think you still need to free(module_name) here as pmdaCacheStore will create its own duplicate internally.  Otherwise all looks good!

natoscott (approved):

src/pmdas/bpf/bpf.c
andreasgerstmayr: I've free'd `module_name` in line 242, and in line 244 I'm reassigning it to the internal duplicate allocated in `pmdaCacheStore`. I've tried this error case with valgrind (running it as non-root-user triggers this condition), and it doesn't report any leaks.

PR #1566, opened 2022-03-22
Update INSTALL.md to add sudo su

saki-osive:
Update INSTALL.md to add sudo su to prevent install from failing if the user is not root and tries to install via the sudo privilege.


saki-osive:
@natoscott Hi, thank you for your reply, I have added the ``$`` before the privileged command as discussed.

Issue #1418, opened 2021-09-27
pmdabpftrace: symlink /var/lib/pcp/pmdas/bpftrace/autostart to /etc/pcp/bpftrace/autostart


andreasgerstmayr:
After the Fedora Silverblue change to not use `/var/lib/pcp` for config files, all files in `/var/lib/pcp` are now symlinks to `/etc/pcp`. The `PCP_PMDAS_DIR` config setting still points to `/var/lib/pcp/pmdas` however, so the bpftrace pmda is looking for autostart scripts at `/var/lib/pcp/pmdas/bpftrace/autostart` instead of `/etc/pcp/bpftrace/autostart`, which is confusing.

Suggest to create a symlink of the directory `/var/lib/pcp/pmdas/bpftrace/autostart` to `/etc/pcp/bpftrace/autostart`

- [ ] migrate existing files (except symlinks) in `/var/lib/pcp/pmdas/bpftrace/autostart` to `/etc/pcp/bpftrace/autostart`
- [ ] check for possible selinux issues

andreasgerstmayr:
> OTOH, I recall also our recent discussion about multi-level configs - possibly we should not be installing any files in /etc for these PMDAs ... (cos then we get into .rpmnew problems with near-dup config files in the same directory, overwriting local sysadmin changes, etc, etc)

I guess the right multi-level way to solve this would be to
* put RPM-supplied default scripts in `/usr/share/pcp/pmdas/bpftrace/autostart`
* put symlinks from `/etc/pcp/bpftrace/autostart` to `/usr/share/pcp/pmdas/bpftrace/autostart`
* the PMDA should read the `/etc/pcp/bpftrace/autostart` folder

But this way we still modify /etc/pcp, even though it's just symlinks.
The other solution would be to make the PMDA read both directories `/usr/share/pcp/pmdas/bpftrace/autostart` and `/etc/pcp/bpftrace/autostart`, however then the user cannot delete the RPM-supplied config anymore unless the user modifies /usr/share (which the user shouldn't do and also these files will be re-installed for every upgrade).

Another option would be to just not pre-install any scripts, and let the user decide which scripts to run. Might be the good solution, but will break existing installations.

Also the above ideas need some migration handling for current scripts in `/var/lib`.

[...]

PR #1568, opened 2022-03-30
pmdabpftrace: move example autostart scripts to /usr/share

andreasgerstmayr:
Resolves: #1418

Got it working, until I found this age-old bug in rpm (open since  2008) which prevents rpm from replacing a directory with a symlink during upgrades: https://bugzilla.redhat.com/show_bug.cgi?id=447156
There is a workaround involving lua code in the `%pretrans` scriptlet: https://docs.fedoraproject.org/en-US/packaging-guidelines/Directory_Replacement/

@natoscott: what do you recommend, should I go down this path?

Alternatively I could make the bpftrace PMDA look into both directories, and implement some logic to not load the same script twice. I'll still move the default scripts to `/usr/share/pcp/pmdas/bpftrace/example`, but I won't replace the `/var/lib/pcp/pmdas/bpftrace/autostart` directory with a symlink, due to this rpm bug.

Not sure what's the best, I'm not 100% happy with either.


natoscott:
> @natoscott: what do you recommend, should I go down this path?

I'd prefer not, it looks like a maintenance headache longer term.

> Alternatively I could make the bpftrace PMDA look into both directories

I think that's a good option.  Maybe it can be simpler still (since its a breaking v6 change already IIRC?) such that if the /etc path exists, we use that and stop looking further.  And in v6 we could stop installing the /var autostart location altogether, so that time people will have to migrate to the correct location and we don't end up with long-term path confusion for users.

andreasgerstmayr:
> since its a breaking v6 change already IIRC?

Partly, yes - I plan to move the default scripts to the `/usr/share/` example folder, so by default no script will be started when the PMDA is started. This way people who don't want these default scripts don't get irritated when they delete them and they re-appear again for every PCP update.

There is another part to backwards compatibility, namely people who already have custom scripts in their `/var` autostart location and people who follow already published guides like https://www.redhat.com/en/blog/visualizing-system-performance-rhel-8-part-3-kernel-metric-graphing-performance-co-pilot-grafana-and-bpftrace

andreasgerstmayr:
Updated PR to scan both directories. Duplicate scripts won't be started. This keeps backwards compatibility with user-provided scripts in `/var` and with guides mentioning this path.

natoscott (approved):
Makes sense, I think your most recent approach is the best so far.  Hadn't thought of the published guides, so yeah - it'd be good not to break that you're right.

PR #1575, opened 2022-04-13
Denki PMDA minor cleanups

natoscott:


christianhorn (changes requested):
Can we remove line 246 and change 247 as follows?
  pmNotifyErr(LOG_DEBUG, "Assuming %s%s is a battery we should provide metrics for.",dirname,ep->d_name);
That will give us that one line of detail which directory is considered a battery, good to have that by default.

christianhorn (approved):

christianhorn (changes requested):
Can we remove the conditional in line 664?
That would lead to one single announcement in log that (if) RAPL was found.

natoscott:
> Can we remove line 246 and change 247 as follows? pmNotifyErr(LOG_DEBUG, "Assuming %s%s is a battery we should provide metrics for.",dirname,ep->d_name); That will give us that one line of detail which directory is considered a battery, good to have that by default.

Will do (on both suggestions), thanks!  I'll also use LOG_INFO for these diagnostics rather than LOG_DEBUG.

christianhorn (approved):

PR #1576, opened 2022-04-18
pmda denki, multi battery support

christianhorn:
Indoms for batteries, allow multiple batteries. QA test for the multibat-system I have here.
Not sure if the debug option is implemented in the best way.


src/pmdas/denki/Install
```diff
@@ -26,6 +26,7 @@ dso_opt=false
 socket_opt=false
 pmns_source=root_denki
 
+# args="-U root -C appl0"
```
natoscott, line 29: See below for details, this should be using -D.

src/pmdas/denki/denki.c
```diff
@@ -412,20 +452,20 @@ static pmLongOptions longopts[] = {
     PMDA_OPTIONS_END
 };
 static pmdaOptions opts = {
-    .short_options = "D:d:i:l:r:pu:U:6:?",
+    .short_options = "C:D:d:i:l:r:pu:U:6:?",
```
natoscott, line 455: This -C option is not needed - the -D option already supports what you seek here.

src/pmdas/denki/denki.c
```diff
@@ -412,20 +452,20 @@ static pmLongOptions longopts[] = {
...
     .long_options = longopts,
 };
 
 /*
- * Our rapl readings are in a 2-dimensional array, map them here
+ * Our rapl readings are in a 2-dimension array, map them here
```
natoscott, line 460: "dimensional" was correct grammar there, not sure why this change?

src/pmdas/denki/denki.c
```diff
@@ -448,20 +488,23 @@ denki_fetchCallBack(pmdaMetric *mdesc, unsigned int inst, pmAtomValue *atom)
 
 	if (inst != PM_IN_NULL &&
 		!((cluster == 0 && item == 0) || 
-		(cluster == 0 && item == 1)))
+	 	  (cluster == 0 && item == 1) ||
+	 	  (cluster == 1 && item == 0) ||
+	 	  (cluster == 1 && item == 1) ||
+	 	  (cluster == 1 && item == 2)))
```
natoscott, line 494: I believe this could be expressed more compactly as:
```
        if (inst != PM_IN_NULL && mdesc->m_desc.indom == PM_INDOM_NULL)
                return PM_ERR_INST;
```

and this way if new metrics are added in future we don't need to keep updating the instance-checking logic here.

src/pmdas/denki/denki.c
```diff
@@ -594,21 +695,24 @@ denki_label(int ident, int type, pmLabelSet **lpp, pmdaExt *pmda)
...
 			}
 			break;
-		case PM_LABEL_ITEM:
-			if (pmID_cluster(ident) != 1)
-				break;
-			if (pmID_item(ident) == 0)	/* energy_now_raw */
-				pmdaAddLabels(lpp, "{\"units\":\"watt hours\"}");
-			if (pmID_item(ident) == 1)	/* energy_now_rate */
-				pmdaAddLabels(lpp, "{\"units\":\"watts\"}");
-			break;
```
natoscott, line 611: Hmm - this seems like useful units information that we're losing here - was it not correct?

src/pmdas/denki/denki.c
```diff
@@ -693,6 +798,9 @@ main(int argc, char **argv)
         		strncpy(rootpath, opts.optarg, sizeof(rootpath));
 			rootpath[sizeof(rootpath)-1] = '\0';
             		break;
+		case 'C':
+                	pmSetDebug("appl0");
+                	break;
```
natoscott, line 803: Can drop these three lines.  The -D option is the standard everywhere in PCP for enabling debugging diagnostics, and this agent already supports it automatically via pmdaGetOptions.

natoscott (changes requested):

src/pmdas/denki/denki.c
christianhorn: Right, was just taking over the rapl indoms here - but denki.bat has proper units.

natoscott (approved):
Looks good @christianhorn - I'll make one tiny change then merge after I do a QA run here.

The additional line in the pmLongOptions structure for -D is not necessary - this is provided 4 lines prior by the PMOPT_DEBUG macro.

PR #1577, opened 2022-04-22
Fix pmda-denki help to cover the new battery indoms

christianhorn:


christianhorn:
Looks like the centos-stream8 build error is not caused by my change, I guess building without the change would also fail..

PR #1586, opened 2022-05-04
Correct operation of `pidstat -p ALL|SELF`

portante:
Note that this PR is layered on PR #1585, where the `DRY out ...` commit lives.


portante:
The question is what to do with `pcp pidstat -p SELF` when using an archive?  I believe the command should report an error saying that `-p SELF` is not valid for that case.

natoscott:
@portante I think either an error or perhaps simply no output at all (just the headers) would be fine behaviours.

PR #1592, opened 2022-05-10
chore: fix pylint errors

andreasgerstmayr:
a couple of minor Python fixes:

* don't modify the dict while iterating over it
* variables should be declared before the try/except block; in the `except` block, don't access variables declared in the `try` block
* don't shadow the `error` variable when nesting try/except blocks


natoscott (approved):
LGTM, @myllynen might want to take a look too.

myllynen:
Thanks, LGTM!

PR #1596, opened 2022-05-13
pmdabpf: check libbpf version before including execsnoop

andreasgerstmayr:
and create symlinks for required source files only

### DIFF
--- a/src/pmdas/bpf/modules/GNUmakefile
+++ b/src/pmdas/bpf/modules/GNUmakefile
@@ -6,134 +6,31 @@ LDLIBS = $(PCP_WEBLIB) -lbpf -lelf -lz -lm -ldl
 MODULETMP = $(PCP_PMDAS_DIR)/$(IAM)/modules
 MODULEDIR = $(PCP_PMDASADM_DIR)/$(IAM)/modules
 
-MODULES	= \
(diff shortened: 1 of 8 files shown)


src/pmdas/bpf/modules/GNUmakefile
```diff
@@ -9,114 +9,28 @@ MODULEDIR = $(PCP_PMDASADM_DIR)/$(IAM)/modules
 MODULES	= \
 	biolatency.so \
 	runqlat.so    \
-	execsnoop.so  \
+
+# conditional enabled bpf PMDA modules
+## modules requiring libbpf >= 0.7.0
+ifeq ($(shell printf "%s\n" 0.7.0 `pkg-config --modversion libbpf` | sort --version-sort | head -n1),0.7.0)
```
natoscott, line 15: This sort of thing would usually be worked out by the configure script (see the Qt version checks, for an example) - its often handy to print out the version there too - and not by individual Makefiles during the build.

andreasgerstmayr: Good point, I've moved the logic to the configure script now and added additional output in case the execsnoop module is not included because the libbpf version is too old.

Issue #1597, opened 2022-05-16
denki support for battery charge remaining metric


natoscott:
( @chorn reminder issue, as requested )

```
<nathans> chorn, any chance you could add /sys/class/power_supply/BAT*/capacity as a metric to the denki PMDA?
<nathans> there's a dstat plugin from the original tool that we could provide then
<nathans> aiui that metric is 'percentage of remaining battery power'
<chorn> nathans, I guess denki is the best place for it, yes
```

PR #1598, opened 2022-05-16
Add capacity metric, testsuite

christianhorn:
Solves https://github.com/performancecopilot/pcp/issues/1597


natoscott (changes requested):
@christianhorn there seems to be an unrelated change here to do with bpftool - can you drop that please?  Otherwise looks good to me.

natoscott:
@christianhorn fixed and merged now, just as we discussed on IRC.

PR #1601, opened 2022-05-19
Changes to parse Interface stats

veda10:
Added code to parse OVS Interface stats

### DIFF
--- a/src/pmdas/openvswitch/pmdaopenvswitch.python
+++ b/src/pmdas/openvswitch/pmdaopenvswitch.python
@@ -23,9 +23,9 @@ import subprocess
 from pcp.pmda import PMDA, pmdaMetric, pmdaIndom, pmdaInstid
 from pcp.pmapi import pmUnits
 from pcp.pmapi import pmContext as PCP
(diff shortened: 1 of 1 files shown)


natoscott:
@veda10 hi, thanks, looks good.  I've verified qa/1680 still functions (but cannot see any simple way to extend this test, so have left it as-is).  There was an issue which pylint reported (use 'make check' to see it) - I've fixed that in a followup commit.

Oh, one suggestion for future work - please prefix your commit messages with the component that's changing (pmdaopenvswitch: in this case) - that helps maintainers down the track looking just at commit logs.  Thanks!

veda10:
> @veda10 hi, thanks, looks good. I've verified qa/1680 still functions (but cannot see any simple way to extend this test, so have left it as-is). There was an issue which pylint reported (use 'make check' to see it) - I've fixed that in a followup commit.
> @natoscott Thank you! 
> Oh, one suggestion for future work - please prefix your commit messages with the component that's changing (pmdaopenvswitch: in this case) - that helps maintainers down the track looking just at commit logs. Thanks!
Sure

Issue #1341, opened 2021-07-07
Prevent peak memory usage due to buffered Redis requests


andreasgerstmayr:
pmproxy is based on libuv, which works as depicted in this diagram: http://docs.libuv.org/en/v1.x/design.html#the-i-o-loop

When a filesystem change event happens, the discovery code starts and submits Redis requests - up to 66500 Redis requests after starting a single pmlogger. These requests are getting queued up until the next loop iteration. After the filesystem change callback is done, the libuv loop continues from the start and runs the Redis callbacks, which actually sends all Redis requests over the network (and processes any incoming Redis replies).

This pattern is problematic when there are multiple pmloggers, each writing to the disk at a similar time. All filesystem change events are run, i.e. thousands of Redis requests will be buffered (and occupying lots of memory) until they get sent in the next libuv loop iteration.

imho the problem is that the discovery code blocks the libuv loop, and doesn't let the Redis client process any requests or replies. This can be solved by either having a dedicated libuv loop (in a different thread), or partitioning the discovery function, so that not a single callback is blocking the loop, but that the discovery happens in parts.

I have some rough example code ready for the latter solution, i.e. splitting each loop iteration in `series_cache_update` in a new libuv timer callback (https://github.com/andreasgerstmayr/pcp/commit/27220a0bc2ee56eb63edd3421a28aabd74bfadb9). However, currently the discovery code is free'ing the `pmResult` after invoking `pmDiscoverInvokeValuesCallBack`, so running the code async will result in memory access of free'd memory (`process_logvol` in `discover.c`, and likely similar issues in other parts).

andreasgerstmayr:
Another (probably better) solution would be to offload the synchronous, blocking discovery code (calling into libpcp) into a worker thread. The Redis client functions will need to be called from the main thread however (they aren't threadsafe).

natoscott:
> Another (probably better) solution would be to offload the synchronous, blocking discovery code (calling into libpcp) into a worker thread.

My understanding is that this is what's supposed to be happening - pmproxy is multi-threaded due to the discovery code (the libuv code automatically creates threads for it under the covers, it's not something we explicitly do) - no other reason IIRC.  Not sure how that fits in with Redis requests not being sent from the main thread though.

goodwinos:
[...]

PR #1605, opened 2022-05-27
libpcp_web: break recursion and let libuv process the I/O queue when …

andreasgerstmayr:
…loading archives into Redis

Previously, the following recursion occured when loading archives into
Redis (e.g. using pmseries --load):

    [...]
    server_cache_window
    server_cache_update_done
    doneSeriesGetContext
    series_cache_update
    server_cache_window

The recursion does end when the archive was read entirely, however
the call stack could grew too large, resulting in a segmentation fault.

This commit breaks the recursion by scheduling server_cache_window() to
run on the next libuv loop iteration, after 5ms. 5ms was choosen so that
libuv can process pending (Redis) I/O callbacks.

This change also keeps the memory usage at about 130 MB, where
previously I saw RSS values of 10 GB+ due to queued Redis requests.

### DIFF
--- a/src/libpcp_web/src/load.c
+++ b/src/libpcp_web/src/load.c
@@ -553,8 +553,10 @@ server_cache_update_done(void *arg)
     context->count++;
     context->done = NULL;
 
-    /* begin processing of the next record if any */
-    server_cache_window(baton);
+    /* schedule processing of the next record (if any) after 5ms,
+     * to break the recursion and let libuv process pending Redis I/O callbacks
+     */
+    libuv_set_timeout(uv_default_loop(), server_cache_window, baton, 5);
 }
 
 void
--- a/src/libpcp_web/src/util.c
+++ b/src/libpcp_web/src/util.c
@@ -13,6 +13,7 @@
  */
 #include <assert.h>
 #include <ctype.h>
+#include <uv.h>
 #include "pmapi.h"
 #include "libpcp.h"
 #include "pmwebapi.h"
@@ -1389,4 +1390,48 @@ pmwebapi_nsectimestamp(sds s, struct timespec *timestamp)
     pmLocaltime(&now, &tmp);
     return sdscatfmt(s, "%02d:%02d:%02d.%09d",
 		tmp.tm_hour, tmp.tm_min, tmp.tm_sec, (int)timestamp->tv_nsec);
+}
+
+
+typedef struct libuv_set_timeout_data {
+    libuv_set_timeout_callback_t callback;
+    void *arg;
+} libuv_set_timeout_data;
+
+static void
+libuv_on_timer_close_complete(uv_handle_t *handle)
+{
+    free(handle);
+}
+
+static void
+libuv_set_timeout_invoke_callback(uv_timer_t *timer)
+{
+    libuv_set_timeout_data *timer_data = (libuv_set_timeout_data*)timer->data;
+
+    timer_data->callback(timer_data->arg);
+    free(timer_data);
+    uv_close((uv_handle_t*)timer, libuv_on_timer_close_complete);
+}
+
+/*
+ * Execute the given callback with an argument after X milliseconds
+ *
+ * This function is especially useful to defer some function invocation
+ * to the next libuv loop iteration (by setting timeout to 0).
+ */
+void
+libuv_set_timeout(uv_loop_t *loop, libuv_set_timeout_callback_t callback, void *arg, uint64_t timeout)
+{
+    uv_timer_t		*timer;
+    libuv_set_timeout_data *timer_data;
+
+    timer = calloc(1, sizeof(uv_timer_t));
+    timer_data = calloc(1, sizeof(libuv_set_timeout_data));
+    timer_data->callback = callback;
+    timer_data->arg = arg;
+    timer->data = timer_data;
+
+    uv_timer_init(loop, timer);
+    uv_timer_start(timer, libuv_set_timeout_invoke_callback, timeout, 0);
 }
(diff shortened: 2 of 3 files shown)


andreasgerstmayr:
I'm not very excited about the artificial `5ms` delay (it's not actually a delay, just an elaborate guess to let libuv process I/O callbacks).

Thinking more about it, I think a better way to fix the underlying problem is to wait until the Redis requests are processed before continuing with the next record.

Specifically, here: https://github.com/performancecopilot/pcp/blob/bb865a67fef74adfb0dbef9dc76366ad71bb517d/src/libpcp_web/src/load.c#L500-L507

`server_cache_metric()` initiates write requests to Redis, but is not waiting for them to complete. Instead, the loop continues and after the loop `doneSeriesGetContext(context, "series_cache_update");` is called, which calls `server_cache_update_done` which then calls `server_cache_window` to process the next record (if any).

I think `doneSeriesGetContext(context, "series_cache_update");` should only be called after *all* `server_cache_metric()` write requests are completed. Then the recursion is also broken up and we won't get segmentation faults because the call stack got too big anymore. As a bonus, afaics this will also greatly improve memory usage when pmproxy is run in discovery mode, with many archives (less inflight Redis requests = less memory usage). Related: #1341

natoscott:
Ah, this is great progress @andreasgerstmayr - nicely done.  Instead of a timer-based approach I had been thinking more along the lines of size-limiting things, i.e. once we reach a certain size pending (10MB?  100?), flush it and carry on.  But timer-based certainly has its merits too (may have additional benefits around interactive performance in pmproxy).

| I think a better way to fix the underlying problem is to wait until the Redis requests are processed before continuing with the next record.

Maybe, worth experimenting with - and possibly a combination of both might be ideal.  One thing we want to avoid would be creating a ping-pong with the redis-server process, such that we send-then-wait, redis-waits-then-replies (i.e. synchronous round trips), and so on - best-case I think may be if we can setup a pipeline where we get a bit ahead and feed data at a constant rate.  Recall of course that we need to do I/O on our end too (esp. reading the data volume) so we don't want to block on that if we can avoid it.

Next step would be to start getting some numbers I think - as well as the memory numbers you have, also see if/how much this initial change speeds up archive import... and then improve incrementally from there.

natoscott:
One other thought, I notice there is a non-blocking event-loop mechanism in libuv:

```
uv_run(loop, UV_RUN_NOWAIT);
```

I'm not sure about this, but we might be able to call that (with the existing event loop as first parameter) at convenient times rather than introducing an additional libuv timer.

natoscott:
@andreasgerstmayr another factor that occurred to be is we should consider

```
$ ulimit -s
8192
```

here, and ensure we stop recursing before we reach anywhere near that size.  Tricky to calculate exactly how much each call adds to the stack, but we could estimate some conservative amount plus fudge factor for existing stack use, divide by the above value and ensure we make fewer calls than that before flushing.

andreasgerstmayr:
>  i.e. once we reach a certain size pending (10MB? 100?), flush it and carry on

Is there a libuv function to flush the queue manually (I couldn't find any, the timer is more or less a workaround to flush the queue by giving libuv time to process it, and also to break the recursion)?

> Maybe, worth experimenting with - and possibly a combination of both might be ideal. One thing we want to avoid would be creating a ping-pong with the redis-server process, such that we send-then-wait, redis-waits-then-replies (i.e. synchronous round trips), and so on - best-case I think may be if we can setup a pipeline where we get a bit ahead and feed data at a constant rate. Recall of course that we need to do I/O on our end too (esp. reading the data volume) so we don't want to block on that if we can avoid it.

+1
I didn't plan to wait for a reply from Redis, but wait until the request got sent out to Redis. Afaics currently each loop iteration adds requests to the Redis send queue/backlog at a rate faster than it is processing this queue (sending out the requests over the network). This results in the high memory usage.

> Next step would be to start getting some numbers I think - as well as the memory numbers you have, also see if/how much this initial change speeds up archive import... and then improve incrementally from there.

Yep. To be honest, I'm not 100% sure if my conclusions above are correct. It's a tricky issue, and fixing this once and for all would also help with the pmlogger farm memory usage.

> `uv_run(loop, UV_RUN_NOWAIT);`

I think I experimented with that last year (looked like an easy way to drain the libuv queue), however the docs mention "It must not be called from a callback." (http://docs.libuv.org/en/v1.x/loop.html#c.uv_run)

> here, and ensure we stop recursing before we reach anywhere near that size

I think there should be a way to replace the recursion with an iteration. Checking the max stack size and calculating some fudge factor could lead to potential future problems imho (on different platforms, compilers etc).

natoscott:
> Is there a libuv function to flush the queue manually (I couldn't find any, the timer is more or less a workaround to flush the queue by giving libuv time to process it, and also to break the recursion)?

Not that I know of - uv_run was the closest thing I found.

> 
> Yep. To be honest, I'm not 100% sure if my conclusions above are correct. It's a tricky issue, and fixing this once and for all would also help with the pmlogger farm memory usage.
> 

Your conclusions make sense to me, I think you're on the right track here.

> > `uv_run(loop, UV_RUN_NOWAIT);`
> 
> iirc I experimented with that last year (looked like an easy way to drain the libuv queue), however the docs mention "It must not be called from a callback." (http://docs.libuv.org/en/v1.x/loop.html#c.uv_run)

pmproxy uses uv_callback.c to dodge that bullet IIRC, perhaps it could move into libpcp_web.

> > here, and ensure we stop recursing before we reach anywhere near that size
> 
> I think there should be a way to replace the recursion with an iteration.

It's a tough one - I've not been able to come up with an approach so far, but agree this would be an excellent option if we can find a way.

cheers.

andreasgerstmayr:
Superseeded by #1634.

> Recall of course that we need to do I/O on our end too (esp. reading the data volume) so we don't want to block on that if we can avoid it.

That was the key, I don't know how I missed that earlier. I guess I thought we're already using a worker thread when interacting with the (synchronous/blocking) PCP API.

andreasgerstmayr:
Obsolete by #1634.

PR #1607, opened 2022-06-01
pmdabpf: check libbpf version before running any qa

smalinux:
Fixes: 2d9476cb0 ("qa/1931: pmdabpf: execsnoop")

### DIFF
--- a/qa/1931
+++ b/qa/1931
@@ -12,6 +12,7 @@ echo "QA output created by $seq"
 
 _pmdabpf_check
 _pmdabpf_require_kernel_version 5 0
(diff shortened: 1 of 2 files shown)


natoscott (changes requested):
@smalinux good progress, thanks.  I think the addition of a new python program here is overkill though - can you update it to to use the same approach that _pmdabpf_require_kernel_version uses please?  Also "_more_or_equal" part is implied so you can drop that from the name - again, see _pmdabpf_require_kernel_version naming conventions and the '+' in its diagnostic message.

PR #1610, opened 2022-06-15
libpcp_import: support for an archive-append mode

natoscott:
In order to implement sysstat/sar semantics, where an
archive is appended to by short-lived processes - and
not by a long-running process ala pmlogger - we're in
need of an append-mode from the LOGIMPORT(3) API.

This is a prototype of the sorts of changes I believe
we'll need.  I've updated just the one QA test utility
to exercise it at this stage and the rest will follow
once we've settled on appropriate APIs.

### DIFF
--- a/qa/src/check_import.c
+++ b/qa/src/check_import.c
@@ -19,52 +19,12 @@ check(int sts, char *name)
     }
 }
 
-int
-main(int argc, char **argv)
+void
+test_basics(int Vflag, int version)
 {
     int		sts;
     int		ctx1;
-    int		ctx2;
     int		hdl1;
-    int		hdl2;
-    int		errflag = 0;
-    int		c;
-    int		Vflag = 0;
-    int		version;
-    static char	*usage = "[-D debugspec] [-V version]";
-
-    pmSetProgname(argv[0]);
-
-    while ((c = getopt(argc, argv, "D:V:")) != EOF) {
-	switch (c) {
-
-	case 'D':	/* debug options */
-	    sts = pmSetDebug(optarg);
-	    if (sts < 0) {
-		fprintf(stderr, "%s: unrecognized debug options specification (%s)\n",
-		    pmGetProgname(), optarg);
-		errflag++;
-	    }
-	    break;
-
-	case 'V':	/* output archive version */
-	    version = atoi(optarg);
-	    Vflag++;
-	    break;
-
-	case '?':
-	default:
-	    errflag++;
-	    break;
-	}
-    }
-
-    if (errflag) {
-	printf("Usage: %s %s\n", pmGetProgname(), usage);
-	exit(1);
-    }
-
-    pmiDump();
 
     ctx1 = pmiStart("myarchive", 0);
     check(ctx1, "pmiStart");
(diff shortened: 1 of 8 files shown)


kmcdonell:
@natoscott Basically LGTM.

My only questions (that I can't resolve from the diffs) are:

1. Is the state of the auxiliary data (and in particular the hashed indom data) is the **same** in the open for append case as it would be in the "just keep writing" case?  At a first stab this probably needs QA coverage, especially with a V3 archive, a dynamic indom and delta-indom records in play).
2. There may need to be a <mark> record added, but the logic for this is complicated (consider what pmlogger has to do when a pmcd connection is re-established, or the tests pmlogextract does on pmcd.seqnum).  It is unclear to me if this should be part of the libpcp functionality or something the caller needs to take care of ... even in the sysstat case one needs to decide how kernel counter resets to zero on reboot will be accommodated.

kmcdonell:
I see some of my issues above have been addressed in the first commit (I was only looking at the second commit) ... proper review will be done shortly.

qa/src/check_import.c
```diff
@@ -349,5 +317,87 @@ main(int argc, char **argv)
...
+void
+test_append(void)
+{
+    int		sts;
+    int		ctx3;
+    int		hdl3;
+
+    ctx3 = pmiStart("myarchive", PMI_FLAG_APPEND);
+    check(ctx3, "pmiStart");
+    sts = pmiAddMetric("my.metric.bar", pmID_build(245, 0, 2), PM_TYPE_U64, pmInDom_build(245,1), PM_SEM_INSTANT, pmiUnits(1,-1,0,PM_SPACE_MBYTE,PM_TIME_SEC,0));
```
kmcdonell, line 333: Why is this pmiAddMetric() needed?  my.metric.bar should have bee

kmcdonell (commented):
LGTM.
QA will need some work.
Q on semantics in comments for qa/src/check_import.c
libpcp_import does not (yet) support delta indoms for V3 archives ... this may be now even trickier, or may be we don't do it!

qa/src/check_import.c
natoscott: The model I'm catering a need for is the sysstat/sadc scenario, where we don't (want to) know whether we are appending or creating - whichever case it is, it should Just Work and do the Right Thing, transparently and correctly.  IOW, the application using libpcp_import doesn't want to know about this subtelty, and the correct behavior should just magically happen.

If the archive doesn't exist create it and populate all metric metadata.  If it does exist, check the metric metadata still matches (in-memory vs on-disk - fail if not) and append to the existing archive.

natoscott:
> LGTM. QA will need some work. Q on semantics in comments for qa/src/check_import.c libpcp_import does not (yet) support delta indoms for V3 archives ... this may be now even trickier, or may be we don't do it!

Yep, QA definitely needs to be fleshed out further.  Maybe in some cases V3 archives could be easier for deltas - we do now know when appending and needing to add instance metadata.  Certainly this could be improved on later anyway, I think in the default case for sysstat we wont be logging fast-changing indoms.

PR #1611, opened 2022-06-15
Time domain operation for max

ShiyaoC:
pmseries: implemented max operation


src/libpcp_web/src/query_parser.y
```diff
@@ -960,7 +981,8 @@ static const struct {
 } func[] = {
     { L_AVG,		sizeof("avg")-1,	"avg" },
     { L_COUNT,		sizeof("count")-1,	"count" },
-    { L_MAX,		sizeof("max")-1,	"max" },
```
natoscott, line 963: I think we should keep "max" since its part of the language now and people are using it (so if we remove it, things will break) - we can make it a synonym of "max_inst" though.

src/libpcp_web/src/query.c
```diff
@@ -2623,6 +2627,69 @@ series_calculate_rate(node_t *np)
...
+
+    n_series = np->left->value_set.num_series;
+    np->value_set.num_series = n_series;
+    np->value_set.series_values = (series_sample_set_t *)calloc(n_series, sizeof(series_sample_set_t));
+    for (i = 0; i < n_series; i++) {
+		n_samples = np->left->value_set.series_values[i].num_samples;
+		if (n_samples > 0){
+			np->value_set.series_values[i].num_samples = n_samples;
+			np->value_set.series_values[i].series_sample = (series_instance_set_t *)calloc(n_samples, sizeof(series_instance_set_t));
+			n_instances = np->left->value_set.series_values[i].series_sample[0].num_instances;
```
natoscott, line 2650: Did you find the cause of the memory allocation / valgrind issue from the other day?  (was it n_samples == 0?)

src/libpcp_web/src/query.c
```diff
@@ -2623,6 +2627,69 @@ series_calculate_rate(node_t *np)
...
+				}
+				
+				np->value_set.series_values[i].series_sample[j].series_instance[0].timestamp = 
+					sdsnew(np->left->value_set.series_values[i].series_sample[j].series_instance[max_pointer].timestamp);
+				np->value_set.series_values[i].series_sample[j].series_instance[0].series = 
+					sdsnew(np->left->value_set.series_values[i].series_sample[j].series_instance[max_pointer].series);
+				np->value_set.series_values[i].series_sample[j].series_instance[0].data = 
+					sdsnew(np->left->value_set.series_values[i].series_sample[j].series_instance[max_pointer].data);
+				np->value_set.series_values[i].series_sample[j].series_instance[0].ts = 
+					np->left->value_set.series_values[i].series_sample[j].series_instance[max_pointer].ts;
```
natoscott, line 2680: There's a code pattern I've noticed developing throughout query.c which I'm not sure is ideal, where we have these very long statements repeatedly de-referencing pointers (I know you're just continuing the same pattern here & usually that's the best approach to take).  I wonder if it would make the code more readable, however, if we used some intermediate local variables.

For example here we repeat the: "np->value_set.series_values[i].series_sample[j].series_instance[0]" string several times, as well as "np->left->value_set.series_values[i].series_sample[j].series_instance[max_pointer]".  Shall we start moving code over to assigning earlier on to a local variable (e.g. "inst", and then using "inst->timestamp = ...; inst->series = ...; " and so on) ... ?

natoscott (changes requested):
@ShiyaoC looks good - I'll chat to you tomorrow about that failing test case - we'll need to extend that as part of this change to add regression testing.  The missing piece of the puzzle here is an update to the man page (man/man1/pmseries.1) - we'll need to add a description of these new functions.

src/libpcp_web/src/query.c
ShiyaoC: Thank you for your suggestions, and I agree with you that defining some local variables like "inst" will improve the readability of our code. I will update this.

natoscott (commented):
@ShiyaoC we still need a QA test update here I think?

ShiyaoC:
QA test updated

PR #1624, opened 2022-07-13
packaging: support dynamic systemd allocation of system uid/gid

natoscott:
Resolves Red Hat BZ #2095445

### DIFF
--- a/build/rpm/GNUmakefile
+++ b/build/rpm/GNUmakefile
@@ -30,7 +30,7 @@ default install install_pcp default_pcp :
 # going to make sure it'll be rebuild next time we come here
 # We used to override --target=FOO, but it's better to let the
 # local build distro defaults do their jobs.
-pack_pcp : rpmmacros $(SPEC)
+pack_pcp : rpmmacros systemd-sysuser-configs $(SPEC)
 	DEFS=`grep '^--define' rpmmacros`; \
 	eval $(RPMPROG) -ba $$DEFS --clean $(SPEC)
 	@if [ -n "$$RPM_SIGN_KEYID" ]; then \
@@ -122,6 +122,10 @@ pcp.spec: pcp.spec.in
 	    -e's|@have_manpages@|$(HAVE_MANPAGES)|g' \
 	    -e's|@make@|$(PCP_MAKE_PROG)|g' <$? >$@
 	./do-maybe-requires
+
+systemd-sysuser-configs:
+	$(MAKE) $(MAKEOPTS) -C $(TOPDIR)/src/include pcp.user.install
+	$(MAKE) $(MAKEOPTS) -C $(TOPDIR)/qa pcpqa.user
 
 include $(BUILDRULES)
 
--- a/build/rpm/GNUmakefile
+++ b/build/rpm/GNUmakefile
@@ -30,7 +30,7 @@ default install install_pcp default_pcp :
 # going to make sure it'll be rebuild next time we come here
 # We used to override --target=FOO, but it's better to let the
 # local build distro defaults do their jobs.
-pack_pcp : rpmmacros systemd-sysuser-configs $(SPEC)
+pack_pcp : rpmmacros $(SPEC)
 	DEFS=`grep '^--define' rpmmacros`; \
 	eval $(RPMPROG) -ba $$DEFS --clean $(SPEC)
 	@if [ -n "$$RPM_SIGN_KEYID" ]; then \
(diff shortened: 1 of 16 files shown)


andreasgerstmayr:
lgtm, let's merge this for v6?

andreasgerstmayr:
```
$ build/ci/ci-run.py fedora36-container reproduce

[pcpbuild@2853676f57bc pcp]$ sudo -i -u pcpqa ./check 002
Installing "localconfig"
./mk.localconfig: line 23: src/localconfig.h: Permission denied
mv: cannot move '/var/tmp/315610.out' to 'localconfig': Permission denied
./common.check: line 14: ./localconfig: No such file or directory

[pcpbuild@2853676f57bc pcp]$ stat /var/lib/pcp/testsuite/src
  File: /var/lib/pcp/testsuite/src
  Size: 7294      	Blocks: 0          IO Block: 4096   directory
Device: 0,137	Inode: 84407900    Links: 1
Access: (0755/drwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
```

On my system this folder is owned by `pcpqa`:
```
$ stat /var/lib/pcp/testsuite/src
  File: /var/lib/pcp/testsuite/src
  Size: 7294      	Blocks: 0          IO Block: 4096   directory
Device: 0,36	Inode: 6916406     Links: 1
Access: (0755/drwxr-xr-x)  Uid: (  385/   pcpqa)   Gid: (  385/   pcpqa)
```

natoscott:
> I'll investigate.

Thanks - yeah, it looks like the useradd is not happening correctly but I've not had a chance to look closer yet.

andreasgerstmayr:
My last commit should fix the issue for `pcp.spec.in`. I'm not sure how to fix it for `redhat.spec` though, because in that case we can't run any script prior to the build.

build/rpm/GNUmakefile
```diff
@@ -123,5 +123,9 @@ pcp.spec: pcp.spec.in
 	    -e's|@make@|$(PCP_MAKE_PROG)|g' <$? >$@
 	./do-maybe-requires
 
+systemd-sysuser-configs:
+	$(MAKE) $(MAKEOPTS) -C $(TOPDIR)/src/include pcp.user.install
+	$(MAKE) $(MAKEOPTS) -C $(TOPDIR)/qa pcpqa.user
+
```
natoscott, line 129: Hmm, I don't follow ... these files have already been built elsewhere in the source tree haven't they?  (I'm probably missing something, but if not - why do this here as well?)

natoscott (commented):
Ohh!  I understand what I got wrong now - thanks @andreasgerstmayr ... so to solve this for both styles of rpm build I think we'll need to do something more like the way debian/control is handled.  This has the exact same requirement in that it must be available before the build start.  IOW, we would have a git-committed 'control' file equivalent (for the redhat.spec case) and a control.master file equivalent (for the pcp.spec.in case).  Then there's a debian/fixcontrol script - here's we'd need a fixsysuser or something along those lines.

andreasgerstmayr:
@natoscott: My current understanding of the Makepkgs process is:
`Makepkgs` --> `./configure` --> `make -C build pack_pcp` --> generate `pcp.spec` from `pcp.spec.in` using the variables found by the configure script --> run `rpmbuild` (which again runs configure, as part of the build process). I've updated the requisites of the `pack_pcp` command to generate the systemd-sysuser config files before running `rpmbuild`.

For the `redhat.spec` file, afaics this file just gets synced to Fedora pagure. With the Fedora build process, we can't run any script in the Koji builders before `rpmbuild` (well, we could run a script locally on our machine before `fedpkg build`. But that won't be executed for automated rebuilds triggered by Fedora infra, and the only replacements are the username and the home directory of the user, which can be hardcoded anyway for the Fedora RPMs).

Re Debian: Oh, yes, I forgot about the debian builds. I found https://manpages.debian.org/testing/debhelper/dh_installsysusers.1.en.html, afaics it expects the sysuser config files in `debian/package.sysusers`. I'm not familiar with the debian build process, do the debian build machines run the Makefile? The Makepkgs script runs `make -C debian control` before starting `dpkg-buildpackage`.

andreasgerstmayr:
`debhelper` supports `dh_installsysusers` since version 13.3, i.e. Debian >= bullseye and Ubuntu >= 21.10, so this needs to be conditional. Currently we do support Debian buster (oldstable) and multiple Ubuntu LTS releases.

What do you think, does it make sense to go down this road and implement it conditional for Debian? imho it seems like it is a bit much effort and increase in complexity of the build process for little gain.

natoscott:
Ignore that last comment (re deb builds) - I think we're fine without it since we install the files anyway.  I'll push something tomorrow after some more testing.

PR #1628, opened 2022-07-15
libpcp_web: solve memory leak

ShiyaoC:
The previous freeSeriesQueryNode() function did not free p->value_set.series_values[i].series_desc.*, so changing the values of series_desc.type or series_desc.semantics would cause memory leaks.

The updated code assigned a sdsnew() value to each of p->value_set.series_values[i].series_desc, and free them in freeSeriesQueryNode().


natoscott (commented):
Could you create a new QA test for these @ShiyaoC ?
The process is described in qa/README but basically involves running "cd qa; ./new" ... there's lots of other tests using valgrind you could base a new one on

Issue #1606, opened 2022-05-27
pmseries --load should have an option to ignore metrics/indoms


andreasgerstmayr:
For pmproxy discovery, we exclude
```
# comma-separated metrics name (globs) to skip during discovery
exclude.metrics = proc.*,acct.*

# comma-separated list of instance domains to skip during discovery
exclude.indoms = 3.9,3.40,79.7
```
by default.

When you load an archive with `pmseries --load` into Redis, these settings are not respected and afaics there is currently no way to ignore these metrics/indoms while importing.

I'm not sure what's the best way; maybe read `pmproxy.conf` and use these settings (however, they are under the `[discover]` section), and add a new CLI argument for the case when I explicitly do not want to exclude these metrics?
Excluding these metrics will of course considerably speed up the import process.

natoscott:
FWIW pmseries already reads pmproxy.conf via the pmseries.conf symlink - I vote we just honour that discovery setting in pmseries too.

andreasgerstmayr:
> FWIW pmseries already reads pmproxy.conf via the pmseries.conf symlink - I vote we just honour that discovery setting in pmseries too.

+1

PR #1632, opened 2022-07-27
libpcp_web: honor exclude.metrics setting in pmseries --load

andreasgerstmayr:
Resolves: #1606


natoscott (commented):
@andreasgerstmayr LGTM

PR #1634, opened 2022-07-29
libpcp_web: run pmFetchArchive() in a worker thread

andreasgerstmayr:
Run pmFetchArchive() in a worker thread instead of the main thread because it uses blocking (synchronous) I/O calls.

This way the libuv loop does not get blocked, and queued Redis requests can be sent out while the worker thread is reading PCP archives. As a side effect, this also breaks the recursion as mentioned in #1605.

I think this is a better solution than #1605 for the same problem. It wasn't actually much new code, mainly moving existing code around. On my system, with pcp-zeroconf (10s logging interval) it takes about 1:30 minutes to load one archive (i.e. one day of metrics) into Redis (ignoring the proc.* metrics, as done in #1632). Without this commit, it segfaults due to the recursion, as mentioned in #1605 (it does work with archives in the `qa` directory though, afaics they are too small to reproduce the issue).

Overall I think about 1:30 minutes for importing one day of metrics is still a bit much (around 10 minutes for a week of metrics), but it's a start.

fyi: The diff looks a bit convoluted, basically I splitted this code from `server_cache_window()`: https://github.com/performancecopilot/pcp/blob/362ca7bb68bbaddba3d76ddeaca515e87554fd0a/src/libpcp_web/src/load.c#L580-L602 and moved `pmFetchArchive()` to `fetch_archive()` (running on the worker thread) and the conditional to `fetch_archive_done` (running on the main thread). Because `pmFetchArchive()` uses a thread-local variable (the context), I had to create a new function `pmFetchArchiveCtx` which accepts a context handle.

### DIFF
--- a/src/include/pcp/pmapi.h
+++ b/src/include/pcp/pmapi.h
@@ -559,6 +559,7 @@ PCP_CALL extern int pmHighResFetch(int, pmID *, pmHighResResult **);
  * Variant that is used to return a result from an archive.
  */
 PCP_CALL extern int pmFetchArchive(pmResult **);
+PCP_CALL extern int pmFetchArchiveCtx(int, pmResult **);
 PCP_CALL extern int pmFetchHighResArchive(pmHighResResult **);
 
 /*
--- a/src/libpcp/src/exports.in
+++ b/src/libpcp/src/exports.in
@@ -842,4 +842,5 @@ PCP_3.37 {
     __pmFreeLogInDom;
     __pmLogEncodeLabels;
     __pmDumpStackInit;
+    pmFetchArchiveCtx;
 } PCP_3.36;
--- a/src/libpcp/src/fetch.c
+++ b/src/libpcp/src/fetch.c
@@ -307,37 +307,47 @@ __pmFetchArchive(__pmContext *ctxp, __pmResult **result)
 {
     int		sts;
 
-    if ((sts = pmWhichContext()) >= 0) {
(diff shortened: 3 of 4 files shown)


natoscott:
@andreasgerstmayr nice!

> I had to create a new function pmFetchArchiveCtx which accepts a context handle.

We can more simply call pmUseContext in the thread, and not require this function I think.  This is the usual PMAPI model and I'd rather not stray from that now.  If we did, the "Ctx" convention is used internally in libpcp for a slightly different purpose, so we'd have needed to use "Context" here I think - and we'd need man page updates and a high-res variant.  Simpler to not change libpcp in this way - I've updated the code accordingly.

andreasgerstmayr:
> @andreasgerstmayr nice!
> 
> > I had to create a new function pmFetchArchiveCtx which accepts a context handle.
> 
> We can more simply call pmUseContext in the thread, and not require this function I think. This is the usual PMAPI model and I'd rather not stray from that now. If we did, the "Ctx" convention is used internally in libpcp for a slightly different purpose, so we'd have needed to use "Context" here I think - and we'd need man page updates and a high-res variant. Simpler to not change libpcp in this way - I've updated the code accordingly.

+1
Thanks!

For long-term plans, I think it would make sense to provide additional API functions which don't make use of thread-local state (especially looking at async programming, thread pools and when interacting from a different language with the pmapi).

PR #1637, opened 2022-08-02
Pmseries nth percentile

ShiyaoC:
libpcp_web: add nth_percentile_inst() and nth_percentile_sample() functions


src/libpcp_web/src/query.c
```diff
@@ -3290,6 +3304,170 @@ series_calculate_standard_deviation(node_t *np)
...
+			n_pointer[l] = j;
+			break;
+			}
+		    }
+		}
+		if (rank == n_samples){
+		    instance_idx = n_pointer[0];
+		}
+                else{
+		    instance_idx = n_pointer[n_samples-1-rank];
```
natoscott, line 3446: I observed this line crash with some large (possibly invalid) input numbers - if 'rank' is greater than 'n_samples' this code will attempt to access memory outside of the n_pointer array.

src/libpcp_web/src/query_parser.y
```diff
@@ -524,6 +526,50 @@ func_sid
 		  lp->yy_np->left = $3;
 		  $$ = lp->yy_series.expr = lp->yy_np;
 		}
+	| L_NTH_PERCENTILE_INST L_LPAREN sid_vec L_COMMA L_INTEGER L_RPAREN
+		{ lp->yy_np = newnode(N_NTH_PERCENTILE_INST);
+		  lp->yy_np->left = $3;
+		  if (atoi($5) < 0 || atoi($5) > 100){
```
natoscott, line 532: Instead of atoi, let's use strtoul here - that way, we can do some error checking.
Also, more efficient if we call it just once and save the result in a temporary variable rather than doing the string conversion twice.

src/libpcp_web/src/query_parser.y
```diff
@@ -524,6 +526,50 @@ func_sid
 		  lp->yy_np->left = $3;
 		  $$ = lp->yy_series.expr = lp->yy_np;
 		}
+	| L_NTH_PERCENTILE_INST L_LPAREN sid_vec L_COMMA L_INTEGER L_RPAREN
```
natoscott, line 529: Should the L_INTEGER in this line be "integer" like in your topk branch?  (similar case for lines 540, 551, 562 below)

qa/1886
```diff
@@ -243,6 +243,30 @@ echo Compared to stdev:
...
+pmseries $args 'nth_percentile_sample(kernel.all.load{instance.name == "5 minute"}[count:5], 50)'
+
+echo;echo "== Verify nth_percentile() functions for a singular metric"
+echo Original with no function:
+pmseries $args 'kernel.all.uptime[count:10]'
+echo Compared to nth_percentile_inst:
+pmseries $args 'nth_percentile_inst(kernel.all.uptime[count:10], 50)'
+echo Compared to nth_percentile_sample:
+pmseries $args 'nth_percentile_sample(kernel.all.uptime[count:10], 50)'
+
```
natoscott, line 269: @ShiyaoC this test could be improved by also checking error conditions in the input statements - i.e. making sure that we handle bad input for the 'nth_percentile_sample' expression - e.g. lacking parameters, has too many parameters, non-integer second parameter, too-large second parameter, etc.  The error handling should always produce consistent output, so its a good thing to add to the test to make sure we always handle it.

src/libpcp_web/src/query.c
```diff
@@ -2361,6 +2361,14 @@ series_expr_canonical(node_t *np, int idx)
...
     case N_STDEV_INST:
     case N_STDEV_SAMPLE:
+    case N_NTH_PERCENTILE_INST:
+    	left = series_expr_canonical(np->left, idx);
+	right = series_expr_canonical(np->right, idx);
+	break;
+    case N_NTH_PERCENTILE_SAMPLE:
+    	left = series_expr_canonical(np->left, idx);
+	right = series_expr_canonical(np->right, idx);
+	break;
```
natoscott, line 2371: the above section can be compressed (follow the switch/case pattern there - we only need to do the left/right assignments once).

natoscott (changes requested):
@ShiyaoC looks good - a few small possible improvements for your consideration here.

src/libpcp_web/src/query.c
ShiyaoC: rank should not be greater than the n_sample, since the formula to calculate rank is (int)(n/100 * n_samples), where n is the percentile value. For example if n_samples = 10, n = 90, then rank = 9, which is the largest value. So the range will be 0 to n_samples.

src/libpcp_web/src/query_parser.y
ShiyaoC: It gets updated.

qa/1886
ShiyaoC: I will add more tests for edge cases.

PR #1673, opened 2022-09-07
fix: add linux bpftool to allow bpf to be packaged on ubuntu2204

### LINKED ISSUE #1672: ubuntu22.04 - bpf pmda not packaged
It looks like `bpftool` is required on the build stack, and I had missed this earlier.

This is available by `apt install linux-tools-common linux-tools-5.15.0-47-generic`.

```
checking clang major version installed... 14
checking for libbpf... yes
checking libbpf version... 0.5.0
checking for libelf... yes
checking for clang... /usr/bin/clang
checking for llvm-strip... /usr/bin/llvm-strip
checking for bpftool... no                                <<------
checking BPF CO-RE architecture identifier... x86
checking for vmlinux.h... /home/pcpbuild/pcp/build/deb/pcp-6.0.1/vendor/github.com/iovisor/bcc/libbpf-tools/x86/vmlinux.h
checking if the bpf PMDA should be included... no
checking if the bpftrace PMDA should be included... yes
```

A diff such as this would suffice, however _only_ if the host operating system kernel version has a matching linux-tools build for ubuntu:
```
diff --git a/qa/admin/package-lists/Ubuntu+22.04+x86_64 b/qa/admin/package-lists/Ubuntu+22.04+x86_64
index d3cad9d55..91f87ce48 100644
--- a/qa/admin/package-lists/Ubuntu+22.04+x86_64
+++ b/qa/admin/package-lists/Ubuntu+22.04+x86_64
@@ -140,4 +140,7 @@ clang
 llvm-14
 llvm
 fakeroot
+linux-tools-common
+linux-tools-generic
 ```

We have the `--with-pmdabpf-bpftool  path to bpftool (default is empty for auto` configure flag that would allow us to specify the correct directory.

### PR DESCRIPTION
**Pull request by jasonk000**
Also move the Github build workflows to run on ubuntu22 as the base since the kernel version that linux-tools-xx will install needs to have a valid package, and bpftool will use it.

There is a configure option to pass --with-pmdabpf-bpftool as an alternative.

Hopefully fixes #1672.

### CODE REVIEW

**Comment by jasonk000** [comment]
As expected, this fails due to a kernel version mismatch between the underlying infrastructure and what's requested in the CI scripts and docker.

```
WARNING: bpftool not found for kernel 5.15.0-1017

  You may need to install the following packages for this specific kernel:
    linux-tools-5.15.0-1017-azure
    linux-cloud-tools-5.15.0-1017-azure

  You may also want to install one of the following packages to keep up to date:
    linux-tools-azure
    linux-cloud-tools-azure
```

I'll do some digging.

jasonk000:
I believe there are three possible solutions here:
1. Update the CI environment to depend on `linux-tools-5.15.0-1017-azure`, which is available on Ubuntu22. Pro: works. Con: depends on underlying infrastructure version staying the same (it probably will not).
2. Package and upload `bpftool` as a binary; this is the strategy iovisor project used in the past. Pros: simple. Con: needs to be maintained.
3. Vendor in or submodule in `bpftool` and build it as part of the build; this is the current iovisor strategy. Pros: clean and works. Con: adds another thing to build process.

Personally I think (3) is probably the right approach here. Thoughts?

@andreasgerstmayr @kmcdonell @natoscott

andreasgerstmayr:
> I believe there are three possible solutions here:
> 
>     1. Update the CI environment to depend on `linux-tools-5.15.0-1017-azure`, which is available on Ubuntu22. Pro: works. Con: depends on underlying infrastructure version staying the same (it probably will not).
> 
>     2. Package and upload `bpftool` as a binary; this is the strategy iovisor project used in the past. Pros: simple. Con: needs to be maintained.
> 
>     3. Vendor in or submodule in `bpftool` and build it as part of the build; this is the current iovisor strategy. Pros: clean and works. Con: adds another thing to build process.
> 
> 
> Personally I think (3) is probably the right approach here. Thoughts?

I think (3) is our only option, as (1) will be breaking all the time and (2) requires a binary blob to generate parts of our sources, which is afaics not allowed in downstream distros (or frowned upon at least).

andreasgerstmayr:
Thanks for the PR!
I've moved the compilation to the Makefile, otherwise everything looks good :+1:

andreasgerstmayr:
Argh I broke the PR with my commit, I didn't realize we're using the bpftool binary later in the configure script.

@jasonk000 I was wondering if we still need the `--with-pmdabpf-btf` setting, or in general if we can remove any of the config options for simplicity:
* `--with-pmdabpf-clang`
* `--with-pmdabpf-llvm-strip`
* `--with-pmdabpf-bpftool`
* `--with-pmdabpf-btf`

I think we can get rid of all of them? For bpftool, is there a reason to use the distro's bpftool over the (probably more recent) vendored bpftool? Upstream iovisor/bcc does vendor `vmlinux.h` and `bpftool`, same as in PCP now.

jasonk000:
We currently use most of these options internally with 5.3.7 to build a BPF PMDA for older Ubuntu editions. I believe with this change, at least bpftool can go. I'll experiment a little but I think at least we would need `-btf` to specify a custom BTF location for our kernel (older Ubuntu installs do not provide BTF).

jasonk000:
@andreasgerstmayr Good news - I've unhooked this now, and we no longer need these additional `--with-pmdabpf-*` flags!

PR #1675, opened 2022-09-08
libpcp_web: guard client struct during callbacks

andreasgerstmayr:
This resolves multiple race conditions when the client closes the connection before the REST API is sending the entire response.

Case 1: Client sends request, Redis is queried, client closes connection due to timeout and client struct gets freed, Redis reply arrives and is getting sent to the already free'd client. Added locks in `pmseries_request_done`/`on_pmseries_done`.

Case 2: client_write checks if the client is still opened, and then calls uv_callback_fire which calls on_write_callback in a later loop iteration. Meanwhile, the client could be closed. Solved by adding another client_is_closed check in on_write_callback.

Case 3: Guard client struct while client_write/uv_callback_fire and on_write_callback. Client connection could be closed, and the client free'd meanwhile.

I only encountered these race conditions with a PCP archive with a big amount of instances (almost 400 cores, dozens of disks) and Grafana/grafana-pcp closing the connection early because of a timeout. Additional investigation is required to improve the performance for metrics with a large number of instances.

### DIFF
--- a/src/pmproxy/src/series.c
+++ b/src/pmproxy/src/series.c
@@ -593,6 +593,9 @@ on_pmseries_done(int status, void *arg)
 	flags |= HTTP_FLAG_JSON;
     }
     http_reply(client, msg, code, flags, options);
+
+    /* release lock of pmseries_request_done */
+    client_put(client);
 }
 
 static void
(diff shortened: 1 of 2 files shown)


natoscott:
@andreasgerstmayr good work, looks correct to me.  I see this is similar to the reference counting done in webapi.c as well around its 'done' routine (that we need it here too seems to be an oversight I made a few years back).  We also need this same logic in search.c so I'll add that and merge shortly.

PR #1680, opened 2022-09-23
build: remove obsolete pmdabpf-specific build options

andreasgerstmayr:
This removes the `--with-pmdabpf` options because we're vendoring bpftool and vmlinux.h now. Clang and llvm-strip can be configured by using env variables.


configure.ac
```diff
@@ -1202,60 +1178,29 @@ AS_IF([$have_libbpf], [
...
+if test -f "$pmdabpf_vmlinuxh"; then
+    AC_MSG_RESULT($pmdabpf_vmlinuxh)
 else
-    pmdabpf_arch_vmlinux="`pwd`/vendor/github.com/iovisor/bcc/libbpf-tools/$pmdabpf_arch/vmlinux.h"
-    if test -f "$pmdabpf_arch_vmlinux"; then
-	pmdabpf_vmlinuxh="$pmdabpf_arch_vmlinux"
-    else
-	AC_MSG_NOTICE([vmlinux.h for architecture $pmdabpf_arch is not vendored, please use --with-pmdabpf-btf to generate a vmlinux.h for $pmdabpf_arch])
-    fi
+    AC_MSG_RESULT(not found);
```
jasonk000, line 1200: Seems like a useful notice to mention that it's vendored or arch-specific. Something like `AC_MSG_RESULT(not found for arch $pmdabpf_arch)`

jasonk000:
LGTM.

configure.ac
andreasgerstmayr: Good point!
I've updated the error message.

PR #1690, opened 2022-10-12
pcp-ps:initial commit of PCP implementation of ps

orasagar:
pcp ps utility has been developed to view the ps like data


natoscott:
@orasagar nice!

The CI has picked up some LGTM.com warnings and "make check" with pylint installed also a few (probably overlapping).
Do the command line options for this pcp-ps map to any other version of ps (Linux? BSD?) or are they totally unrelated?

orasagar:
In pcp-ps options i tried to map it with the linux ps options:-
for eg pcp-ps -e option is mapped to ps -e option and so on.

natoscott:
> In pcp-ps options i tried to map it with the linux ps options:-

I agree that's the best approach.  In order that we don't end up with overlapping options between the (de-facto) standard PCP options and the ps(1) options, I recommend removing the use of -a/-s/... (PCP option handling) from pcp-ps entirely, and leave that up to the pcp(1) frontend.  Otherwise users will get surprised when options behave differently to expectation.

```
pcp -a /var/log/pcp/archives/host/file ps -a
```

orasagar:
Yeah @natoscott i agree with you on that point and i have updated the code and man page with the changes required ,please let me know if anything else need to be done .

natoscott:
> i have updated the code and man page with the changes required ,please let me know if anything else need to be done .

@orasagar thanks, this is the next issue - we run pylint as part of CI, so need all python code to be "pylint clean".  This is the output I see with latest Fedora (Python 3.10.7):

```
src/pcp/ps$ make check
for d in `echo test `; do if test -d "$d" -a -f "$d/GNUmakefile"; then /usr/bin/echo === $d ===; make --no-print-directory -C $d check || exit $?; fi; done
=== test ===
make[1]: Nothing to be done for 'check'.
pylint --rcfile=../../../.pylintrc --output-format=colorized pcp-ps.py
************* Module pcp-ps
pcp-ps.py:423:0: R1707: Disallow trailing comma tuple (trailing-comma-tuple)
pcp-ps.py:627:0: C0301: Line too long (122/120) (line-too-long)
pcp-ps.py:21:0: C0410: Multiple imports on one line (sys, signal, time) (multiple-imports)
pcp-ps.py:423:12: W0106: Expression "(print('Timestamp', end='\t'), )" is assigned to nothing (expression-not-assigned)
pcp-ps.py:708:20: R1714: Consider merging these comparisons with "in" to "opts in ('p', 'c', 'P', 'U', 'o')" (consider-using-in)
pcp-ps.py:748:8: R1702: Too many nested blocks (6/5) (too-many-nested-blocks)
pcp-ps.py:815:50: R1722: Consider using sys.exit() (consider-using-sys-exit)

------------------------------------------------------------------
Your code has been rated at 9.88/10 (previous run: 9.85/10, +0.03)

make: *** [GNUmakefile:43: check] Error 28
```

natoscott:
After that, the next two areas I observed needing further attention are:
- The output doesn't behave like ps(1) by default.  It seems to run in a loop, and doesn't produce columns (or headings) matching the ps output.
- The unit tests are great.  An integration test would be excellent - a new script of form ./qa/nnnn (see qa/README) that runs pcp-ps with one/more of the ./qa/archives/xxx files containing per-process metrics.  Running the tool on archives and with different command line options should produce deterministic results (use "pcp -z -a xxx ps ..." to control the test host timezone).

PR #1695, opened 2022-10-25
libpcp_web: fix scalar multiplication memory leaks

ShiyaoC:
Fix scalar multiplication memory leaks and float multiplication to series_calculate_star()


natoscott:
Thanks @ShiyaoC - rebased and merged.

PR #1696, opened 2022-10-26
selinux: fix perf_event selinux policy for el8 platforms

natoscott:
Since el8 selinux includes parts of the perf_event policy from latest Fedora/el9 we cannot define it completely away as we are currently.  Instead we need to carefully pick up just the bits we need.


natoscott:
Rebased and merged.

PR #1698, opened 2022-10-28
pmdabpf: fix qa/1810 (oomkill) to run in limited memory footprint

### LINKED ISSUE #1644: investigate feasibility of QA test for oomkill bpf PMDA module
Here's a cgroup example:

Single concat, runs inside 10MB
```
jkoch@jkoch:/sys/fs/cgroup/user.slice$ sudo systemd-run --scope -p MemoryLimit=10M perl -e '$a .= "A" x 1024'
Running scope as unit: run-r43d2f5e56b8847fea964d1d1a0bc19ec.scope
[completes]
```

With a loop, is terminated:
```
jkoch@jkoch:/sys/fs/cgroup/user.slice$ sudo systemd-run --scope -p MemoryLimit=10M perl -e 'while (1) { $a .= "A" x 1024 }'
Running scope as unit: run-r9ae1c624743547f58b4abdb773a73498.scope
Killed
```

Verifying:
```
jkoch@jkoch:/sys/fs/cgroup/user.slice$ sudo journalctl --since -5m | grep oom-kill
Aug 10 09:45:38 jkoch kernel: perl invoked oom-killer: gfp_mask=0xcc0(GFP_KERNEL), order=0, oom_score_adj=0
Aug 10 09:45:38 jkoch kernel: oom-kill:constraint=CONSTRAINT_MEMCG,nodemask=(null),cpuset=run-r9ae1c624743547f58b4abdb773a73498.scope,mems_allowed=0,oom_memcg=/system.slice/run-r9ae1c624743547f58b4abdb773a73498.scope,task_memcg=/system.slice/run-r9ae1c624743547f58b4abdb773a73498.scope,task=perl,pid=100884,uid=0
```

_Originally posted by @jasonk000 in https://github.com/performancecopilot/pcp/pull/1612#discussion_r942681513_

jasonk000:
resolves #1644

### DIFF
--- a/qa/1810
+++ b/qa/1810
@@ -42,7 +42,7 @@ $sudo systemd-run \
     --no-ask-password \
     --uid=pcpqa --gid=pcpqa \
     -p MemoryLimit=10M \
-    perl -e 'while (1) { $a .= "A" x 1024 }' >> $here/$seq.full 2>&1
+    perl -e 'while (1) { $a .= "A" x 1024 }' > $here/$seq.full 2>&1
 
 echo "=== report metric values for fcomm ==="
 pminfo -dfmtT bpf.oomkill.fcomm 2>&1 | tee -a $here/$seq.full \
--- a/qa/1810
+++ b/qa/1810
@@ -10,8 +10,6 @@ echo "QA output created by $seq"
 
 . ./common.bpf
 
-_notrun "see https://github.com/performancecopilot/pcp/issues/1644 for discussions about triggering oom-killer in a QA test"
-
 _pmdabpf_check
 _pmdabpf_require_kernel_version 5 0
 _pmdabpf_require_libbpf_version 0 7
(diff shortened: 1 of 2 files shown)


andreasgerstmayr:
Thanks for the PR!
When I try it as-is on F36 I'm seeing this error from systemd in 1810.full:
```
Failed to start transient service unit: Connection reset by peer
```

Replacing `>> $here/$seq.full` with `> $here/$seq.full` works though. I'm not sure why that is, afaics appending should work without issues even if the file doesn't exist. Any ideas?

In this case the file doesn't exist, so changing it to `>` shouldn't make any difference to the test.

jasonk000:
> Replacing >> $here/$seq.full with > $here/$seq.full works though. I'm not sure why that is, afaics appending should work without issues even if the file doesn't exist. Any ideas?

The `>>` should be handled by whatever is running the script, which is `#!/bin/sh`. 

Regardless - I've rebased on main and changed to `>`.

andreasgerstmayr (approved):

andreasgerstmayr:
I was also quite surprised why `>>` vs `>` makes any difference regarding the systemd call.
Yep, I also think it has to do with the stdout file descriptor.

PR #1715, opened 2022-11-16
build: determine clang version without llvm-config if not installed

natoscott:


configure.ac
```diff
@@ -1167,6 +1167,9 @@ AC_MSG_CHECKING([clang major version installed])
 if which llvm-config >/dev/null 2>&1
```
andreasgerstmayr, line 1167: this condition could be removed now

andreasgerstmayr (commented):
I noticed the configure script has 10441 lines changed (different autoconf version?), is this expected?
otherwise, lgtm

jasonk000:
Two thoughts:
1. If we are detecting `clang` version directly, we can rely on that only, instead of `llvm-config`, .. but
2. It seems that we use `llvm-strip` [here](https://github.com/performancecopilot/pcp/blob/main/src/pmdas/bpf/modules/GNUmakefile#L98-L101), which implies (at least for Ubuntu) that the user should have `llvm` installed.

Maybe we should use your approach for detecting clang-version and separately detect presence of matching llvm-strip?

configure.ac
natoscott: +1 & yeah, new version of autoconf in f37 I guess.

natoscott:
> Maybe we should use your approach for detecting clang-version and separately detect presence of matching llvm-strip?

Yep - I think the reason I've encountered this is related to the way the llvm/clang packaging is done on Fedora, where llvm-config is in the optional llvm-devel package while other clang tools are in either llvm or clang packages.  I'll update it to check for the binaries we use I think and drop use of llvm-config.

configure.ac
```diff
@@ -1190,6 +1179,17 @@ test -z "$CLANG" && AC_PATH_PROG(CLANG, clang)
 clang=$CLANG
 AC_SUBST(clang)
 
+AC_MSG_CHECKING([clang major version installed])
+if which $CLANG >/dev/null 2>&1
+then
+    clang_major_version=`$CLANG --version | awk '/clang version/ {print $3}' | cut -f1 -d.`
```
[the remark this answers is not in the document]
natoscott: @jasonk000 ah, thanks - this seems to be a more robust approach ...

```
clang --version | sed -n -e 's/.*clang version \([0-9][0-9]*\).*/\1/p'
```

PR #1719, opened 2022-11-23
dstat: fix columns for --net-packets with multiple network interfaces

jkter:
When using `dstat --net-packets` with more network interfaces (defined by `-N`), both terminal and CSV outputs are corrupted, e.g.
pcp-dstat-current
This patch fixes this.
pcp-dstat-patched
The output is the same (up to general differences) as with the original no-pcp `dstat` command.


natoscott:
Thanks Jakub, I've extended regression test qa/1187 to exercise this case as well - running tests now, will merge shortly.

PR #1725, opened 2022-12-09
pcp-ps:added capabilities to show n sample with archives

orasagar:
we are checking if our current context is archive then don't limit the output print to one instance of data.


natoscott:
@orasagar let's add a regression test with this new functionality.  Can you make an archive with pmlogger containing two or three samples of the pcp-ps metrics and add it to this PR, below qa/archives/?  (you might want to add a mk.ps script like qa/archives/mk.atop to help generate this).  Then we can add a little script that will produce deterministic output from pcp-ps with this new command line option.

Thanks!

orasagar:
@natoscott i have added the regression test for pcp-ps utility. please take a look and let me know something else is required

natoscott:
@orasagar that's perfect - thanks!

PR #1727, opened 2022-12-12
configure.ac: Fixes for improved C99 compatibility

fweimer-rh:
Avoid calling the undeclared exit function.  Include <string.h> for strcmp.  C99 removed implicit function declarations from the language, and future compilers are likely to reject them by default.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3103,7 +3103,7 @@ AC_TRY_LINK(
 ],
 [
     double x = 123.456;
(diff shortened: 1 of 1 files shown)


natoscott:
Thanks @fweimer-rh - LGTM.

PR #1728, opened 2022-12-14
build: switch to using libbpf from vendor subdir

natoscott:
This aims to resolve recent fallout on Fedora36 which, although a recent Fedora, does not have the symbols we need for latest pmdabpf modules.  As a result, the build is cleaner - we don't need to have so much configure.ac logic for pmdabpf code.  We can always ship all the latest modules now rather then trying to detect newer symbols in the system libbpf and conditionally including some.


kmcdonell:
@natoscott I pulled this and gave it a try on vm36 (F36) where the bpf PMDA QA tests are mostly failing.

First, the configure.ac change is NQR as contains the classical 'configure-is-really-m4-so-[-is-special' issue.  The change below seems to fix that (without this the bpf PMDA is never included in the build).
```
git diff --no-ext-diff
diff --git a/configure b/configure
index abffb031e..c59ebbb9e 100755
--- a/configure
+++ b/configure
@@ -9079,7 +9079,7 @@ clang=$CLANG
 printf %s "checking clang major version installed... " >&6; }
 if which $CLANG >/dev/null 2>&1
 then
-    clang_major_version=`$CLANG --version | sed -n -e 's/.*clang version \(0-90-9*\).*/\1/p'`
+    clang_major_version=`$CLANG --version 2>&1 | sed -n -e 's/.*clang version \([0-9][0-9]*\).*/\1/p'`
 else
         clang_major_version=0
 fi
diff --git a/configure.ac b/configure.ac
index 4378a1303..8de3bc593 100644
--- a/configure.ac
+++ b/configure.ac
@@ -1182,7 +1182,7 @@ AC_SUBST(clang)
 AC_MSG_CHECKING([clang major version installed])
 if which $CLANG >/dev/null 2>&1
 then
-    clang_major_version=`$CLANG --version | sed -n -e 's/.*clang version \([0-9][0-9]*\).*/\1/p'`
+    clang_major_version=`$CLANG --version 2>&1 | sed -n -e 's/.*clang version \([[0-9]][[0-9]]*\).*/\1/p'`
 else
     dnl version=0 => not installed
     clang_major_version=0
```

But I'm still seeing QA failures.   I checked and libbpf is not a DSO for pmdabpf (so the static link part is working), however for example qa/1475 fails and this seems to be the key error message:
```
ok
> [Sat Dec 17 12:13:55] pmdabpf(3089542) Info: loading modules
> [Sat Dec 17 12:13:55] pmdabpf(3089542) Info: loading modules
> [Sat Dec 17 12:13:55] pmdabpf(3089542) Info: loading mountsnoop.so
> [Sat Dec 17 12:13:55] pmdabpf(3089542) Info: loading: mountsnoop.so from /var/lib/pcp/pmdas/bpf/modules/mountsnoop.so
> [Sat Dec 17 12:13:55] pmdabpf(3089542) Info: dlopen failed: /var/lib/pcp/pmdas/bpf/modules/mountsnoop.so, /var/lib/pcp/pmdas/bpf/modules/mountsnoop.so: undefined symbol: bpf_map__set_autocreate
```
I suspect the problem is that the symbols are in libbpf and mountsnoop.o but not pmdabpf ... as a consequence of static linking rather than a DSO version of libbpf.

Anyway, that's my 2 cents worth.

kmcdonell:
Indeed @natoscott , the remaining issue is that the modules are not linked with the static libbpf, but are linked against the system-installed DSO.

The change below appears to fix it, and now most of the -g pmda.bpf tests are passing on vm03 (except 1444 [no values for comm?], 1452 [triaged and has never passed?], 1810 [has been failing for other reasons prior to this change] and 1900 [looks like real valgrind issues]).
```
diff --git a/src/pmdas/bpf/modules/GNUmakefile b/src/pmdas/bpf/modules/GNUmakefile
index 2f7b6c2aa..62bbbb8b3 100644
--- a/src/pmdas/bpf/modules/GNUmakefile
+++ b/src/pmdas/bpf/modules/GNUmakefile
@@ -2,7 +2,7 @@ TOPDIR = ../../../..
 include $(TOPDIR)/src/include/builddefs
 
 IAM    = bpf
-LDLIBS = $(PCP_WEBLIB) -lbpf -lelf -lz -lm -ldl
+LDLIBS = $(BPF_LIBRARY) $(PCP_WEBLIB) -lelf -lz -lm -ldl
 LDIRT = $(HELPERS_H) $(HELPERS_C) $(APPS_H) $(APPS_BPF) vmlinux.h
 MODULETMP = $(PCP_PMDAS_DIR)/$(IAM)/modules
 MODULEDIR = $(PCP_PMDASADM_DIR)/$(IAM)/modules
```

natoscott:
Closing as we took a different approach over in #1739 that continues using the system libbpf.so

PR #1729, opened 2022-12-15
statsd: char static metric memory leak fix

Erbenos:
Even though all string static metrics were being returned with PMDA_FETCH_STATIC, the vars were being allocated each time the metric was requested.


Erbenos:
Addresses #1729

qa/statsd/src/cases/15.py
```diff
@@ -119,6 +119,20 @@
...
+    'echo "statsd.pmda.settings.verbose"',
+    'echo "statsd.pmda.settings.max_unprocessed_packets"',
+    'echo "statsd.pmda.settings.max_udp_packet_size"',
+    'echo "statsd.pmda.time_spent_aggregating"',
+    'echo "statsd.pmda.time_spent_parsing"',
+    'echo "statsd.pmda.metrics_tracked"',
+    'echo "statsd.pmda.aggregated"',
+    'echo "statsd.pmda.dropped"',
+    'echo "statsd.pmda.parsed"',
+    'echo "statsd.pmda.received"',
```
natoscott, line 135: @Erbenos do all of the above additions need to be prefixed with 'fetch'?  (this is passed to dbpmda right?)

Erbenos: Right, was too fast on the trigger

Erbenos: Addressed by 516de22

PR #1738, opened 2023-02-03
build: update actions set-output handling based on githubs advice

natoscott:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -161,6 +161,7 @@ jobs:
         continue-on-error: true
         with:
           script: |
+            const core = require('@actions/core');
             const fs = require('fs');
             const head_sha = context.payload.pull_request ? context.payload.pull_request.head.sha : context.sha;
             const response = await github.rest.checks.create({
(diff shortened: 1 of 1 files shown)


andreasgerstmayr:
lgtm, can you do the same change for the `Publish test summary` step as well?

natoscott:
> 

@andreasgerstmayr yeah I looked at that, but not sure it needs the same treatment because its not logging to stdout?  (seems to be using console.log there)

andreasgerstmayr:
> @andreasgerstmayr yeah I looked at that, but not sure it needs the same treatment because its not logging to stdout? (seems to be using console.log there)

When JavaScript runs in a browser environment, `console.log` logs to the browsers' web console (hence the name), but when it runs in Node.js, `console.log` prints to stdout. This code sets the "View test summary" link in the daily Slack message.

natoscott:
> > @andreasgerstmayr yeah I looked at that, but not sure it needs the same treatment because its not logging to stdout? (seems to be using console.log there)
> 
> When JavaScript runs in a browser environment, `console.log` logs to the browsers' web console (hence the name), but when it runs in Node.js, `console.log` prints to stdout. This code sets the "View test summary" link in the daily Slack message.

OK - any idea's on how to incorporate $GITHUB_OUTPUT into that though?  Looks like that'd have to be within console.log wouldn't it?  (so, do we just change the input string part?)  Thanks!

PR #1742, opened 2023-02-10
build: raise the minimum libbpf version to 0.8.0

natoscott:
Recent updates to the vendored bcc code resulted in a newer version of libbpf-tools/compat.c merged which uses bpf_map__set_autocreate. This is only available in libbpf 0.8.0 and later and quite a few of the shipped modules are affected as a result.

This resolves a series of CI problems observed in the f36 BPF tests.


jasonk000:
fwiw this now precludes build of pmdabpf on ubuntu jammy and below [as per their support for libbpf](https://packages.ubuntu.com/search?lang=pt&keywords=libbpf&searchon=sourcenames).

PR #1756, opened 2023-04-17
pmieconf support for a webhook action

natoscott:


natoscott:
Thanks Jan & agreed - will add that change.

PR #1766, opened 2023-06-20
pcp2json: extend pcp2json with an option to send HTTP POSTs

natoscott:


src/pcp2json/pcp2json.1
```diff
@@ -439,6 +442,11 @@ Refer to
 for a complete description of the syntax for
 .IR origin .
 .TP
+\fB\-p\fR, \fB\-\-http-pass\fR
+Password when using HTTP basic authentication wh the
```
myllynen, line 446: Typo.

src/pcp2json/pcp2json.1
```diff
@@ -547,6 +555,14 @@ determines the actual time
 will run.
 .RE
 .TP
+\fB\-u\fR, \fB\-\-url\fR
+URL for sending an HTTP POST (instead of default standard output).
+.TP
+\fB\-U\fR, \fB\-\-http-user\fR
+Username when using HTTP basic authentication wh the
```
myllynen, line 562: Typo.

src/pcp2json/pcp2json.1
```diff
@@ -547,6 +555,14 @@ determines the actual time
 will run.
 .RE
 .TP
+\fB\-u\fR, \fB\-\-url\fR
```
myllynen, line 558: There's a section for 'pcp2json specific options' in the man page describing how these should be in the config file, I think these should be mentioned there as well.

src/pcp2json/pcp2json.py
```diff
@@ -47,6 +48,7 @@
 CONFVER = 1
 INDENT = 2
 TIMEFMT = "%Y-%m-%d %H:%M:%S"
+TIMEOUT = 2.5 # seconds
```
myllynen, line 51: Is this something that doesn't need to be configured? If so, would slightly larger timeout be better?

src/pcp2json/pcp2json.py
```diff
@@ -116,6 +118,10 @@ def __init__(self):
         self.space_scale_force = None
         self.time_scale = None
         self.time_scale_force = None
+        self.timeout = TIMEOUT
+        self.url = None
+        self.http_user = None
+        self.http_pass = None
```
myllynen, line 124: These should be listed few lines below in the tool specific block.

src/pcp2json/pcp2json.py
```diff
@@ -151,7 +157,7 @@ def options(self):
         opts = pmapi.pmOptions()
         opts.pmSetOptionCallback(self.option)
         opts.pmSetOverrideCallback(self.option_override)
-        opts.pmSetShortOptions("a:h:LK:c:Ce:D:V?HGA:S:T:O:s:t:rRIi:jJ:4:58:9:nN:vmP:0:q:b:y:Q:B:Y:F:f:Z:zxXE")
+        opts.pmSetShortOptions("a:h:LK:c:Ce:D:HGA:S:T:O:s:t:rRIi:jJ:4:58:9:nN:vmP:0:q:b:y:Q:B:Y:F:f:Z:zxXEp:U:u:V?")
```
myllynen, line 160: IIRC there was some sort of logic to keep "common parameters" at the beginning followed by "mostly common" and then lastly tool specific options but I don't think anyone remembers that logic anymore.

src/pcp2json/pcp2json.py
```diff
@@ -524,13 +540,11 @@ def create_attrs(value, inst_id, inst_name, unit, pmid, desc, labels):
 
     def finalize(self):
         """ Finalize and clean up """
+        data = json.dumps(self.data, indent=INDENT, sort_keys=True, ensure_ascii=False, separators=(',', ': '))
+        print("RAW:", data)
```
myllynen, line 544: Looks like debuggging leftover?

src/zshrc/_pcp
```diff
@@ -183,6 +183,9 @@ _pcp () {
       "(-m --include-labels -m --include-labels $exargs)"{-m,--include-labels}'[include labels]' \
       "(-n --invert-filter $exargs)"{-n,--invert-filter}'[perform ranking before live filtering]' \
       "(-N --predicate $exargs)"{-N+,--predicate=}'[specify predicate metric]:metric:->metrics' \
+      "(-u --url $exargs)"{-u+,--url=}'[URL of endpoint to receive HTTP POST]' \
+      "(-U --http-user $exargs)"{-U+,--http-user=}'[username for HTTP basic authentication]' \
+      "(-p --http-pass $exargs)"{-p+,--http-pass=}'[password for HTTP basic authentication]' \
```
myllynen, line 188: A nit-pick comment here, earlier these HTTP POST / URL related params were added after the other tool-specific params but here they are added before them, ideally would consistently add always after the earlier params but no biggie.

src/pcp2json/pcp2json.1
natoscott: Fixed, thanks.

src/pcp2json/pcp2json.1
natoscott: Fixed, thanks.

src/pcp2json/pcp2json.py
natoscott: Fixed, thanks.

src/pcp2json/pcp2json.py
natoscott: Ah, I see the pattern - fixed now.

src/pcp2json/pcp2json.py
natoscott: Yep, gone now.

src/zshrc/_pcp
natoscott: Fixed, thanks.

src/pcp2json/pcp2json.py
natoscott: Oh - I just did the same thing pcp2influxdb does by hardcoding this.  But I think this use-case is more likely to send to a faraway host so I've added a new config option now.

src/pcp2json/pcp2json.1
natoscott: Fixed, thanks.

PR #1768, opened 2023-06-20
Update help with proper description of DC status

rsponholtz:
the field ha_cluster.pacemaker.nodes.status.dc actually tracks whether the node is the DC in the cluster or not.


natoscott:
@rsponholtz thanks Ross - looks good.  The CI issue is an unrelated networking issue in updating a container so I'll merge this and that should turn green shortly.

PR #1769, opened 2023-06-22
pcp-meminfo: initial commit of PCP implementation of meminfo

mohith-kumar-thummaluru:
pcp meminfo utility has been developed to view the /proc/meminfo data on live as well as on archive data


natoscott:
@mohith-kumar-thummaluru thanks for your contribution - I'm away on holiday shortly but will take a closer look when back online.

src/pcp/meminfo/pcp-meminfo.py
```diff
@@ -0,0 +1,211 @@
...
+# for more details.
+#
+# pylint: disable=bad-whitespace,too-many-arguments,too-many-lines, bad-continuation
+# pylint: disable=redefined-outer-name,unnecessary-lambda
+#
+
+import sys
+import time
+from pcp import pmapi
+from pcp import pmcc
```
natoscott, line 23: These can be combined on a single line using "fom pcp import pmapi, pmcc"

src/pcp/meminfo/pcp-meminfo.py
```diff
@@ -0,0 +1,211 @@
...
+            sys.exit(0)
+
+class MeminfoOptions(pmapi.pmOptions):
+    context = None
+    timefmt = "%H:%M:%S"
+    samples = 0
+
+    def __init__(self):
+        pmapi.pmOptions.__init__(self, "a:s:S:T:z:P:A:R:t:T:x:")
+        self.pmSetLongOption("archive", 1, "a", "FILENAME","Fetch /proc/meminfo for a specified archive file")
```
natoscott, line 172: Use pmSetLongOptionArchive here

src/pcp/meminfo/pcp-meminfo.py
```diff
@@ -0,0 +1,211 @@
...
+        self.pmSetLongOption("samples", 1, "s", "count","Get the meminfo for specified number of samples count")
```
natoscott, line 173: Use pmSetLongOptionSamples here

src/pcp/meminfo/pcp-meminfo.py
```diff
@@ -0,0 +1,211 @@
...
+        self.pmSetLongOption("start_time", 1, "S", "TIME","Filter the samples from the archive from the given time")
```
natoscott, line 174: Use pmSetLongOptionStart

src/pcp/meminfo/pcp-meminfo.py
```diff
@@ -0,0 +1,211 @@
...
+        self.pmSetLongOption("end_time", 1, "T", "TIME","Filter the samples from the archive till the given time")
```
natoscott, line 175: pmSetLongOptionFinish here

src/pcp/meminfo/pcp-meminfo.py
```diff
@@ -0,0 +1,211 @@
...
+        self.pmSetOptionCallback(self.options)
+        self.pmSetOverrideCallback(self.override)
```
natoscott, line 177: Don't need this Override line - you're using only standard options.

src/pcp/meminfo/pcp-meminfo.py
```diff
@@ -0,0 +1,211 @@
...
+        self.pmSetLongOptionHelp()
+
+    def override(self, opt):
```
natoscott, line 180: Can remove this function, its not needed

src/pcp/meminfo/pcp-meminfo.py
```diff
@@ -0,0 +1,211 @@
...
+        if opt == 's':
+            return 1
+        return 0
+
+    def options(self,opt,optarg,index):
+        if opt == 's':
+            MeminfoOptions.samples = int(optarg)
```
natoscott, line 187: Ah I see why you handle 's' differently now - you don't need this either, use opts.pmGetOptionSamples() at around line 140 (you could pass this value into a MeminfoReport __init__ function, called at line ~201).

src/pcp/GNUmakefile
```diff
@@ -25,6 +25,7 @@ SUBDIRS = \
 	iostat \
         ipcs \
 	mpstat \
+	meminfo \
```
natoscott, line 28: This list is in alphabetic order IIRC - if so, this line needs to be up one spot.

natoscott (changes requested):
@mohith-kumar-thummaluru Good stuff!  Can you look through my notes here and then also add a regression test similar to the one at qa/1398 for pcp-pidstat (will need to create a small, compressed archive or two to go along with the test - see 1398 for an example of that) - thanks!

src/pcp/meminfo/pcp-meminfo.py
mohith-kumar-thummaluru: Hi scott, 

I tried getting number of samples using the opts.pmGetOptionSamples(), but it was not able to retrieve the number successfully, instead it's showing "None" no matter how many number of samples are passed. So, as a workaround to that, I implemented it in this manner. Anyways, let me take a look at it again if I can modify this using opts.pmGetOptionSamples(). 

Thanks!

mohith-kumar-thummaluru:
Thanks for the comments, Scott. Let me fix these and get back to you.

src/pcp/meminfo/pcp-meminfo.py
natoscott: @mohith-kumar-thummaluru make sure you access it after option-processing has been completed - because of code abstractions its not always obvious where that happens though.  One safe place would be at the point the MeminfoReport object is created as this happens relatively late in the life of the program.

mohith-kumar-thummaluru:
Hi scott,

I have updated the code related changes that you have shared over your notes, and regression tests are not yet done. I'll look into it and update you

Thanks

src/pcp/meminfo/pcp-meminfo.py
```diff
@@ -0,0 +1,195 @@
...
+            idx += 1
+        print()
+
+class MeminfoOptions(pmapi.pmOptions):
+    context = None
+    timefmt = "%H:%M:%S"
+    samples = 0
+
+    def __init__(self):
+        pmapi.pmOptions.__init__(self, "a:s:S:T:z:P:A:R:t:T:x:")
```
natoscott, line 168: @mohith-kumar-thummaluru looking good!  I think some of the short options listed are unused? Definitely P:, R: and x: look like they could be dropped from the string.
Then I think we're good to merge this - thanks!

mohith-kumar-thummaluru: Thanks for the comments, @natoscott. I'll filter the unused options and let you know

natoscott:
@mohith-kumar-thummaluru thanks - I've made minor changes to QA (please double check) & merging now.

PR #1776, opened 2023-07-19
selinux: add policy updates for pmproxy (io_uring,ipc_lock)

natoscott:
Refactoring in core selinux policy has impacted on pmproxy.

Resolves Red Hat BZ #2223568


zpytela:
The content is good, but it does not need to be in the optional block, neither the class needs explicit declaration.
The reason actually is a new feature in libuv which requires the permission for services linked with libuv.

natoscott:
> 

@zpytela thanks - without that though the PCP build fails on platforms like RHEL 7.

PR #1795, opened 2023-08-16
docs: add diversity note to contributing.md and other small updates

natoscott:


kmcdonell (approved):
LGTM ... other than the respond in 24 hrs obligation 😊

natoscott:
> LGTM ... other than the respond in 24 hrs obligation 😊

Oh yeah, good point - I left that as-is ... is it time for an update?  48hrs be better?  2/3 business days?  Other?  Thanks.

christianhorn:
I wonder if projects can get away without referring to a time for initial assessment, but "reasonable time". Feels a bit like ticking clock otherwise.
Reporters are probably also setting time expectations, which might or might not fit the projects expectations.
LGTM (also if sticking with 24 or 48 hours for first assessment).

jasonk000 (approved):
lgtm, agree with other comments!

sfeifer (approved):

lmchilton (approved):

andreasgerstmayr (approved):

kmcdonell:
I think "within 3 business days" for initial response is both reasonable and matches what we've been able to deliver for the most part for almost 30 years.   Given the maturity of the code, we're unlikely to be seeing any "mission-critical" issues that would demand a more expeditious response.

natoscott:
Thanks all - let's go with 3 business days.

Issue #1799, opened 2023-08-29
Provide more reporting flexibility when running pmie rules against archives 


orasagar:
Running pmie rule in live system 

$ /usr/bin/pmie -c ~/myrule
Tue Aug 29 13:21:02 2023: load average OK true
Tue Aug 29 13:21:03 2023: load average OK true
Tue Aug 29 13:21:04 2023: load average OK true
Tue Aug 29 13:21:05 2023: load average OK true
$

Running the same rule on the archive 
 $ /usr/bin/pmie -c ~/myrule -a /etc/pcp/pmlogger/localhost/20230829.00.10.3 | more
print Tue Aug 29 00:10:15 2023: load average unavailable
print Tue Aug 29 00:10:16 2023: load average OK true
print Tue Aug 29 00:10:17 2023: load average OK true

There is extra "print" that gets printed in the case of the archive which makes it inconsistent for archive and live system cases.

orasagar:
#1800  resolves this issue

kmcdonell:
Since PR #1800 has been abandoned, here is the text (from there) describing how the archive-mode pmie action reported could be made more flexible:

We cannot make the existing behaviour go away due to backwards compatibility goals (who know what scripts an unconditional change in pmie behaviour might break).

So we'd have to add some flag or option, e.g. -f|--format where is a combination of literal text and the metafield identifiers %a (action), %m (message), %d (ctime date as is), so the current (and default) behaviour would be the same as -f '%a %d %m', and what was originally desired in this issue would be -f '%d %m'.

This mechanism could be extended to provide additional metafields, e.g. %f (config file name), %l (source line number of rule in config file), %t (higher precision time), ...

I will also amend the title of this issue to more accurately reflect what needs to be done.

Issue #1779, opened 2023-07-25
pmie -a archive, shell actions do not work as expected


GerryHickman:
pmie can evaluate "rules" and trigger "actions", in this case it's checking how many processes are running
```
cat pmie_test.txt
delta = 10 sec;
nprocs_high = kernel.all.nprocs > 10 count/sec * hinv.ncpu -> shell "whoami";

```
every ten seconds, if more than ten processes are running, the action will trigger.

When I run this against a live host using '-h', it works correctly, the shell starts and prints my username.

When I run this against an archive using '-a', the action triggers, but the shell does NOT start and does NOT print my username, it basically just prints the action as a literal string onto stdout

natoscott:
Yes, I believe that's expected behaviour and operating as designed.  OOC what would be the use case for not behaving that way?  (I'm sure you can see the downside of not going this, such as repeatedly writing to syslog when testing new rules, running commands "live" when the timestamp for the condition being detected is days ago... etc, etc).

GerryHickman:
Thanks for the update; if it's expected behaviour, where does it say so in the manual pages, docs etc?

The use case would be where you want to analyze yesterday's archive file, e.g. at 10 minutes after midnight, and log abnormal activity, e.g. to a database, or you might want to analyze the most recent 10 minutes of today's archive file (from pmlogger).

For testing new rules, I don't think you'd be using an archive file for that, and you can easily change the action from 'print' to 'shell' once the new rule is working. For timestamps being from days ago, I don't see this being an issue in my own testing, e.g. I want to analyze an archive file from last week, I can just specify the file name and a time of 10:00 to 11:00.

Either way, printing the action as a string seems wrong, the print command doesn't even work properly, it prints the word 'print'.

natoscott:
https://pcp.readthedocs.io/en/latest/UAG/PerformanceMetricsInferenceEngine.html#introduction-to-pmie

```
Note
When pmie is processing performance metrics from a set of PCP archive logs, the actions will be processed in the expected manner; however, the action methods are modified to report a textual facsimile of the action on the standard output.
```

| For testing new rules, I don't think you'd be using an archive file for that

I've personally used it for years in this way and have found it very useful myself.  YMMV.

natoscott:
I'll leave this issue be until @kmcdonell  is back from his travels, as he knows pmie far better than I do.

If this functionality would be useful to you though, I'm sure a case could be made for a new pmie command line option to toggle your preferred behaviour here.

[...]

PR #1800, opened 2023-08-29
pmie: making pmie rule print consistent along with or without archive input

orasagar:
There is "print" written while running the pmie rule on archives but it is not there while running the same rule on the live system, this commit makes it consistent for both cases.

### DIFF
--- a/src/pmie/src/act.sk
+++ b/src/pmie/src/act.sk
@@ -392,7 +392,7 @@ actFake(Expr *x)
 	x->smpls[0].stamp = now;
 	pmCtime(&clock, bfr);
 	bfr[24] = '\0';
-	printf("%s %s: %s\n", opStrings(x->op), bfr, (char *)arg1->ring);
+	printf("%s: %s\n", bfr, (char *)arg1->ring);
     }
 }
 


kmcdonell (commented):
@orasagar I understand sentiment here, but I don't agree with this change.

There is a long discussion over at https://github.com/performancecopilot/pcp/issues/1779 on a very similar topic, and I've recently update the man page and docs to explain why "actions" (it is more than just print) are different for live vs archive rule firing.

Without a compelling use case (that I asked for but have not yet received over on #1779) I'd prefer to leave pmie as is.

orasagar:
@kmcdonell  thanks, for the clarification

But I still feel like the action is a different thing but while just reading from the archive shouldn't be different from the live system.

YoderExMachina:
I can understand why you would not want alerts popping up or syslog being updated when you are looking at archive data.  Here's my use case:
I'm writing a health checker that will be run exclusively on archived data uploaded by various customers.  The print at the beginning of the line is redundant in this circumstance.

It seems that this functionality exists to troubleshoot what would happen on a live system, but I'm dealing exclusively with archive data.

Thanks for your time and consideration.

kmcdonell:
Thanks @YoderExMachina.
Your use case matches the majority of my uses of pmie over the decades, namely in retrospective performance analysis.
Further the pmie-with-archives behaviour assists with rule development and tuning against historical data ... so pmie was conceived to be useful for much more than live system monitoring.

If I invoke circa-1970's Unix-thinking, I would suggest that
```
$ pmie .... | sed -e 's/^print //'
```
is a better solution that adding another flag to pmie to introduce a small variant in behaviour.

We cannot make the existing behaviour go away due to backwards compatilibility goals (who know what scripts an unconditional change in pmie behaviour might break), so we'd have to **add** some flag or option ... and if I was persuaded to do this it would be much more general that "don't emit print", e.g. -f|--format <format> where <format> is a combination of literal text and the metafield identifiers %a (action), %m (message), %d (ctime date as is), %D (ctime date with sub-second precision), so the current (and default) behaviour would be the same as -f '%a %d %m', and what you want would be -f '%d %m'.

Note also that the change as proposed in this PR is not acceptable because it removes the action type for **all** actions, not just print.

YoderExMachina:
Okay I understand.  Will stick with using sed.

YoderExMachina:
@kmcdonell This is off topic and sorry if this isn't kosher, but is there some kind of community forum for PMIE?  I'm new to the language and would like to read up on what others have done.

kmcdonell:
@YoderExMachina  I am willing to consider the -f|--format extension if that would really help ... it solves your issue, but also it provides a framework  to expose additional information in the future (different date formats, configuration file and line number for the rule that is firing, ...).

I'd need some indications from others that this is wothwhile, and then some cycles when this itch gets to the top of my TODO list.

As to the pmie "forum", there really isn't one per se.  Best options are https://pcp.readthedocs.io/en/latest/QG/AutomatedReasoningBasics.html then the mailing list pcp@groups.io or Slack  (join at https://h7zo83mvt1.execute-api.us-west-2.amazonaws.com/Express/) or email directly to me (kenj at kenj.id.au).

kmcdonell:
@orasagar Kudos for being brave enough to jump into the pmi source, but the change you're suggestion does not really work.  Are you OK if I close this PR and move the discussion to a new issue?

orasagar:
@kmcdonell  Yeah sounds good to me, Although I have already opened an issue for this and we can continue our discussion there

kmcdonell:
We're moving the discussion to https://github.com/performancecopilot/pcp/issues/1799 and expect a different approach to emerge compared to the changes outlined in this PR.

PR #1813, opened 2023-09-18
pmdagfs2: Correct permissions in makefile

pauljevans:
Correct permissions to 755 instead of 775 in makefile for GFS2 PMDA.


kurik:
To make this fix work, the line number 50 of the `src/pmdas/gfs2/GNUmakefile` file needs to change permissions from 775 to 755 as well. @pauljevans can you please check it and fix this line as well ?
Thanks in advance.

PR #1814, opened 2023-09-21
pcp-zoneinfo:Fixing the issue where pcp-zoneinfo utility will fail with older version of pcp

orasagar:
The requirement of __print_old_version_values is because the older version of PCP metrics is giving data for movable zones also. For eg:- In an older version
$ pminfo -f mem.zoneinfo.nr_vmscan_write
mem.zoneinfo.nr_vmscan_write
    inst [0 or "DMA::node0"] value 0
    inst [1 or "DMA32::node0"] value 0
    inst [2 or "Normal::node0"] value 0
    inst [3 or "Movable::node0"] value 0
    inst [4 or "Device::node0"] value 0
$ pcp --version
pcp version 4.3.2
$

but with the latest version
$ pminfo -f mem.zoneinfo.nr_vmscan_write

mem.zoneinfo.nr_vmscan_write
    inst [0 or "node0"] value 0
$ pcp --version
pcp version 6.1.1
$
So to make it consistent we have implemented new print function

### DIFF
--- a/src/pcp/zoneinfo/pcp-zoneinfo.py
+++ b/src/pcp/zoneinfo/pcp-zoneinfo.py
@@ -259,7 +259,7 @@ class ZoneinfoReport(pmcc.MetricGroupPrinter):
                 # these nodes will have only pages information for them so just printing them out
                 remaining_nodes = set(node_types) - set(nodes)
                 if remaining_nodes:
-                    for node in remaining_nodes:
+                    for node in sorted(remaining_nodes):
                         print(self.__format_node_name(node))
                         for key, value in ZONEINFO_PAGE_INFO:
                             print("\t", key, manager.metric_value(value, node))
--- a/src/pcp/zoneinfo/pcp-zoneinfo.py
+++ b/src/pcp/zoneinfo/pcp-zoneinfo.py
@@ -21,14 +21,15 @@ import sys
 import time
 from pcp import pmapi, pmcc
 from cpmapi import PM_CONTEXT_ARCHIVE
+from cpmapi import PM_CONTEXT_HOST
 
-# By default we are assuming we are runnig on the latest pcp version
+# By default we are assuming we are running on the latest pcp version
 # Later will figure out the version and will change the newVersionFlag
 
 newVersionFlag = True
 
-SYS_MECTRICS = ["kernel.uname.sysname","kernel.uname.release",
-               "kernel.uname.nodename","kernel.uname.machine","hinv.ncpu"]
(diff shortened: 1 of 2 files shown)


src/pcp/zoneinfo/pcp-zoneinfo.py
```diff
@@ -290,16 +298,54 @@ def __init__(self):
...
         opts = ZoneinfoOptions()
         mngr = pmcc.MetricGroupManager.builder(opts,sys.argv)
         opts.context = mngr.type
+
+        # After this VERSION new zoneifo metric has been introduced so checking the current version,
+        # if it is the latest version of pcp then get all the data otherwise drop the metrics
+        # which will not be there in older version.
+        # this change make pcp-zoneinfo comptabile with rhel7 and ol7 where VERSION 4.x.x being used.
+        VERSION="5.0.0"
+        current_version=pmapi.pmContext.pmGetConfig("PCP_VERSION")
```
natoscott, line 332: @orasagar this is looking at the local installed PCP version though.  What if we're replaying an archive or connected to a remote host with different version?  A better approach is to attempt to use the new metric, and fallback to the other form if the new metric doesn't exist.

natoscott (changes requested):
@orasagar please drop the explicit version check as mentioned, and use a dynamic fallback mechanism.

src/pcp/zoneinfo/pcp-zoneinfo.py
orasagar: @natoscott that sound reasonable ,will fix it and get back

natoscott:
@orasagar thanks - can you check test qa/1973 please?  its failing with this change, but the simple ./remake fix doesn't seem to be correct, I see non-deterministic output like:

```
1973 2s ... - output mismatch (see 1973.out.bad)
384c384
< Node 0, zone    Device
---
> Node 0, zone    Movable
393c393
< Node 0, zone    Movable
---
> Node 0, zone    Device
516c516
< Node 0, zone    Device
---
> Node 0, zone    Movable
525c525
< Node 0, zone    Movable
---
> Node 0, zone    Device
648c648
< Node 0, zone    Device
---
> Node 0, zone    Movable
657c657
< Node 0, zone    Movable
---
> Node 0, zone    Device
782c782
< Node 0, zone    Device
---
> Node 0, zone    Movable
791c791
< Node 0, zone    Movable
---
> Node 0, zone    Device
914c914
< Node 0, zone    Device
---
> Node 0, zone    Movable
923c923
< Node 0, zone    Movable
---
> Node 0, zone    Device
```

PR #1818, opened 2023-09-27
pmdabpftrace: rework exit/wait handling to avoid errors

natoscott:
Resolves an issue Ken noticed in the bpftrace PMDA during shutdown - reproducible via regression test qa/1700 ...
```
    File "/usr/libexec/pcp/pmdas/bpftrace/bpftrace/process_manager.py", line 84, in stop_bpftrace_process
     _done, pending = await asyncio.wait({process.wait()}, timeout=5)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/usr/lib64/python3.11/asyncio/tasks.py", line 415, in wait
     raise TypeError("Passing coroutines is forbidden, use tasks explicitly.")
TypeError: Passing coroutines is forbidden, use tasks explicitly.
```

### DIFF
--- a/src/pmdas/bpftrace/bpftrace/process_manager.py
+++ b/src/pmdas/bpftrace/bpftrace/process_manager.py
@@ -81,14 +81,18 @@ class ProcessManager():
         process.send_signal(signal.SIGINT)
 
         # wait max. 5s for graceful termination of the bpftrace process
-        _done, pending = await asyncio.wait({process.wait()}, timeout=5)
-        if pending:
+        try:
+            terminate = asyncio.create_task(process.wait())
+            await asyncio.wait_for(terminate, timeout=5)
+        except asyncio.TimeoutError:
             self.logger.info(f"stop: {script} is still running, sending SIGKILL...")
             process.kill()
 
             # wait again max. 5s until bpftrace process is terminated
-            _done, pending = await asyncio.wait({process.wait()}, timeout=5)
-            if pending:
+            try:
+                terminate = asyncio.create_task(process.wait())
+                await asyncio.wait_for(terminate, timeout=5)
+            except asyncio.TimeoutError:
                 self.logger.info(f"stop: {script} is still running after sending SIGKILL...")
 
         # stopping state change in run_bpftrace task (script can also stop itself, without getting SIGINT)


src/pmdas/bpftrace/bpftrace/process_manager.py
```diff
@@ -81,14 +81,16 @@ async def stop_bpftrace_process(self, script: Script, script_tasks: ScriptTasks)
         process.send_signal(signal.SIGINT)
 
         # wait max. 5s for graceful termination of the bpftrace process
-        _done, pending = await asyncio.wait({process.wait()}, timeout=5)
```
andreasgerstmayr, line 84: `process.wait()` returns a coroutine: https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.subprocess.Process.wait

The error reads "Passing coroutines is forbidden, use tasks explicitly." (I assume this is a new rule in some recent Python version), can you try wrapping the coroutine in a `Task`?

```
_done, pending = await asyncio.wait({asyncio.create_task(process.wait())}, timeout=5)
```

src/pmdas/bpftrace/bpftrace/process_manager.py
```diff
@@ -81,14 +81,16 @@ async def stop_bpftrace_process(self, script: Script, script_tasks: ScriptTasks)
...
-        if pending:
+        try:
+            await asyncio.wait_for(process.wait(), timeout=5)
+        except TimeoutExpired:
```
andreasgerstmayr, line 86: Afaics this exception is only thrown with the `subprocess` module (https://docs.python.org/3/library/subprocess.html#subprocess.TimeoutExpired) but not with the async version of it (`asyncio.subprocess.Process`, https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.subprocess.Process.wait).

andreasgerstmayr (commented):
looks good!

src/pmdas/bpftrace/bpftrace/process_manager.py
natoscott: Good call - something like the last update?

src/pmdas/bpftrace/bpftrace/process_manager.py
natoscott: Yep, thanks - after further reading I've switched over to what I believe is the correct exception now.

kmcdonell (approved):
As you all know, since this is Python I have nothing good, nor informed to say about the code change per se ...  but lots to say about why this change is forced upon us by the bozos who can't even spell compatibility.
But, I extracted the patch from Nathan's last commit and applied this to vm03 (F38) where qa/1700 now passes, and so do the previously failing qa/1701, qa/1703 and qa/1723 tests.
I still see failures for qa/1705 and qa/1724, but these seem unrelated to the code changes in this PR.

natoscott:
@kmcdonell I see the issue in qa/1705 - looks like an bpftrace update has changed the exit status in certain conditions (fixed our test to handle both cases).  Both qa/1723 and qa/1724 are passing for me on f38 though (the whole pmda.bpftrace group is passing here).

PR #1820, opened 2023-10-04
Allow pcp-pmproxy read and write to the io_uring api

zpytela:
With the 65b9e0bdceb ("Implement proper anon_inode support") selinux-policy commit, individual SELinux types are used for each anon_inode type, so domains using any of them need corresponding allow rules.

### DIFF
--- a/src/selinux/pcp.te
+++ b/src/selinux/pcp.te
@@ -225,6 +225,9 @@ optional_policy(`
     allow pcp_pmproxy_t self:io_uring { sqpoll };
 ')
 
+ifdef(`kernel_io_uring_use',`
+	kernel_io_uring_use(pcp_pmproxy_t)
+')
 kernel_search_network_sysctl(pcp_pmproxy_t)
 
 logging_send_syslog_msg(pcp_pmproxy_t)


zpytela:
Please note it requires https://github.com/fedora-selinux/selinux-policy/pull/1889 be merged first. These changes are rawhide only.

zpytela:
This error is currently expected:
```
pcp.te:228:ERROR 'syntax error' at token 'kernel_io_uring_use' on line 12642:
```

natoscott:
> This error is currently expected:
> 
> ```
> pcp.te:228:ERROR 'syntax error' at token 'kernel_io_uring_use' on line 12642:
> ```

We'll need to find a way to make this co-exist with older platforms though (we can't break existing installations when we next upgrade PCP).

zpytela:
> > This error is currently expected:
> > ```
> > pcp.te:228:ERROR 'syntax error' at token 'kernel_io_uring_use' on line 12642:
> > ```
> 
> We'll need to find a way to make this co-exist with older platforms though (we can't break existing installations when we next upgrade PCP).
I'll wrap it with ifdef.

zpytela:
The latest rawhide now contains the changes, so this PR should be merged into pcp.
For other releases, ifdef is used not to make the build fail.

zpytela:
FYI the selinux-policy rawhide build is now available at https://bodhi.fedoraproject.org/updates/FEDORA-2023-5be6d65113

Issue #1706, opened 2022-11-04
pmseries --load overwrites previously loaded archive


vbendel:
So for example I have more pmlogger archives from another non-local host (locally downloaded, nothing connected online), ex.:
~~~
Downloads]$ ls pcp/pmlogger/my_favourite_hostname/
20221017.0.xz  20221017.2.xz  20221017.4.xz  20221017.6.xz  20221017.8.xz   20221017.meta.xz  20221018.1.xz  20221018.3.xz  20221018.5.xz  20221018.7.xz  20221018.index
20221017.1.xz  20221017.3.xz  20221017.5.xz  20221017.7.xz  20221017.index  20221018.0.xz     20221018.2.xz  20221018.4.xz  20221018.6.xz  20221018.8.xz  20221018.meta.xz
~~~

What I want to do is use `pmseries --load` to load both archives into redisDB and see it in via Grafana.

Loading one archive (ex. `pmseries -L 20221017.index`) works as expected, but when I load the second one afterwards (ex. `pmseries -L 20221018.index`), it seems it's dropping the first archive (20221017).. or at least I don't see the 20221017 data in dashboards anymore, while 20221018 are shown as expected.

Following the 'load' routine from `pmseries.c` I think it happens here:
~~~
[EDIT: This is not the culprit (tested custom build)

main@src/pmseries.c
~~~> pmseries_execute() ~~~>
on_series_setup@src/pmseries.c ->
  series_load@src/pmseries.c ->
    pmSeriesLoad@src/libpcp_web/src/query_parser.y ->
      series_load@src/libpcp_web/src/load.c ->
        series_string_mapping@src/libpcp_web/src/load.c ->
          series_string_mapping@src/libpcp_web/src/load.c ->

void
redisGetMap(redisSlots *slots, redisMap *mapping, unsigned char *hash, sds mapStr,
        redisDoneCallBack on_done, redisInfoCallBack on_info,
        void *userdata, void *arg)
{   
    redisMapBaton   *baton;
    redisMapEntry   *entry;
    sds         mapKey;

    pmwebapi_string_hash(hash, mapStr, sdslen(mapStr));
    mapKey = sdsnewlen(hash, 20);

    if ((entry = redisMapLookup(mapping, mapKey)) != NULL) {    <<<<----
    sdsfree(mapKey);
    on_done(arg);
    } else {
    /*
     * This string is not cached locally; so we always send it to server;
     * it may or may not exist there yet, we must just make sure it does.
[...]

PR #1822, opened 2023-10-06
libpcp_web: implement config override by canonical env vars

vbendel:
======== Commit message ========
~~~
Extending utility of PCP_* environment variables to be generally
usable to override settings detected via libpcp_web config.c

Introduces canonical form of config options as:
    "PCP_<SECTION|GROUP>_<OPTION|KEY>"
where dots '.' are represented as underscores '_'.

Example usage:
    $ export PCP_PMSERIERS_STREAM_MAXLEN=424242
    $ pmseries --load <archive> -Dlibweb
    ...
    pmIniFileParse set pmseries.stream.maxlen = 424242

Primary motivation was to create an easy and convenient way to
to adjust configuration inside the archive-analysis container.
~~~
========= Extra info ===========

The archive analysis container uses `pmseries --load` to add the archive data into the Redis DB. However it always passes the optional MAXLEN parameter to Redis XADD command (see `src/libpcp_web/src/schema.c`), which limits the number of entries per loaded metric.

This can be inconvenient when loading more than 1 day worth of PCP data (for example all-default pmlogger archives).

The used MAXLEN parameter passed to Redis XADD is taken from pmproxy.conf file as `stream.maxlen`. (default = 86400 ~= 1 day)

======== PR content notes ========

This PR contains two commits:
1) Note down this limitation to pmseries(1) man page (I spent many hours investigating source code to figure this out - ref https://github.com/performancecopilot/pcp/issues/1706 - while such a simple note in man page would have saved all that time)
2) Enchance libpcp_web config detection to check for canonical ENV variables that override set configuration. (see commit message)

### DIFF
--- a/man/man1/pmseries.1
+++ b/man/man1/pmseries.1
@@ -721,6 +721,13 @@ then a short-hand form can be used:
 .SAMPLE
 $ pmseries --load $PCP_LOG_DIR/pmlogger/acme.0
 .ESAMPLE
+.PP
+NOTE: Timeseries loading is
+.B append-only
+(timestamp-wise) and if more than
+.B stream.maxlen
+entries (defined in /etc/pcp/pmproxy/pmproxy.conf) are loaded for
+a given metric, the oldest entries are dropped.
 .SH OPTIONS
 The available command line options, in addition to timeseries
 metadata and sources options described above, are:
--- a/src/libpcp_web/src/config.c
+++ b/src/libpcp_web/src/config.c
@@ -15,6 +15,7 @@
 #include "libpcp.h"
 #include "util.h"
 #include <inih/ini.h>
+#include <ctype.h>
 
 int
 pmIniFileParse(const char *progname, ini_handler handler, void *data)
@@ -53,12 +54,62 @@ pmIniFileParse(const char *progname, ini_handler handler, void *data)
     return 0;
 }
 
+char *
+construct_canonic_envvar_name(const char *group, const char *key)
+{
+    int i;
+    size_t section_length = strlen(group);
+    size_t option_length = strlen(key);
+    char section[section_length];
+    char option[option_length];
+    char *tmp;
+
+    /* Canonical form: "PCP_<SECTION>_<OPTION>\0" */
+    int envvar_name_len = 4 + section_length + 1 + option_length + 1;
+    char *envvar_name = malloc((size_t)envvar_name_len*sizeof(char));
+    if (!envvar_name)
+	return NULL;
+
+    for (i = 0; i < section_length; i++)
+	section[i] = toupper(group[i]);
+
+    for (i = 0; i < option_length; i++)
+	if (key[i] == '.')
+		option[i] = '_';
+	else
+		option[i] = toupper(key[i]);
+
+    tmp = memcpy(envvar_name, "PCP_", 4) + 4;
+    tmp = memcpy(tmp, section, section_length) + section_length;
+    tmp = memcpy(tmp, "_", 1) + 1;
+    tmp = memcpy(tmp, option, option_length);
+    envvar_name[envvar_name_len - 1] = '\0';
+
+    return envvar_name;
+}
+
+const char *
+check_envvar_override(const char *group, const char *key, const char *value)
+{
+    char *tmp_value;
+    char *envvar_name = construct_canonic_envvar_name(group, key);
+    if (!envvar_name)
+	return value;
+
+    if ((tmp_value = getenv(envvar_name)) != NULL)
+	value = tmp_value;
+
+    free(envvar_name);
+    return value;
+}
+
 static int
 dict_handler(void *arg, const char *group, const char *key, const char *value)
 {
     dict	*config = (dict *)arg;
     sds		name = sdsempty();
 
+    value = check_envvar_override(group, key, value);
     name = sdscatfmt(name, "%s.%s", group ? group : pmGetProgname(), key);
     if (pmDebugOptions.libweb)
 	fprintf(stderr, "pmIniFileParse set %s = %s\n", name, value);


build/containers/archive-analysis/README.md
```diff
@@ -15,6 +15,8 @@ $ podman run \
 
 This command starts the container, which runs Redis and Grafana (inside the container) and loads all PCP archives of the selected directory on the host into Redis. You can point your browser to `http://localhost:3000/d/pcp-archive-analysis/pcp-archive-analysis` and can start inspecting the archives using Grafana. 
 
+Optionally add `-e STREAM_MAXLEN_ADJUST=<value>` (default is 86400) to `podman run` to adjust the _stream.maxlen_ parameter in order to increase the number of loaded timeseries entries. Useful when loading archives covering more than one day.
+
```
natoscott, line 19: I'd prefer if we could make this a more general solution, rather than for just one variable.

I suggest we name this variable PMSERIES_STREAM_MAXLEN and also allow PMSERIES_STREAM_EXPIRE to be set.

build/containers/archive-analysis/root/etc/systemd/system/archive-import.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=PCP archives import
 Requires=redis.service
-After=redis.service
+After=redis.service stream_maxlen_config.service
```
natoscott, line 4: So, perhaps rename this to more like pmseries-config.service  (hyphens seem to be more common in systemd?)

build/containers/archive-analysis/root/etc/systemd/system/stream_maxlen_config.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Adjust pmproxy.conf stream.maxlen parameter
+Before=pmproxy.service
+
+[Service]
+Type=oneshot
+PassEnvironment=STREAM_MAXLEN_ADJUST
+ExecStart=bash -c 'if [ ! -z $STREAM_MAXLEN_ADJUST ]; then sed -i "s/stream.maxlen = [0-9]*/stream.maxlen = $STREAM_MAXLEN_ADJUST/" /etc/pcp/pmproxy/pmproxy.conf; fi'
```
natoscott, line 8: I think this will quickly get out of hand done inline in the service file - let's move this to a pmseries-config.sh script and call that.

natoscott: In fact, considering further - how about we make the tools aware of the PMSERIES_ environment variables directly?  (i.e. override any config file setting with an env var setting).  Then we don't need this additional service configuration systemd magic at all, everything would be simpler I think.

natoscott (changes requested):
@vbendel thanks for looking into this.  Some suggestions inline.

build/containers/archive-analysis/README.md
vbendel: Well, a problem is that the container is somewhat meant to be immutable. As you can see it's not too convenient to even simply export a dynamic modification of a file. So not sure how much generic you'd wanna go :D , perhaps it would be easier to supply a custom pmproxy.conf file as a whole instead then - but a question arises when it's easier to simply bash into the container and modify & service-restart it manually inside the container.

We can rename, sure, although I'm not sure if we explicitly need to adjust the "expire" config. IIUC that's pretty long by default anyway.

TBH I think it's a question about what usecases is this container for - I personally use it only for quick one-off inspection of pcp archives, for which I only encounter the limitation for number of entries (stream.maxlen) so far. I think that's it's primary meant use case and I'm not sure we want to encourage any other use. (i.e. if a user needs anything more than this container offers, then I'd say a better way is to set up the environment in OS/VM and optionally use pmseries directly)

build/containers/archive-analysis/root/etc/systemd/system/stream_maxlen_config.service
natoscott: Right - we have a well-established convention in several PCP tools where a configuration hierarchy exists.  Command line option overrides are highest priority, then environment variables, then configuration files, then hard-coded default fallback.

My suggestion is to add awareness of PMSERIES_STREAM_MAXLEN and PMSERIES_STREAM_EXPIRE to the libpcp_web library used by pmseries and pmproxy such that these environment variables overrides the default configuration file settings, following this convention.  Since the simplest docker/podman interface seems to be passing in environment variables, this will require no special case code in the Dockerfile - it will Just Work because the tools themselves respect those variables now.

I'm happy to write the C code if it helps.  It seems like a good fit for existing PCP conventions and simplifies the Dockerfile for long-term maintenance IMO.  The downside is users will need to wait for a new PCP release to use it, but I think its best to go for the longer-term cleaner-code solution.

andreasgerstmayr: +1, I think a generic way to override configuration options from the config file with env vars is a good solution for this issue. Grafana has the same feature: https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#override-configuration-with-environment-variables

In the pcp container I used the `envsubst(1)` utility to override values in the configuration file:
https://github.com/performancecopilot/pcp/blob/e9530dc38827ae31db6bdad0b8012dd1bf4865b9/build/containers/pcp/root/usr/bin/container-entrypoint#L13
https://github.com/performancecopilot/pcp/blob/e9530dc38827ae31db6bdad0b8012dd1bf4865b9/build/containers/pcp/root/usr/share/container-scripts/pcp/pmproxy.conf.template#L74
In this case, the default value moved from the configuration file to the bashscript: https://github.com/performancecopilot/pcp/blob/e9530dc38827ae31db6bdad0b8012dd1bf4865b9/build/containers/pcp/root/usr/bin/container-entrypoint#L7

kurik: Having the ability to override config variables using PMSERIES_* variables as proposed by @natoscott and @andreasgerstmayr is definitely +1 for me. As Andreas already mentioned, we have this in the Grafana container, where this override is tested by some automated tests. So, having this implemented in the same way as we do it in Grafana will be beneficial for QA as we do not need to develop anything new from the testing point of view.

vbendel: OK, since PCP has a defined hierarchy like that (I didn't know :)), then it makes sense. Since you asked _as generic as possible_, I'd suggest something like this:
~~~
pcp]$ git diff
diff --git a/src/libpcp_web/src/config.c b/src/libpcp_web/src/config.c
index 816bc8813..1dafa0492 100644
--- a/src/libpcp_web/src/config.c
+++ b/src/libpcp_web/src/config.c
@@ -53,12 +53,60 @@ pmIniFileParse(const char *progname, ini_handler handler, void *data)
     return 0;
 }
 
+char *
+construct_canonic_envvar_name(const char *group, const char *key)
+{
+    int i;
+    size_t section_length = strlen(group);
+    size_t option_length = strlen(key);
+    char section[section_length];
+    char option[option_length];
+    char *tmp;
+
+    /* Canonical form: "PCP_<SECTION>_<OPTION>\0" */
+    int envvar_name_len = 4 + section_length + 1 + option_length + 1;
+    char *envvar_name = malloc((size_t)envvar_name_len*sizeof(char));
+    if (!envvar_name)
+       return NULL;
+
+    for (i = 0; i < section_length; i++)
+       section[i] = toupper(group[i]);
+    for (i = 0; i < option_length; i++)
+       if (key[i] == '.')
+               option[i] = '_';
+       else
+               option[i] = toupper(key[i]);
+
+    tmp = mempcpy(envvar_name, "PCP_", 4);
+    tmp = mempcpy(tmp, section, section_length);
+    tmp = mempcpy(tmp, "_", 1);
+    tmp = mempcpy(tmp, option, option_length);
+    envvar_name[envvar_name_len] = '\0';
+
+    return envvar_name;
+}
+
+const char *
+check_envvar_override(const char *group, const char *key, const char *value)
+{
+    char *tmp_value;
+    char *envvar_name = construct_canonic_envvar_name(group, key);
+    if (!envvar_name)
+       goto out_ENOMEM;
+
+    if ((tmp_value = getenv(envvar_name)) != NULL)
+       value = tmp_value;
+
+    free(envvar_name);
+
+out_ENOMEM:
+    return value;
+}
+
 static int
 dict_handler(void *arg, const char *group, const char *key, const char *value)
 {
     dict       *config = (dict *)arg;
     sds                name = sdsempty();
 
+    value = check_envvar_override(group, key, value);
     name = sdscatfmt(name, "%s.%s", group ? group : pmGetProgname(), key);
     if (pmDebugOptions.libweb)
        fprintf(stderr, "pmIniFileParse set %s = %s\n", name, value);
~~~
The "hook" point is in dict_handler(), that is performing the actual add to config dictionary dudring parsing of the config file.. call chain roughly: 
~~~
main@src/pmseries/pmseries.c
-> pmIniFileParse@src/libpcp_web/src/config.c
  -> ini_parse@vendor/github.com/benhoyt/inih/ini.c
    -> ini_parse_file@vendor/github.com/benhoyt/inih/ini.c
      -> ini_parse_stream@vendor/github.com/benhoyt/inih/ini.c
        -> handler == dict_handler@src/libpcp_web/src/config.c
~~~
As part of the handler it checks if an environment variable for the particular config exists and if so, then overrides it from the env var. 
It uses the `group` and `key` to construct the canonical environment variable name of form `PCP_<SECTION>_<OPTION>`, where dots `.` in the option name are represented as underscores `_`.

A sort of unit test of the envvar_name string construction (putting the code into a dedicated `test.c`):
~~~
//    char group[] = "section";
//    char key[] = "option.suboption";

]$ ./test
section len = 7
option len = 16
envvar_name=PCP_SECTION_OPTION_SUBOPTION, len=28
~~~

I know that particular bits, namely the canonic envvar string constuction, ought to be put into some different source file (possibly some header), but I wanted to get your opinion on the approach first. Aesthetics can be done later.

Also, I have yet to try building it from source and testing for PCP actually - will do that tomorrow. Not sure if there are relevant (though still generically standard) libraries already included from some header around, or would be needed to add, namely <ctype.h> for `toupper()` and also `#define _GNU_SOURCE` for `mempcpy()` (the latter can be changed to more generic memcpy, I just went with the more convenient one for now).

Looking forward to your thoughts on the approach. :)

natoscott: | Looking forward to your thoughts on the approach. :)

That's a neat solution - I like it.  I wouldn't lose any sleep over a wasted byte.  :)

mempcpy is a GNU extension, we'll need to adjust that part because PCP is cross-platform - either just use standard interfaces or via autoconf-based ifdef guards and a fallback (the first approach likely simpler).

I also recommend dropping the out_ENOMEM goto, in this case (IMO) the jump adds little value in terms of readability or code simplification - we can just return directly there on failure instead of using goto.

vbendel: So it builds and runs, however it doesn't seem to catch the override (tried direct manual test also). I tried adding some debug prints (even simple printf()), but it seems I misunderstand the callchain and am hooking into the wrong function (my debug prints in `dict_handler@src/libpcp_web/src/config.c` are not even printed ??).

Could you perhaps provide some hint what I'm missing -- see the dev branch for commit https://github.com/vbendel/pcp/tree/envvar_override ? 
Optionally how to work with the `pmDebugOptions.libweb` or other debug options? (I see pmSetConfig() C functions - are those the only interface to enable the debug points?)

natoscott: @vbendel yes, libpcp_web is a shared library - your code does appear to be invoked:

```
$ export LD_LIBRARY_PATH=`pwd`/../libpcp_web/src
$ ./pmseries hinv.ncpu
DEBUG: group=pmproxy key=pcp.enabled value=true
DEBUG: group=pmproxy key=http.enabled value=true
DEBUG: group=pmproxy key=redis.enabled value=true
DEBUG: group=pmproxy key=secure.enabled value=true
DEBUG: group=redis key=enabled value=true
DEBUG: group=redis key=servers value=localhost:6379
DEBUG: group=discover key=enabled value=true
DEBUG: group=discover key=exclude.metrics value=proc.*,acct.*
DEBUG: group=discover key=exclude.indoms value=3.9,3.40,79.7
DEBUG: group=pmsearch key=enabled value=true
DEBUG: group=pmsearch key=count value=10
DEBUG: group=pmseries key=enabled value=true
DEBUG: group=pmseries key=cursor.count value=256
DEBUG: group=pmseries key=stream.expire value=86400
DEBUG: group=pmseries key=stream.maxlen value=8640
6b01fd0761645cbccb6312e0cf3bb9c838dfe7af
a55d8e2a7a7ef7e0f70e8ab6b1ce767cd3a0ce0b
cbf17c2f8421d6a1c96b64dc582a58ef177bccbe
```

You can access the debugging flags through the -D/--debug option available with all PCP binaries, e.g.

```
$ ./pmseries -Dlibweb hinv.ncpu
DEBUG: group=pmproxy key=pcp.enabled value=true
pmIniFileParse set pmproxy.pcp.enabled = true
DEBUG: group=pmproxy key=http.enabled value=true
pmIniFileParse set pmproxy.http.enabled = true
DEBUG: group=pmproxy key=redis.enabled value=true
pmIniFileParse set pmproxy.redis.enabled = true
DEBUG: group=pmproxy key=secure.enabled value=true
pmIniFileParse set pmproxy.secure.enabled = true
DEBUG: group=redis key=enabled value=true
pmIniFileParse set redis.enabled = true
DEBUG: group=redis key=servers value=localhost:6379
pmIniFileParse set redis.servers = localhost:6379
DEBUG: group=discover key=enabled value=true
pmIniFileParse set discover.enabled = true
DEBUG: group=discover key=exclude.metrics value=proc.*,acct.*
pmIniFileParse set discover.exclude.metrics = proc.*,acct.*
DEBUG: group=discover key=exclude.indoms value=3.9,3.40,79.7
pmIniFileParse set discover.exclude.indoms = 3.9,3.40,79.7
DEBUG: group=pmsearch key=enabled value=true
pmIniFileParse set pmsearch.enabled = true
DEBUG: group=pmsearch key=count value=10
pmIniFileParse set pmsearch.count = 10
DEBUG: group=pmseries key=enabled value=true
pmIniFileParse set pmseries.enabled = true
DEBUG: group=pmseries key=cursor.count value=256
pmIniFileParse set pmseries.cursor.count = 256
DEBUG: group=pmseries key=stream.expire value=86400
pmIniFileParse set pmseries.stream.expire = 86400
DEBUG: group=pmseries key=stream.maxlen value=8640
pmIniFileParse set pmseries.stream.maxlen = 8640
6b01fd0761645cbccb6312e0cf3bb9c838dfe7af
a55d8e2a7a7ef7e0f70e8ab6b1ce767cd3a0ce0b
cbf17c2f8421d6a1c96b64dc582a58ef177bccbe
```

pmdbg -l/--list for the full list - multiple debug settings can be used too via comma-separated list.

vbendel: Thank you for the hints!

I seems to be working:
~~~
[root@localhost pcp]# export PCP_PMSEARCH_COUNT=12
[root@localhost pcp]# export PCP_PMSERIES_STREAM_MAXLEN=456789
[root@localhost pcp]# ./src/pmseries/pmseries -Dlibweb --load <archive> 
pmIniFileParse set pmproxy.pcp.enabled = true
pmIniFileParse set pmproxy.http.enabled = true
pmIniFileParse set pmproxy.redis.enabled = true
pmIniFileParse set pmproxy.secure.enabled = true
pmIniFileParse set redis.enabled = true
pmIniFileParse set redis.servers = localhost:6379
pmIniFileParse set discover.enabled = true
pmIniFileParse set discover.exclude.metrics = proc.*,acct.*
pmIniFileParse set discover.exclude.indoms = 3.9,3.40,79.7
pmIniFileParse set pmsearch.enabled = true
pmIniFileParse set pmsearch.count = 12   <<<---
pmIniFileParse set pmseries.enabled = true
pmIniFileParse set pmseries.cursor.count = 256
pmIniFileParse set pmseries.stream.expire = 86400
pmIniFileParse set pmseries.stream.maxlen = 456789  <<<---
~~~

Also tested in practice and can confirm the records are not overwritten in Redis DB, i.e. the updated XADD MAXLEN param is indeed properly used.

Now the question is where to put the `construct_canonic_envvar_name()` function (needs also `<ctype.h>`)? Should we:
- Create a `src/libpcp_web/src/config.h` (only config.c is there atm); or
- Put it somewhere else, like general `libpcp.h` for example (or other suggestions); or
- Leave it simply as a "local" function in `src/libpcp_web/src/config.c`, as depicted right now?
(Personally I'd create a new `src/libpcp_web/src/config.h` for it.)

Looking forward to your suggestions. Afterwards I'll take your feedback and update this PR overall.

vbendel:
Reopening this after PR update. Continuing from last comment - I decided to push it without creation of extra .h files, as it seems unnecessary.

man/man1/pmseries.1
```diff
@@ -721,6 +721,13 @@ then a short-hand form can be used:
 .SAMPLE
 $ pmseries --load $PCP_LOG_DIR/pmlogger/acme.0
 .ESAMPLE
+.PP
+NOTE: Timeseries loading is
+.B append-only
+(timestamp-wise) and if more than
+.B stream.maxlen
+entries (defined in /etc/pcp/pmproxy/pmproxy.conf) are loaded for
```
natoscott, line 729: This reference to /etc/pcp needs to be via PCP_SYSCONF_DIR as it may differ between platforms - I'll fix that up.

natoscott:
@vbendel when Coverity scan ran on this it produced the following warnings - can you take a look?  Thanks!

```
3 new defect(s) introduced to performancecopilot/pcp found with Coverity Scan.
3 defect(s), reported by Coverity Scan earlier, were marked fixed in the recent build analyzed by Coverity Scan.

New defect(s) Reported-by: Coverity Scan
Showing 3 of 3 defect(s)


** CID 404740:  Null pointer dereferences  (REVERSE_INULL)
/src/libpcp_web/src/config.c: 113 in dict_handler()


________________________________________________________________________________________________________
*** CID 404740:  Null pointer dereferences  (REVERSE_INULL)
/src/libpcp_web/src/config.c: 113 in dict_handler()
107     dict_handler(void *arg, const char *group, const char *key, const char *value)
108     {
109         dict        *config = (dict *)arg;
110         sds         name = sdsempty();
111     
112         value = check_envvar_override(group, key, value);
>>>     CID 404740:  Null pointer dereferences  (REVERSE_INULL)
>>>     Null-checking "group" suggests that it may be null, but it has already been dereferenced on all paths leading to the check.
113         name = sdscatfmt(name, "%s.%s", group ? group : pmGetProgname(), key);
114         if (pmDebugOptions.libweb)
115             fprintf(stderr, "pmIniFileParse set %s = %s\n", name, value);
116         return dictReplace(config, name, sdsnew(value)) != DICT_OK;
117     }
118     

** CID 404739:  Memory - corruptions  (OVERRUN)
/src/libpcp_web/src/config.c: 64 in construct_canonic_envvar_name()


________________________________________________________________________________________________________
*** CID 404739:  Memory - corruptions  (OVERRUN)
/src/libpcp_web/src/config.c: 64 in construct_canonic_envvar_name()
58     construct_canonic_envvar_name(const char *group, const char *key)
59     {
60         int i;
61         size_t section_length = strlen(group);
62         size_t option_length = strlen(key);
63         char section[section_length];
>>>     CID 404739:  Memory - corruptions  (OVERRUN)
>>>     Allocating insufficient memory for the terminating null of the string.
64         char option[option_length];
65         char *tmp;
66     
67         /* Canonical form: "PCP_<SECTION>_<OPTION>\0" */
68         int envvar_name_len = 4 + section_length + 1 + option_length + 1;
69         char *envvar_name = malloc((size_t)envvar_name_len*sizeof(char));

** CID 404738:  Memory - corruptions  (OVERRUN)
/src/libpcp_web/src/config.c: 63 in construct_canonic_envvar_name()


________________________________________________________________________________________________________
*** CID 404738:  Memory - corruptions  (OVERRUN)
/src/libpcp_web/src/config.c: 63 in construct_canonic_envvar_name()
57     char *
58     construct_canonic_envvar_name(const char *group, const char *key)
59     {
60         int i;
61         size_t section_length = strlen(group);
62         size_t option_length = strlen(key);
>>>     CID 404738:  Memory - corruptions  (OVERRUN)
>>>     Allocating insufficient memory for the terminating null of the string.
63         char section[section_length];
64         char option[option_length];
65         char *tmp;
66     
67         /* Canonical form: "PCP_<SECTION>_<OPTION>\0" */
68         int envvar_name_len = 4 + section_length + 1 + option_length + 1;
```

vbendel:
I see. Fixed here: https://github.com/vbendel/pcp/tree/envvar-bugfix
Should I create a new PR? (seem unnecessary overhead)
Or is it possible to somehow amend this PR?

natoscott:
Thanks - nothing else is needed, I've pulled the fix directly from that branch.

PR #1828, opened 2023-10-12
pmda.farm Seagate Field Access Reliability Metrics (FARM) Log

pauljevans:
Add support for parsing and collecting of Seagate's vendor-specific Field Access Reliability Metrics (FARM) log to support ATA drives.

Metric data is collected using "smartctl" through the "-l farm" switch.

FARM expands information available in SMART, providing statistical information for health monitoring and reliability prediction, including metrics which are tracked to a per-surface level.

Additional information can be found in Seagate's [open-source FARM specification](https://github.com/Seagate/openSeaChest_LogParser/blob/develop/docs/FARM_Specification.pdf)


src/pmdas/farm/help
```diff
@@ -0,0 +1,705 @@
...
+
+@farm.ata.error_statistics.cto_count_over_5s CTO count over 5s
+CTO count over 5s for the drive
+
+@farm.ata.error_statistics.cto_count_over_7s CTO count over 7.5s
+CTO count over 7.5s for the drive
+
+@farm.ata.error_statistics.total_flash_led_assert_events Total Flash LED (Assert) events for the drive
+@farm.ata.error_statistics.index_of_last_flash_led Index of the last Flash LED for the drive
+@farm.ata.error_statistics.flash_led_event_0.event_information Flash LED Event 0: Event information
```
natoscott, line 234: Could this instead be a single metric with a flash_led_event instance domain? (with instances event_0 through event_7). This is more efficient (e.g. less ondisk space used by pmlogger) and more flexible if additional events come along later.

src/pmdas/farm/help
```diff
@@ -0,0 +1,705 @@
...
+@farm.ata.error_statistics.flash_led_event_7.event_information Flash LED Event 7: Event information
+@farm.ata.error_statistics.flash_led_event_7.timestamp_of_event Flash LED Event 7: Power Cycle Event
+@farm.ata.error_statistics.flash_led_event_7.power_cycle_event Flash LED Event 7: Power Cycle Event
+@farm.ata.error_statistics.uncorrectable_errors Uncorrectable errors count 
+Number of uncorrectable errors for the drive
+
+@farm.ata.error_statistics.cumulative_lifetime_unrecoverable_errors_due_to_erc Cumulative unrecoverable read errors ERC
+Number of cumulative lifetime unrecoverable read errors due to ERC
+
+@farm.ata.error_statistics.cum_lifetime_unrecoverable_errors_head_0.cumulative_lifetime_unrecoverable_read_repeating Head 0
```
natoscott, line 264: Similarly, this and related metrics seems to me like it should have a 'head' instance domain - fewer metrics this way - its more efficient for client tools, more flexible, easier to maintain longer term.

src/pmdas/farm/help
```diff
@@ -0,0 +1,705 @@
...
+@farm.ata.error_statistics.cum_lifetime_unrecoverable_errors_head_22.cumulative_lifetime_unrecoverable_read_unique Head 22
+Drive head 22: Cumulative lifetime  unrecoverable read unique for the drive
+
+@farm.ata.error_statistics.cum_lifetime_unrecoverable_errors_head_23.cumulative_lifetime_unrecoverable_read_repeating Head 23
+Drive head 23: Cumulative lifetime unrecoverable read repeating for the drive
+
+@farm.ata.error_statistics.cum_lifetime_unrecoverable_errors_head_23.cumulative_lifetime_unrecoverable_read_unique Head 23
+Drive head 23: Cumulative lifetime  unrecoverable read unique for the drive
+
+@farm.ata.environment_statistics.current_temperature Current temperature
```
natoscott, line 407: 'temperature' name repeats alot, consider a sub-namespace for temperature.* metrics here.

src/pmdas/farm/help
```diff
@@ -0,0 +1,705 @@
...
+@farm.ata.reliability_statistics.high_priority_unload_events High priority unload events count
+Number of high priority unload events for the drive
+
+@farm.ata.reliability_statistics.helium_pressure_threshold_tripped Helium pressure threshold trip
+Helium pressure threshold tripped for the drive
+
+@farm.ata.reliability_statistics.lbas_corrected_by_parity_sector LBA corrected by parity sector 
+LBAs corrected by parity sector for the drive
+
+@farm.ata.reliability_statistics.dvga_skip_write_detected_head_0 DVGA skip write detect by head 0 for the drive
```
natoscott, line 490: As above, use an instance domain here for all the different 'head_N' instances of this metric.

src/pmdas/farm/pmda.c
```diff
@@ -0,0 +1,1907 @@
...
+ #include "pmda.h"
+ #include "domain.h"
+ 
+ #include "pmdafarm.h"
+ 
+ static int _isDSO = 1; /* for local contexts */
+ static char *farm_setup_lsblk;
+ static char *farm_setup_smartctl;
+ 
+ pmdaIndom indomtable[] = {
```
natoscott, line 27: This PMDA will benefit from using additional indoms I think.  It looks like you have a 'compound indom' situation here - we tend to solve this using '::' separator on the external instance names, and use of the pmdaCache APIs.  Have a look at the cgroup.io.stat.rbytes metric as an example, which combines cgroups and devices, e.g.

```
pminfo -f cgroup.io.stat.rbytes
    inst [...]
    inst [34 or "/system.slice::vda"] value 75020690432
```

These metrics also have labels for querying the individual parts of the instance, e.g.

```
pminfo -l cgroup.io.stat.rbytes
    [...]
    inst [34 or "/system.slice::vda"] labels {"agent":"proc","cgroup":"/system.slice","device_name":"vda","device_type":"block","domainname":"...","groupid":1000,"hostname":"fedora","machineid":"...","userid":1000}
```

src/pmdas/farm/pmda.c
```diff
@@ -0,0 +1,1907 @@
...
+	{ .m_desc = {
+		PMDA_PMID(CLUSTER_ATA_LOG_HEADER, LOG_VERSION),
+		PM_TYPE_STRING, DISK_INDOM, PM_SEM_DISCRETE,
+		PMDA_PMUNITS(0,0,0,0,0,0) }, },
+	{ .m_desc = {
+		PMDA_PMID(CLUSTER_ATA_LOG_HEADER, PAGES_SUPPORTED),
+		PM_TYPE_U64, DISK_INDOM, PM_SEM_DISCRETE,
+		PMDA_PMUNITS(0,0,0,0,0,0) }, },
+	{ .m_desc = {
+		PMDA_PMID(CLUSTER_ATA_LOG_HEADER, LOG_SIZE),
```
natoscott, line 47: Once the instance domains are tackled, this metric table will shrink significantly.
Next issue to tackle then (easier with fewer metrics) is the units and semantics need to be reviewed and update - I'm seeing lots of things that are bytes, times, etc here.  These need correct units and the semantics are either PM_SEM_COUNTER or PM_SEM_INSTANT for most metrics here I think.

src/pmdas/farm/pmda.c
```diff
@@ -0,0 +1,1907 @@
...
+{
+	int nindoms = sizeof(indomtable)/sizeof(indomtable[0]);
+	int nmetrics = sizeof(metrictable)/sizeof(metrictable[0]);
+
+	if (_isDSO) {
+		char helppath[MAXPATHLEN];
+		int sep = pmPathSeparator();
+		pmsprintf(helppath, sizeof(helppath), "%s%c" "farm" "%c" "help",
+			pmGetConfig("PCP_PMDAS_DIR"), sep, sep);
+		pmdaDSO(dp, PMDA_INTERFACE_4, "FARM DSO", helppath);
```
natoscott, line 1846: We need to use labels, so we'll want to shuffle this up to PMDA_INTERFACE_7 (here and below)

src/pmdas/farm/pmda.c
```diff
@@ -0,0 +1,1907 @@
...
+	}
+
+	if (dp->status != 0)
+		return;
+
+	/* Check for environment variables allowing test injection */
+	farm_instance_setup();
+	farm_stats_setup();
+
+	dp->version.four.instance = farm_instance;
```
natoscott, line 1856: this becomes version.seven (here and below) - each function is backwards compatible, each new version adds new optional callbacks.

src/pmdas/farm/pmda.c
```diff
@@ -0,0 +1,1907 @@
...
+	dp->version.four.fetch = farm_fetch;
+	dp->version.four.text = farm_text;
```
natoscott, line 1858: We don't use dynamic metrics in this PMDA, so I don't think we need farm_text, farm_pmid, farm_name or farm_children callbacks, the libpcp_pmda defaults should do the trick.  If this was based on the Linux agent (?) this is a historical quirk - should be removed there too I think (I'll look into that) - we used to have dynamic interrupts metrics there, but no longer do.

We'll need new label callbacks instead - 
```
    dp->version.seven.label = farm_label;
    pmdaSetLabelCallBack(dp, farm_labelCallBack);
```

See the Linux agent for a reference and pmLookupLabels(3).

src/pmdas/farm/pmda.c
```diff
@@ -0,0 +1,1907 @@
...
+{
+	int sep = pmPathSeparator();
+	char helppath[MAXPATHLEN];
+	pmdaInterface dispatch;
+
+	_isDSO = 0;
+	pmSetProgname(argv[0]);
+	pmsprintf(helppath, sizeof(helppath), "%s%c" "farm" "%c" "help",
+		pmGetConfig("PCP_PMDAS_DIR"), sep, sep);
+	pmdaDaemon(&dispatch, PMDA_INTERFACE_4, pmGetProgname(), FARM, "farm.log", helppath);
```
natoscott, line 1893: 4 -> 7 interface version.

natoscott (changes requested):
@pauljevans nice!  Few comments inline - in particular, the use of a compound instance domain or two should simplify this agent alot & vastly reduce the number of metrics.

natoscott:
@pauljevans please also add a src/pmdas/farm/.gitignore file containing:

```
.gitignore 
domain.h
exports
pmda_farm.so
pmdafarm
```

natoscott:
WRT the CI failures, @sfeifer is fixing the debian12 issue, and the debian10 failure is a spurious condition, will likely pass next time it runs - these are unrelated to your code.

natoscott:
Oh, some other examples of compound instance domains are kernel.percpu.interrupts and kernel.percpu.softirqs if you're looking at the Linux kernel agent (rather than linux_proc where the cgroups metrics reside).

PR #1831, opened 2023-10-18
Update libpcp_archive.pc.in

bryteise:
The Name and Libs fields reference lib_arch but the filename of the the library is libpcp_archive. Correct the fields to match the filename


kmcdonell (approved):
LGTM.
Thanks for catching this @bryteise, and welcome.

PR #1846, opened 2023-11-09
pmproxy: support http compression

### LINKED ISSUE #1685: pmproxy: support HTTP compression
pmproxy should be able to send responses in a compressed format if the client supports it

https://github.com/performancecopilot/pcp/pull/1682#issuecomment-1266096325

lmchilton:
Add http compression to the REST API for grafana-pcp where large transfers have been observed with pmseries queries. This implementation supports zlib compression only so far. Additionally, several new http metrics are added to pmproxy showing compression effectiveness.

Resolves #1685

### DIFF
--- a/qa/1543
+++ b/qa/1543
@@ -390,6 +390,10 @@ curl -s "http://localhost:$port/pmapi/metrics?target=sample.long" \
 	| _filter_text "scrape one metric tree"
 curl -s "http://localhost:$port/pmapi/metrics?target=sample.colour" \
 	| _filter_text "scrape metric instances"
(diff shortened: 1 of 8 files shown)


natoscott:
@lmchilton looks good!  There's a few more things I found on review since we last discussed - could you double check these changes for me please?  Taa.  They include ifdef-protected use of all libz functions and types (we missed a few), making compress_GZIP a local static function, fixing the error handling there (we cannot send both an error reply and a regular reply - and we were ignoring the return code anyway - best if we log the error locally and continue on), and several spacing issues - we'll want to get your editor understanding PCP code spacing conventions better if possible.

PR #1873, opened 2024-01-28
Security - rc/init scripts

kmcdonell:
A bunch of changes to provide improved security, especially for PCP's rc/init/systemd scripts
- cleaner separation of parts run as user "root" from parts run as user "pcp"
- when running as user "root" do not modify (mkdir, cp, chmod, rm, ...) objects owned by the user "pcp"
- symmetry of logic for non-root scripts when run with systemd or init or any-other-replacement


natoscott (commented):
Hi @kmcdonell - its looking good.  I'll add a few small nits inline shortly, but one higher level topic is I think the tmpfiles.d file needs to live elsewhere.  AIUI the /etc location is for sysadmins, and as packagers we are supposed to use the /usr/lib location ... (PCP_SYSTEMDTMPFILES_DIR)

```
$ rpm -ql pcp |grep tmpf
/etc/tmpfiles.d/pcp.conf
/usr/lib/tmpfiles.d/pcp.conf
```

The first is new of course, the latter we generate during the RPM builds (for rpm-ostree).  It'd be ideal if we can wrangle this content to be all in the second file.
I guess that means we need to shift this to be generated in the rpm/deb packaging rather than at this top level.

src/pmie/rc_pmie
```diff
@@ -75,19 +51,6 @@ else
...
-    if [ -n "$PCP_DIR" ]
-    then
-	: running in a non-default installation, do not need to be root
-    else
-	echo "$prog:"'
-Error: You must be root (uid 0) to start or stop pmie.'
-	exit
-    fi
-fi
-
```
natoscott, line 90: As discussed on slack, the restorecon invocation later in this file can go.

man/man1/runaspcp.1
```diff
@@ -0,0 +1,73 @@
...
+from the
+.B \-s
+(or
+.BR \-\-shell)
+command line argument.
+.PP
+Normally
+.B runaspcp
+is used from the PCP ``init'' scripts that are launched
+as ``root'' but need to downgrade their priveleges when running
```
natoscott, line 48: typo - "privileges"

man/man3i/__pmcleanmapdir.3
```diff
@@ -0,0 +1,95 @@
...
+will remove them:
+.IP + 2n
+the name should be numeric, else it should be equal
+.I special
+(if
+.I special
+is not NULL)
+.IP + 2n
+for a numeric directory entry it should be a regular file and
+and the name should be the PID of a running process
```
natoscott, line 61: we have "and and" here (across two lines).

src/libpcp/src/util.c
```diff
@@ -3162,3 +3166,235 @@ __pmLogFeaturesStr(__uint32_t features)
...
+{
+    switch (st_mode & S_IFMT) {
+	case S_IFBLK:	return "block device"; break;
+	case S_IFCHR:	return "character device"; break;
+	case S_IFDIR:	return "directory"; break;
+	case S_IFIFO:	return "FIFO/pipe"; break;
+	case S_IFLNK:	return "symlink"; break;
+	case S_IFREG:	return "regular file"; break;
+	case S_IFSOCK:	return "socket"; break;
+	default:	return "unknown?"; break;
```
natoscott, line 3182: I think most of the "break;" use here (after return) can be dropped?  Personal style, I'd just keep the default: case as a "break;" and return unknown on the final line of the function.

src/pcp-reboot-init/pcp-reboot-init.sh
```diff
@@ -0,0 +1,111 @@
...
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+# for more details.
+#
+# Intended as "run after reboot" to make sure transient things that PCP
+# needs have been setup ...
+#
+
+tmp=/var/tmp/pcp-reboot-init-$$
```
natoscott, line 19: Should this use mktemp here like most PCP scripts do?  Think so.

kmcdonell:
@natoscott re. tmpfiles.d ... we can just place the /run/pcp one in /usr/lib/tmpfiles.d/pcp-run-pcp.conf (avoids name conflict with rpm-generated ones) ... since /usr/lib/tmpfiles.d can exist without systemd, I'd prefer to use the explicit pathname rather than $(PCP_SYSTEMDTMPFILES_DIR) (which may translate to badness on a non-systemd platform).
Also I made it D rather then d in the .conf file so it goes away with package removal.

natoscott:
> @natoscott re. tmpfiles.d ... we can just place the /run/pcp one in /usr/lib/tmpfiles.d/pcp-run-pcp.conf (avoids name conflict with rpm-generated ones) ... since /usr/lib/tmpfiles.d can exist without systemd, I'd prefer to use the explicit pathname rather than $(PCP_SYSTEMDTMPFILES_DIR) (which may translate to badness on a non-systemd platform). Also I made it D rather then d in the .conf file so it goes away with package removal.

@kenj yep, sounds good.  Another option might be to share a name with the pcp-reboot-init service?

kmcdonell:
@natoscott OK, since the tmpfiles.d conf and the pcp-reboot-init script are effectively alternate ways of getting the same job done, I'm happy to rename the conf file.  This part is done in commits  1f28451aae and 97f85835.

src/pmie/rc_pmie
kmcdonell: _reboot_setup() all removed in commit 283b560b1

man/man3i/__pmcleanmapdir.3
kmcdonell: Fixed in commit 7531b6404

man/man1/runaspcp.1
kmcdonell: Fixed in commit 7531b6404

src/libpcp/src/util.c
kmcdonell: Fixed in commit f29a6af00

src/pcp-reboot-init/pcp-reboot-init.sh
kmcdonell: Yep, fixed in commit 35600595c

PR #1875, opened 2024-01-31
pcp2arrow: export PCP metrics to parquet format for pandas

natoscott:


myllynen:
Thanks for working on this, looks very good. Two comments:

1) I think removing the Python int/long workaround which was needed in the past should probably be a separate commit with a comment why it's ok to remove now. It's also touching unrelated tools and may in theory cause some issues on the QA farm or elsewhere.

2) In the past I tried to keep the diff between these similar tools minimal, I'm not sure is there a much in the diff output that could be avoided but that's certainly cosmetic thing so no biggie if left as-is.

Thanks.

natoscott:
Thanks Marko, good point - I'll separate that out & check for any needless differences to other pcp2xxx tools tomorrow.

PR #1881, opened 2024-02-04
Fixed broken pipe issue in pcp-meminfo

orasagar:


src/pcp/meminfo/pcp-meminfo.py
```diff
@@ -222,9 +223,11 @@ def __init__(self):
...
         mngr["sysinfo"] = SYS_METRICS
         mngr.printer = MeminfoReport(opts)
-        sts = mngr.run()
-        sys.exit(sts)
-
+        try:
+            sts = mngr.run()
+            sys.exit(sts)
+        except IOError:
+            signal.signal(signal.SIGPIPE,signal.SIG_DFL)
```
natoscott, line 230: @orasagar hmm, I don't understand what this does.

Do you know where the SIGPIPE is coming from?  (I guess its related to the connection to pmcd?  but libpcp should be ensuring that doesn't propogate up to this level).  Is it reproducible somehow?

Finally, isn't setting the handler to sig_default inside the exception handler is too late?  Or is this a nested failure?

orasagar:
@natoscott when we pipe the output of pcp-meminfo to tail or head it terminates the pipe with given number of outputs .
It is because when the head is done reading data it stops prematurely the pipe, which needs to be handled in python as an IOexpection and terminate the program gracefully.

natoscott:
> @natoscott when we pipe the output of pcp-meminfo to tail or head it terminates the pipe with given number of outputs . It is because when the head is done reading data it stops prematurely the pipe, which needs to be handled in python as an IOexpection and terminate the program gracefully.

Ah, I remember now - it threw me because of the way its nested - we've tackled this in other tools earlier haven't we?  Could we move the exception up to the top level along with the rest, like is done in pcp-buddyinfo.py?

Also, if you have some available time could you audit the other tools below python and check all are OK?  (e.g. I think pcp-numastat may need this too?)

orasagar:
@natoscott Yeah we have tacked in other tools in there run function but here i tried to handle it in it's main .
Sure will audit the other python tools

natoscott:
If I understand the problem correctly, we could be at any point in processing when the tool receives sigpipe, so seems like best bet is to handle the exception at top level (and consistently with other tools as a bonus).

orasagar:
Yes, that's correct but it's best to handle it at the print level because in our Python tools, we generally fill the buffer then with the print function we send out that buffer, where this issue is occurring.

natoscott:
OK - if we're going to change it here to be like that for those reasons, let's do it the same in the other tools too?

orasagar:
Other tools are already addressing it. but If you want the handling to be uniform then will handle it like the way other tools are handling it.

PR #1882, opened 2024-02-06
Rename all /var/run file context entries to /run

zpytela:
With the 1f76e522a ("Rename all /var/run file context entries to /run") selinux-policy commit, all /var/run file context entries moved to /run and the equivalency was inverted. Subsequently, changes in pcp.fc need to be done, too, in a similar manner.

### DIFF
--- a/src/selinux/pcp.fc
+++ b/src/selinux/pcp.fc
@@ -29,4 +29,4 @@
 
 /var/log/pcp(/.*)?		gen_context(system_u:object_r:pcp_log_t,s0)
 
-/var/run/pcp(/.*)?		gen_context(system_u:object_r:pcp_var_run_t,s0)
+/run/pcp(/.*)?		gen_context(system_u:object_r:pcp_var_run_t,s0)


zpytela:
Please note this change is F40+ only and will subsequently apply to RHEL 10, not earlier releases. This change should require >= selinux-policy-40.12-1.fc40.

natoscott:
@zpytela thanks!  If I understand correctly, applying this change will prevent/break new versions of PCP working with all previous versions of selinux prior to 40.12-1 - is that right?  (if so, this is going to be problematic to deliver seamlessly to our users - PCP is often upgraded independently to selinux and we also want new versions of PCP to continue to work on el7, el8, el9).

In the line you've changed, should the pcp_var_run_t not also be updated to pcp_run_t?

In order to allow new PCP versions to continue to function on older versions of selinux, can we use both these lines for a transition period?  (i.e. have both /var/run and /run lines rather than replacing the old with the new)?  Could this be a way to provide a cleaner upgrade path?

zpytela:
> @zpytela thanks! If I understand correctly, applying this change will prevent/break new versions of PCP working with all previous versions of selinux prior to 40.12-1 - is that right? (if so, this is going to be problematic to deliver seamlessly to our users - PCP is often upgraded independently to selinux and we also want new versions of PCP to continue to work on el7, el8, el9).
Correct. With the right Requires: field in specfile, you can ensure updates work smoothly. I don't know how you handle different versions for different target OS, we have different branches. Possibly some `if` statements in pcp SELinux sources, or using both /var/run and /run entries.

> 
> In the line you've changed, should the pcp_var_run_t not also be updated to pcp_run_t?
Not now. In selinux-policy, I plan such changes later, but anyway there needs to be an alias maintained for some time like over 2 releases in Fedora and till end of support in RHEL. If you wish, I can submit it for pcp.

> 
> In order to allow new PCP versions to continue to function on older versions of selinux, can we use both these lines for a transition period? (i.e. have both /var/run and /run lines rather than replacing the old with the new)? Could this be a way to provide a cleaner upgrade path?
A part of the new selinux-policy package is a workaround script which should resolve compatibility problems so I thought this would not be needed, but using both /var/run and /run entries is possible - the other which is overriden by equivalency rule just does not apply.

natoscott:
> > @zpytela thanks! If I understand correctly, applying this change will prevent/break new versions of PCP working with all previous versions of selinux prior to 40.12-1 - is that right? (if so, this is going to be problematic to deliver seamlessly to our users - PCP is often upgraded independently to selinux and we also want new versions of PCP to continue to work on el7, el8, el9).
> 
> Correct. With the right Requires: field in specfile, you can ensure updates work smoothly. I don't know how you handle different versions for different target OS though, we use different branches. Possibly some `if` statements in pcp SELinux sources, or using both /var/run and /run entries together.
> 

This kind of backward-compatibility issue hasn't arisen in the .fc file before, so we're in new territory here.

> > In the line you've changed, should the pcp_var_run_t not also be updated to pcp_run_t?
> 
> Not now. In selinux-policy, I plan such changes later, but anyway there needs to be an alias maintained for some time like over 2 releases in Fedora and till end of support in RHEL. If you wish, I can submit it for pcp.
> 

We may as well switch once rather than twice though?  I see there are a lot of policies using xxx_var_run_t, that PCP  depends on ... hopefully everyone adds aliases otherwise this is going to be a compatibility nightmare.

But anyway, yeah, please do add in a pcp_var_run_t / pcp_run_t alias.

> > In order to allow new PCP versions to continue to function on older versions of selinux, can we use both these lines for a transition period? (i.e. have both /var/run and /run lines rather than replacing the old with the new)? Could this be a way to provide a cleaner upgrade path?
> 
> A part of the new selinux-policy package is a workaround script which should resolve compatibility problems so I thought this would not be needed, but using both /var/run and /run entries is possible - the other which is overriden by equivalency rule just does not apply.

OK, cool - let's go with that for now to limit the fallout.  In several years time we can revisit and drop the original /var compatibility naming.

zpytela:
@natoscott We did not do in selinux-policy mass renaming of _var_run_t to _run_t in the end. Do you want me to do it just for pcp?

Second question: The entries for /var/run are still needed in RHEL 9. Do you share the policy between all releases? In that case there can be duplicate entries.

natoscott:
> @natoscott We did not do in selinux-policy mass renaming of _var_run_t to _run_t in the end. Do you want me to do it just for pcp?

Not if we don't really need it I guess, thanks.

> Second question: The entries for /var/run are still needed in RHEL 9. Do you share the policy between all releases? In that case there can be duplicate entries.

Yep, we share policy between different releases.

PR #1884, opened 2024-02-07
    Add PCP netatop BPF PMDA to extract per process network statistic.

stanfordcox:
Add PCP netatop BPF PMDA to extract per process network statistic.
    
    Add a per process network metric PMDA which exposes per process
    network metrics using bytedance's netatop.bpf.c. Add display of
    metrics to atop.  Add test for network metrics modelled after the
    related bcc pmda test 1286.


stanfordcox:
Creating a new pull request from a clean branch

PR #1885, opened 2024-02-07
Addition of pcp2openmetrics tool 

lmchilton:
Add tool pcp2openmetrics to support conversion of pcp metrics to open metric format. 

Updated relevant QA tests to test basic functionality of tool as well as HTTP transfer function. 

Updated pcp2json QA test 1589 to use a pythonserver instead of nc to test HTTP functionality.

### DIFF
--- a/qa/1131
+++ b/qa/1131
@@ -1,6 +1,6 @@
 #!/bin/sh
 # PCP QA Test No. 1131
-# Exercise pcp2json.
+# Exercise pcp2json and pcp2openmetrics.
 #
 # Copyright (c) 2017 Red Hat.
 #
@@ -16,6 +16,7 @@ $python -c "from collections import OrderedDict" >/dev/null 2>&1
 [ $? -eq 0 ] || _notrun "python collections OrderedDict module not installed"
 
 which pcp2json >/dev/null 2>&1 || _notrun "pcp2json not installed"
+which pcp2openmetrics >/dev/null 2>&1 || _notrun "pcp2openmetrics not installed"
 
 status=1       # failure is the default!
 signal=$PCP_BINADM_DIR/pmsignal
@@ -23,6 +24,9 @@ $sudo rm -rf $tmp $tmp.* $seq.full
 trap "cd $here; rm -rf $tmp.*; exit \$status" 0 1 2 3 15
 
 A="$here/archives/rep"
+hostname=`hostname`
+machineid=`_machine_id`
+domainid=`_domain_name`
 
 _archive_filter()
 {
@@ -34,6 +38,19 @@ _value_filter()
     sed 's/\(\.[0-9]\)[0-9]*/\1/g'
 }
 
+_filter_pcp2openmetrics()
+{
+    tee -a $here/$seq.full \
+    | col -b \
+    | sed \
+    -e '/domainname=/ s/'$domainid'/DOMAINID/' \
+    -e '/machineid=/ s/'$machineid'/MACHINEID/' \
+    -e '/groupid=/ s/'$UID'/GROUPID/' \
+    -e '/userid=/ s/'$UID'/USERID/' \
+    -e '/hostname=/ s/'$hostname'/HOST/' \
+    | LC_COLLATE=POSIX sort
+}
+
 # real QA test starts here
 echo "---"
 pcp2json -a $A -H -I -z "" | _archive_filter
@@ -43,6 +60,11 @@ echo "---"
 pcp2json -a $A -H -I -Z UTC+0 -x "" | _archive_filter
 echo "---"
 pcp2json -a $A -H -I -z -X -b GB -P 2 -F $tmp.outfile ""
+echo "---"
+pcp2openmetrics -s1 -H -z hinv.ncpu | _filter_pcp2openmetrics
+echo "---"
+
+
 cat $tmp.outfile | _archive_filter
 which json_verify > /dev/null 2>&1
 if [ $? -eq 0 ]; then
--- a/qa/1589.out
+++ b/qa/1589.out
@@ -2,11 +2,15 @@ QA output created by 1589
 pcp2json invocation
 pcp2json HTTP POST (sorted):
 
+
+
+
+
 		"value":NCPU
 	      "ncpu": {
 	      }
 	    "@interval": "0",
-	    "@timestamp":SECS,
+	    "@timestamp":DATE TIME
 	    "hinv": {
 	    }
 	  {
@@ -23,11 +27,16 @@ pcp2json HTTP POST (sorted):
   "@pcp": {
   }
 Accept: */*
+Body:
 Content-Length: SIZE
 Content-Type: application/json
+Headers:
 Host: localhost:PORT
-POST /receive HTTP/1.1
+INFO:root:POST request,
+INFO:root:Starting httpd...
+Path: /
 User-Agent: python-requests VERSION
+[IP ADDRESS][DATE] "POST / HTTP/1.1" 200 -
 {
 }
 
--- a/qa/1827
+++ b/qa/1827
@@ -1,8 +1,8 @@
 #!/bin/sh
 # PCP QA Test No. 1589
-# Exercise pcp2json HTTP POST functionality.
+# Exercise pcp2openmetrics HTTP POST functionality.
 #
-# Copyright (c) 2023 Red Hat.  All Rights Reserved.
+# Copyright (c) 2024 Red Hat.  All Rights Reserved.
 # 
 
 #set -x
@@ -12,13 +12,7 @@ echo "QA output created by $seq"
 
 . ./common.python
 
-$python -c "from pcp import pmapi" >/dev/null 2>&1
-[ $? -eq 0 ] || _notrun "python pcp pmapi module not installed"
-$python -c "from collections import OrderedDict" >/dev/null 2>&1
-[ $? -eq 0 ] || _notrun "python collections OrderedDict module not installed"
-
-which pcp2json >/dev/null 2>&1 || _notrun "pcp2json not installed"
-which nc >/dev/null 2>&1 || _notrun "nc binary not installed"
+which pcp2openmetrics >/dev/null 2>&1 || _notrun "pcp2openmetrics not installed"
 
 _cleanup()
 {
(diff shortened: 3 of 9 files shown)


src/pcp2openmetrics/pcp2openmetrics.py
```diff
@@ -0,0 +1,569 @@
...
+import os
+import cpmapi
+
+# PCP Python PMAPI
+from pcp import pmapi, pmconfig
+from cpmapi import PM_CONTEXT_ARCHIVE, PM_INDOM_NULL, PM_IN_NULL, PM_DEBUG_APPL1, PM_TIME_SEC
+
+if sys.version_info[0] >= 3:
+    long = int # pylint: disable=redefined-builtin
+
```
natoscott, line 35: You can remove the above two lines, as long() is not used in this script.

src/pcp2openmetrics/pcp2openmetrics.py
```diff
@@ -0,0 +1,569 @@
...
+        self.version = CONFVER
+        self.source = "local:"
+        self.output = None # For pmrep conf file compat only
+        self.speclocal = None
+        self.derived = None
+        self.header = 1
+        self.globals = 1
+        self.samples = None # forever
+        self.interval = pmapi.timeval(10)      # 60 sec
+        self.opts.pmSetOptionInterval(str(10)) # 60 sec
```
natoscott, line 88: I think the comments are wrong on the previous two lines?

src/pcp2openmetrics/pcp2openmetrics.py
```diff
@@ -0,0 +1,569 @@
...
+            """ pcp.io metric name to openmetrics.io name conventions """
+            return metric.replace('.','_')
+        
+        def openmetrics_help(context, pmid):
+            """ convert pcp.io metric helptext to openmetrics.io HELP """
+            return context.pmLookupText(pmid[0]).decode()
+        
+        def openmetrics_type(context, pmid):
+            """ convert pcp.io metric metadata to openmetrics.io TYPE """
+            desc = context.pmLookupDesc(pmid[0])
```
natoscott, line 462: The caller has the desc already, we don't need this call.  pmLookupDesc involves another round trip to pmcd for every metric, so we don't want to do that on every fetch.

src/pcp2openmetrics/pcp2openmetrics.py
```diff
@@ -0,0 +1,569 @@
...
+                mtype = "unknown"
+            return mtype
+
+        def openmetrics_name(metric):
+            """ pcp.io metric name to openmetrics.io name conventions """
+            return metric.replace('.','_')
+        
+        def openmetrics_help(context, pmid):
+            """ convert pcp.io metric helptext to openmetrics.io HELP """
+            return context.pmLookupText(pmid[0]).decode()
```
natoscott, line 458: Hmm, pmLookupText involves a roundtrip to pmcd on every call (i.e. for every metric here).  We should move this out of the fetch loop - create a dictionary at the start which maps each metric to its help text, and do a lookup here instead of communicating with pmcd (we'll get the same answer every time for help text).

src/pcp2openmetrics/pcp2openmetrics.py
```diff
@@ -0,0 +1,569 @@
...
+            desc = context.pmLookupDesc(pmid[0])
+            if desc.sem == cpmapi.PM_SEM_COUNTER:
+                return 'counter'
+            return 'gauge'
+        
+        def openmetrics_labels(metric, context, inst, name, desc):
+            """ filter pcp.io labels here; pick out the labels needed """
+            result = ''
+            pmid = context.pmLookupName(metric)
+            labels = context.pmLookupLabels(pmid[0])
```
natoscott, line 471: Both of these calls involve round trips communication to pmcd - we already have the pmid in the desc fortunately.  The labels are like the help text - we should do a single lookup at the start of the program for all metrics, store it in a dictionary, and do a quick lookup here.

src/pcp2openmetrics/pcp2openmetrics.py
```diff
@@ -0,0 +1,569 @@
...
+            labels[1].update(instname=name, instid=inst)
+            new_dict = {} 
+            new_dict['semantics'] = self.context.pmSemStr(desc.contents.sem)
+            new_dict['type'] = get_type_string(desc)
+            for key in labels: 
+                new_dict.update(labels[key])
+            if self.everything: 
+                subset = ['domainname', 'test', 'groupid', 'hostname', 'machineid', 'userid', 'instname', 'instid', 'agent', 'device_type', 'indom_name', 'semantics', 'type']
+            else:
```
natoscott, line 480: Shouldn't everything really be everything - all labels - not a subset?  Labels can be arbitrary and user-supplied so we probably don't want to be restricting them here at all.

src/pcp2openmetrics/pcp2openmetrics.py
```diff
@@ -0,0 +1,569 @@
...
+            if desc.sem == cpmapi.PM_SEM_COUNTER:
+                return 'counter'
+            return 'gauge'
+        
+        def openmetrics_labels(metric, context, inst, name, desc):
+            """ filter pcp.io labels here; pick out the labels needed """
+            result = ''
+            pmid = context.pmLookupName(metric)
+            labels = context.pmLookupLabels(pmid[0])
+            labels[1].update(instname=name, instid=inst)
```
natoscott, line 472: If this metric is single-valued (i.e. desc.indom == PM_INDOM_NULL) we should not add instname and instid labels.

src/pcp2openmetrics/pcp2openmetrics.py
```diff
@@ -0,0 +1,569 @@
...
+            try:
+                self.writer.write("\n")
+                self.writer.flush()
+            except IOError as write_error:
+                if write_error.errno != errno.EPIPE:
+                    raise
+            try:
+                self.writer.close()
+            except Exception:
+                pass
```
natoscott, line 544: Hmm, that is a pretty broad exception handler - what is that ignoring in particular?

natoscott (commented):
@lmchilton nice work Lauren, comments inline.

qa/1589
```diff
@@ -51,21 +53,27 @@ _filter_pcp2json_http()
...
+pid=$!
+sleep 2 # let server start up
 
 echo "pcp2json invocation" | tee -a $here/$seq.full
-pcp2json -s1 -f%s -ZUTC --url http://localhost:$port/receive hinv.ncpu > $tmp.json.out 2> $tmp.json.err
+pcp2json -s1 -u http://localhost:$port hinv.ncpu >$tmp.json.out 2>$tmp.json.err
+pid32=$!
+sleep 2
+
+kill $pid2
```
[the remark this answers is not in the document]
lmchilton: @natoscott  I wanted to capture the pcp2json pid so I could use kill to act as the "ctrl C" command that pcp2json requires for metrics to output. Is there a different approach that you would take to accomplish that?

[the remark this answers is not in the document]
natoscott: The -s1 option means "take one sample only" (and then exit) so there should be no need to kill it.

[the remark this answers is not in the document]
lmchilton: @natoscott very odd, running with -s2 in the QA test requires ctrl C but -s1 does not. I guess I will roll with -s1.

qa/1827.out
```diff
@@ -0,0 +1,23 @@
...
+Content-Length: SIZE
+Content-Type: application/openmetrics
+Headers:
+Host: localhost:PORT
+INFO:root:POST request,
+INFO:root:Starting httpd...
+Metric hinv.ncpu details (last fetch: 1707343573):# HELP hinv_ncpu number of CPUs in the system
+Path: /
+User-Agent: python-requests VERSION
+hinv_ncpu{domainname="localdomain",groupid="4209557",hostname="lchilton-thinkpadp1gen4i.remote.csb",machineid="56fdc9f7f62c4799ba51476639d78805",userid="4209557",instname="None",instid="4294967295",agent="linux"} 16
```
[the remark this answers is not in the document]
lmchilton: @natoscott I filtered out all of the labels and just have metric name and metric value, but is it better to keep the labels and then just replace hostname with "HOSTNAME" etc with the rest of them?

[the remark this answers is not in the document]
natoscott: Yep - have a look at similar tests for an example, this label filtering pattern is common.

qa/1589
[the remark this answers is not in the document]
natoscott: It will depend on the sample interval I think (-t option) ... -s2 means "take a sample, sleep for sample interval, take a sample, exit" whereas -s1 means "take a sample, exit".

natoscott (changes requested):
@lmchilton there's a few QA issues lurking still:

- the output in test 1131.out is not filtered (most labels)m
- 1827.out contains unfiltered domainname label value,
- not sure what "+Metric hinv.ncpu details (last fetch: 1707935533):" is?  but it looks at least unfiltered, possibly just a debug diagnostic?
- maybe update the header comment in qa/1131 also to say it tests both pcp2json and pcp2openmetrics now

pcp2openmetrics.1 has some incorrect dates in the copyright notice as well.
Otherwise looking good I think.

qa/1131
```diff
@@ -9,20 +9,25 @@ seq=`basename $0`
 echo "QA output created by $seq"
 
 . ./common.python
+. ./common.check
```
natoscott, line 12: common.python already brings in common.check so we don't need this new line AFAICT.

qa/1589.out
```diff
@@ -2,11 +2,15 @@ QA output created by 1589
...
+
+
+
+
 		"value":NCPU
 	      "ncpu": {
 	      }
 	    "@interval": "0",
-	    "@timestamp":SECS,
+	    "@timestamp": "2024-02-14 18:31:30",
```
natoscott, line 13: Something strange happened here and we no longer seem to be filtering this timestamp field?

qa/1589.out
```diff
@@ -22,11 +26,16 @@ pcp2json HTTP POST (sorted):
     ]
   "@pcp": {
   }
+127.0.0.1 - - [14/Feb/2024 13:31:30] "POST / HTTP/1.1" 200 -
```
natoscott, line 29: This date and timestamp will need to be filtered, and possibly that IP address.

qa/1827
```diff
@@ -0,0 +1,70 @@
...
+# Copyright (c) 2024 Red Hat.  All Rights Reserved.
+# 
+
+#set -x
+
+seq=`basename $0`
+echo "QA output created by $seq"
+
+. ./common.python
+. ./common.check
```
natoscott, line 14: Can drop common.check here as noted above.

qa/1827
```diff
@@ -0,0 +1,70 @@
...
+pid=$!
+sleep 2 # let server start up
+
+echo "pcp2openmetrics invocation" | tee -a $here/$seq.full
+pcp2openmetrics -s1 -u http://localhost:$port hinv.ncpu >$tmp.openmetrics.out 2>$tmp.openmetrics.err
+
+echo "pcp2openmetrics HTTP POST (sorted):"
+_filter_pcp2openmetrics_http <$tmp.python.out
+
+($signal $pid ) >>$seq.full 2>&1 &
```
natoscott, line 67: Consider moving this line (signalling the server) to the _cleanup routine, because it will be called even when someone presses ^C while the test is running.

natoscott (approved):
Just a few small QA things left, then this is good to go.

PR #1888, opened 2024-02-08
Add PCP netatop BPF PMDA to extract per process network statistic.

stanfordcox:
Add a bpf per process network metric PMDA which exposes per process network metrics using netatop-bpf.c. Add display of metrics to atop. Add test for network metrics.

### DIFF
--- a/qa/1264
+++ b/qa/1264
@@ -49,12 +49,10 @@ _stop_auto_restart pmcd
 
 # real QA test starts here
 
-_pmdabpf_install $tmp.conf
-sudo ln -s  /usr/local/bpf.log /usr/local//var/log/pcp//pmcd/
+_pmdabpf_install $tmp.conf | _filter_pmda_install
 _pmdabpf_wait_for_metric
 
 # Generate system activity for the BCC netproc module
-pwd
 $PCP_PYTHON_PROG $here/src/bcc_netproc.py server >> "$seq.full" &
 server_pid=$!
 
@@ -73,13 +71,13 @@ wait ${server_pid}
 echo "server PID: ${server_pid}" >> "$seq.full"
 echo "client PID: ${client_pid}" >> "$seq.full"
 
-for metric in bpf.proc.net.tcp.send.packs \
+for metric in bpf.proc.net.tcp.send.packets \
               bpf.proc.net.tcp.send.bytes \
-              bpf.proc.net.tcp.recv.packs \
+              bpf.proc.net.tcp.recv.packets \
               bpf.proc.net.tcp.recv.bytes \
-              bpf.proc.net.udp.send.packs \
+              bpf.proc.net.udp.send.packets \
               bpf.proc.net.udp.send.bytes \
-              bpf.proc.net.udp.recv.packs \
+              bpf.proc.net.udp.recv.packets \
               bpf.proc.net.udp.recv.bytes
 do
     echo && echo && echo "=== report metric values for $metric ==="
--- a/qa/1264
+++ b/qa/1264
@@ -1,0 +1,96 @@
+#!/bin/sh
+# PCP QA Test No. 1286
+# Exercise the BPF PMDA netproc hits module - install, remove and values.
+#
+# Copyright (c) 2024 Red Hat.
+#
+
+# 1286 is the bcc variant of this
+seq=`basename $0`
+echo "QA output created by $seq"
+
+. ./common.bpf
+
+_pmdabpf_check
+_pmdabpf_require_kernel_version 5 0
+_pmdabpf_require_libbpf_version 0 7
+which curl >/dev/null 2>&1 || _notrun "No curl binary installed"
+
+# target_ip='http://localhost:44322/pmapi/fetch?names=hinv.ncpu,disk.dev.read'
+status=1       # failure is the default!
+signal=$PCP_BINADM_DIR/pmsignal
+$sudo rm -rf $tmp.* $seq.full
+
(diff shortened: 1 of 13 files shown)


qa/1264
```diff
@@ -0,0 +1,96 @@
...
+}
+
+_prepare_pmda bpf
+trap "_cleanup; exit \$status" 0 1 2 3 15
+_stop_auto_restart pmcd
+
+# real QA test starts here
+
+_pmdabpf_install $tmp.conf
+sudo ln -s  /usr/local/bpf.log /usr/local//var/log/pcp//pmcd/
```
natoscott, line 53: I think this may be a leftover from local work / debugging?

src/pmdas/bpf/modules/netatop.c
```diff
@@ -0,0 +1,473 @@
...
+}
+
+static unsigned int netatop_indom_count(void)
+{
+    return INDOM_COUNT;
+}
+
+static void netatop_set_indom_serial(unsigned int local_indom_id, unsigned int global_id)
+{
+    pmNotifyErr(LOG_INFO, "%s", __FUNCTION__);
```
natoscott, line 121: Here and in several other places we have unguarded diagnostics that may be filling the PMDA bpf.log file during operation.  Could we add guards for all/most of these such as:

```
if (pmDebugOptions.appl0 || pmDebugOptions.desperate)
    pmNotifyErr(LOG_INFO, "%s", __FUNCTION__);
```

The full list of debug flags is available running: 'pmdbg -l'.

src/pmdas/bpf/modules/netatop.c
```diff
@@ -0,0 +1,473 @@
...
+    }
+
+    return 0;
+}
+
+
+static void netatop_register(unsigned int cluster_id, pmdaMetric *metrics, pmdaIndom *indoms)
+{
+    pmNotifyErr(LOG_INFO, "%s Adding netatop metrics", __FUNCTION__);
+    /* bpf.netatop.tcpsndpacks */
```
natoscott, line 140: The metric name in the comment (here and below) doesn't match the namespace used - I guess this changed over time?

src/pmdas/bpf/modules/netatop.c
```diff
@@ -0,0 +1,473 @@
...
+enum metric_name {
+  TCPSNDPACKS,
+  TCPSNDBYTES,
+  TCPRCVPACKS,
+  TCPRCVBYTES,
+  UDPSNDPACKS,
+  UDPSNDBYTES,
+  UDPRCVPACKS,
+  UDPRCVBYTES };
+enum metric_indom { NETATOP_INDOM };
```
natoscott, line 58: (consistency) unusual coding style for this enum metric_name.

src/pmdas/bpf/modules/netatop.c
```diff
@@ -0,0 +1,473 @@
...
+static struct env {
+    int process_count;
+} env = {
+    .process_count = 20,
+};
+
+int tgid_map_fd;
+int nr_cpus;
+
+static pmdaInstid *netatop_instances = NULL;
```
natoscott, line 40: No need to assign to NULL here, its mandatory for the compiler to do this for globals.

src/pmdas/bpf/modules/netatop.c
```diff
@@ -0,0 +1,473 @@
...
+#include <signal.h>
+#include <string.h>
+#include <errno.h>
+#include <dirent.h>
+#include <sys/resource.h>
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include <pcp/pmda.h>
+#include <sys/queue.h>
+#include <pcp/pmwebapi.h>
```
natoscott, line 20: should this be <pcp/pmapi.h>?  (not sure - do we use something from webapi here?)
Either way, pmapi.h is included via pmwebapi.h and it includes the headers from line 9 thru 14 I think, so those could be dropped.

src/pmdas/bpf/modules/netatop.c
```diff
@@ -0,0 +1,473 @@
...
+#include <string.h>
```
natoscott, line 21: string.h is included twice too.

src/pmdas/bpf/modules/netatop.h
```diff
@@ -0,0 +1,50 @@
...
+};
+
+extern struct netatop_bpf *skel;
+extern int semid;
+extern int tgid_map_fd;
+extern int tid_map_fd;
+extern int nr_cpus;
+
+#define	NUMCLIENTS	(semctl(semid, 1, GETVAL, 0))
+#define SEMAKEY         1541962
```
natoscott, line 45: Are NUMCLIENTS and SEMAKEY used?  I may be missing it, but if not let's drop these two lines.

natoscott (commented):
@stanfordcox looking great Stan!  Some small improvement suggestions here, but it looks pretty close to done now.

src/pmdas/bpf/modules/netatop.c
stanfordcox: yep, most of the bpf/modules routines include pmwebapi.h for the pmIniFileLookup dcl so I added it back.  Without it rawhide does not build
netatop.c: In function ‘netatop_init’:
netatop.c:271:16: error: implicit declaration of function ‘pmIniFileLookup’ [-Wimplicit-function-declaration]
  271 |     if ((val = pmIniFileLookup(cfg, module_name, "process_count")))

PR #1891, opened 2024-02-13
build: handle missing test file in test-report.py read_testlog()

natoscott:
The 'Parse test results' phase of CI is sometimes failing with
  File "/home/runner/work/pcp/pcp/build/ci/test-report.py", line 83, in read_testlog
    with open(os.path.join(qa_dir, test_no), encoding="utf-8") as f:
FileNotFoundError: [Errno 2] No such file or directory: './qa/4334'

I suspect this maybe a variant on the case handled at the end of read_testlog, where we matched something in a failed tests output. Handle this new scenario too and continue bravely on...

### DIFF
--- a/build/ci/test-report.py
+++ b/build/ci/test-report.py
@@ -80,13 +80,20 @@ def read_testlog(
                 test.path = testartifacts_dir
                 test.start, test.stop = test_durations.get(test_no, (0, 0))
                 test.description = ""
-                with open(os.path.join(qa_dir, test_no), encoding="utf-8") as f:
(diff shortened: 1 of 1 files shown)


andreasgerstmayr (approved):
lgtm :+1: 

To fix the root problem, maybe the qa runner could be changed to emit json, to avoid this regex matching altogether?

natoscott:
| To fix the root problem, maybe the qa runner could be changed to emit json, to avoid this regex matching altogether?

Good idea.

PR #1893, opened 2024-02-16
Add submodule for blazesym

stanfordcox:
Add submodule for vendor/github.com/iovisor/bcc/libbpf-tools/blazesym to avoid issues with: git submodule update 

Add submodule for _vendor/github.com/iovisor/bcc/libbpf-tools/blazesym_ to avoid issues with: _git submodule update --init --recursive._
 Wondering about blazesym though and if there is some cleanup needed here. It is a rust module that only seems to be referenced by _vendor/github.com/iovisor/bcc/libbpf-tools/opensnoop.c._. The patch also removes
_vendor/github.com/iovisor/bcc/libbpf-tools/{blazesym,bpftool)_ as **git submodule update --init --recursive** complains otherwise. bpftool is actually populated in _vendor/github.com/libbpf/bpftool_

### DIFF
--- a/vendor/github.com/iovisor/bcc/libbpf-tools/.gitmodules
+++ b/vendor/github.com/iovisor/bcc/libbpf-tools/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "blazesym"]
+	path = blazesym
+	url = https://github.com/libbpf/blazesym.git


natoscott:
Seems OK to me as a way to fix the problem - @andreasgerstmayr any thoughts?

andreasgerstmayr (commented):
lgtm

PR #1929, opened 2024-03-25
pmcheck: add service and metric checks for the redis PMDA

natoscott:
@kmcdonell here's my first attempt at a pmcheck script, for Redis metrics.

Couple of areas of uncertainty.  I wasn't sure about the protocol for _chk_svc 'state' argument - that one seems to call 'exit' very aggressively (always), which is a problem here because I have other environmental aspects to check too.  I also noticed the use of a local variable called 'exit' in the shell lib code, figured this will end in tears so renamed it to __exit for clarity.  Wondering if all variables in this common file should be updated to use this convention instead of the mixed bag we have currently?

Anyway, please have a look - this seems to do the right thing now for me locally.

### DIFF
--- a/qa/1491.out
+++ b/qa/1491.out
@@ -3,6 +3,7 @@ basic stuff ...
 -- no args --
 pmcd            OK
 pmda-overhead   OK
+pmda-redis      could be activated
 pmda-sample     OK
 pmie            OK
 pmlogger        OK
@@ -11,6 +12,7 @@ zeroconf        OK
 -- -l --
 pmcd           
 pmda-overhead  
+pmda-redis     
 pmda-sample    
 pmie           
 pmlogger       
@@ -19,6 +21,7 @@ zeroconf
 -- -lv --
 pmcd            Performance Metrics Collection Daemon - local source of performance data
 pmda-overhead   overhead PMDA - ...
+pmda-redis      Redis PMDA - metrics from redis-server(1)
 pmda-sample     sample PMDA
 pmie            Inference Engine - rule-based monitoring
 pmlogger        Archive logger - record performance data for subsequent replay
@@ -27,6 +30,7 @@ zeroconf        Zeroconf Package ...
 -- -s --
 pmcd            OK
 pmda-overhead   OK
+pmda-redis      could be activated
 pmda-sample     OK
 pmie            OK
 pmlogger        OK
--- a/qa/913
+++ b/qa/913
@@ -2,7 +2,7 @@
 # PCP QA Test No. 913
 # Exercise the Redis PMDA and metrics.
 #
-# Copyright (c) 2017 Red Hat.
+# Copyright (c) 2017,2024 Red Hat.
 #
 
 seq=`basename $0`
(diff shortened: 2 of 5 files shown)


natoscott:
@kenj oh, another thought - not sure I like the no.xxx.pmcheck approach.  What is the advantage of that over not installing xxx.pmcheck at all in those situations?  The flip side is we have two scripts to maintain for every conditional PMDA - wouldn't it be better to manage this solely in the makefile where the not-built logic already exists?

src/pmcheck/checkproc.sh
kmcdonell, line 1: exit is correct at line 212.  the code above has set $status prior to this and the trap exits with $status ... $rc has not been set in this code block for the "state" action.

And I'm not sure about the renaming of $exit to $__exit ... exit (like name, action, domain, ...) is declared local to _ctl_pmda() so they don't pollute the shell variable namespace of the "plugin" scripts that source checkproc.sh

src/pmdas/redis/no.redis.pmcheck
kmcdonell, line 1: This looks OK to me.

kmcdonell (changes requested):
only the "exit" change needs undoing
I'll add more feedback to your questions below

kmcdonell:
Re. variable naming in checkproc.sh ... there are a number of constraints (and I may not have been entirely consistent), but this is what I was trying to do:
1. controlled pollution of the calling shell's namespace, e.g. from _do_args() we need to set the variables initialized in the block aflag= ... verbose=
2. use local declarations to hide shell variables inside the procedure they are used
3. there are some ugly __use_name ones that are in the one-trip block near the end of checkproc.sh that I wanted to avoid re-evaluating every time one of the _ctl_foo() functions was called

Now in _do_args() I can see $__args should be local $args (this code predates rule 2 above).
And in _ctl_svc the locals __state and __runlevel don't need the leading __.

If I fix $__args, $__state and $__runlevel I think things would be more consistent.  Thoughts?

src/pmcheck/checkproc.sh
natoscott: @kmcdonell mornin'!

The issue re line 212 was it was causing the top level script to quit early on success, when I still had more things I wanted to check beyond service state.

Regarding the __exit thing, my worry there is someday someone will be happily hacking on this code and add a call to 'exit 99' or whatever, and get a surprise :). Looking again, we don't seem to use that variable except in verbose mode?  I recommend we just pick a different name there.

src/pmdas/redis/no.redis.pmcheck
natoscott: I don't like it :) - I copied this from overhead but I don't understand the point of doing this.  If the PMDA doesn't exist (wasn't built, so assume not installable) - seems to me we should not add anything (extra code to maintain) for it?

natoscott:
> Re. variable naming in checkproc.sh ... there are a number of constraints (and I may not have been entirely consistent), but this is what I was trying to do:
> 

Yep, understood.  it was really just the 'exit' part of the naming that raised a red flag as my eye looked over it this time.

> If I fix $__args, $__state and $__runlevel I think things would be more consistent. Thoughts?

Works for me.

PR #1931, opened 2024-03-27
pmproxy: disable Redis protocol proxying by default

natoscott:
If a redis-server has been locked down in terms of connections, we want to prevent pmproxy from being allowed to send arbitrary RESP commands to it.

This protocol proxying doesn't affect PCP functionality at all, its more of a developer/sysadmin convenience when Redis used in cluster mode (relatively uncommon compared to localhost mode).


natoscott:
Merged.

PR #1935, opened 2024-04-02
pmcheck: postgresql service detection and agent recommendation

natoscott:
This commit adds an interface allowing any python PMDA script to be called by its corresponding pmcheck plugin to probe and see if it could be enabled.


natoscott:
Merged.

PR #1948, opened 2024-04-09
build: revert the deb dh_strip override, debugging symbols needed

natoscott:


natoscott:
This causes an unexplained failure in dbgsym generation & installation, so not proceeding.

PR #1956, opened 2024-04-15
libpcp_web: ensure openmetrics scrape sees all labels

natoscott:
Marko noticed that cluster labels were not being requested in the agent labels callback - missed calls in libpcp_web.


myllynen:
This fixes the issue I was seeing, thanks a lot!

PR #1961, opened 2024-04-18
Fixes issues RHEL-32434 and RHEL-32407

lmchilton:
Fixes issues RHEL-32434 and RHEL-32407. Also pcp2openmetrics- addition of timestamps in archive mode.


src/pcp2openmetrics/pcp2openmetrics.py
```diff
@@ -483,14 +486,24 @@ def openmetrics_labels(inst, name, desc, labels):
...
+            semantics = self.context.pmSemStr(desc.contents.sem)
+            units = desc.contents.units
+            pmIDStr = context.pmIDStr(pmid[0])
+            pmIndomStr = context.pmInDomStr(desc)
+            
             help_dict = {}
             help_dict[metric] = context.pmLookupText(pmid[0])
 
+            if self.context.type == PM_CONTEXT_ARCHIVE:
+                body += "\nMetric %s details (last fetch: %d)\n:" % (metric, ts)
```
natoscott, line 504: @lmchilton this bit doesn't look right...?

natoscott (commented):
I think the timestamp part is missing - is that part work-in-progress?

src/pcp2openmetrics/pcp2openmetrics.py
lmchilton: @natoscott that is the part that adds the timestamps in archive mode. What part does not look right?  The body string format I took from pcp2template, if it should be in a different format?

natoscott: (discussed elsewhere, but yep, different format needed)

PR #1967, opened 2024-04-23
tmpfiles.d: extend the tmpfiles.d directory set to logdirs

natoscott:
Resolves Fedora BZ #2275408


martinpitt:
Thanks @natoscott !

PR #1975, opened 2024-05-03
Add an amdgpu pmda

fberat:
This pmda retrieves data using the libdrm and libdrm-amdgpu libraries. It only retrieves general information, no process specific data.

Data retrieved includes memory usage, memory speed, GPU speed, temperature, etc ...

Old Radeon (Pre GCN 1.1) aren't supported.


src/pmdas/amdgpu/GNUmakefile
```diff
@@ -0,0 +1,51 @@
+#
+# Copyright (c) 2014-2015,2020 Red Hat.
```
fberat, line 2: I forgot to update the dates.

src/pmdas/amdgpu/help
```diff
@@ -0,0 +1,66 @@
+#
+# Copyright (c) 2014,2019,2021 Red Hat.
+# Copyright (c) 2000-2004 Silicon Graphics, Inc.  All Rights Reserved.
```
fberat, line 3: Copyrights need to be revised.

man/man1/pmdaamdgpu.1
```diff
@@ -0,0 +1,156 @@
...
+.ft CR
+.nf
+.in +0.5i
+# cd $PCP_PMDAS_DIR/amdgpu
+# ./Remove
+.in
+.fi
+.ft 1
+.PP
+If you want to establish access to the names, help text and values for the amdgpu
```
natoscott, line 89: amdgpu -> AMD GPU

src/pmdas/amdgpu/GNUmakefile
```diff
@@ -0,0 +1,51 @@
...
+
+IAM	= amdgpu
+DOMAIN	= AMDGPU
+
+CMDTARGET = pmdaamd$(EXECSUFFIX)
+LIBTARGET = pmda_amd.$(DSOSUFFIX)
+CFILES	= localdrm.c amdgpu.c
+HFILES	= localdrm.h
+DFILES	= README
+LLDLIBS	= $(PCP_PMDALIB) $(LIB_FOR_DLOPEN) -ldrm -ldrm_amdgpu
```
natoscott, line 26: Depending on how the configure.ac machinery ends up (discussed elsewhere), the libraries listed here will likely end up accessed as makefile macros.

src/pmdas/amdgpu/GNUmakefile
```diff
@@ -0,0 +1,51 @@
...
+LCFLAGS += -DDSOSUFFIX=\"$(DSOSUFFIX)\" -I/usr/include/drm
```
natoscott, line 27: Likewise any CFLAGS additions.

src/pmdas/amdgpu/Install
```diff
@@ -0,0 +1,26 @@
...
+# under the terms of the GNU General Public License as published by the
+# Free Software Foundation; either version 2 of the License, or (at your
+# option) any later version.
+# 
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+# for more details.
+#
+# Install the trivial PMDA and/or PMNS
```
natoscott, line 15: trivial -> "AMD GPU"

src/pmdas/amdgpu/Remove
```diff
@@ -0,0 +1,38 @@
...
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+# for more details.
+# 
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+#
+# Remove the amd PMDA
```
natoscott, line 19: amd -> "AMD GPU"

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,547 @@
...
+  AMDGPU_GPU_LOAD,
+  AMDGPU_GPU_AVG_PWR,
+
+  AMDGPU_METRIC_COUNT
+};
+
+/* Table of metrics exported by this PMDA */
+static pmdaMetric metrictab[] = {
+    {NULL,
+     {PMDA_PMID(0, AMDGPU_NUMCARDS), PM_TYPE_U32, PM_INDOM_NULL, PM_SEM_INSTANT,
```
natoscott, line 54: PM_SEM_DISCRETE maybe, or are these cards hot-pluggable?

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,547 @@
...
+      PMDA_PMUNITS(1, 0, 0, PM_SPACE_BYTE, 0, 0)}},
+    {NULL,
+     {PMDA_PMID(0, AMDGPU_MEMTOTAL), PM_TYPE_U64, GCARD_INDOM, PM_SEM_DISCRETE,
+      PMDA_PMUNITS(1, 0, 0, PM_SPACE_BYTE, 0, 0)}},
+    {NULL,
+     {PMDA_PMID(0, AMDGPU_MEMFREE), PM_TYPE_U64, GCARD_INDOM, PM_SEM_INSTANT,
+      PMDA_PMUNITS(1, 0, 0, PM_SPACE_BYTE, 0, 0)}},
+    {NULL,
+     {PMDA_PMID(0, AMDGPU_SAMPLES), PM_TYPE_U64, GCARD_INDOM, PM_SEM_COUNTER,
+      PMDA_PMUNITS(0, 0, 0, 0, 0, 0)}},
```
natoscott, line 70: Units for a basic counter are usually like:

PMDA_PMUNITS(0,0,1,0,0,PM_COUNT_ONE)

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,547 @@
...
+    {NULL,
+     {PMDA_PMID(0, AMDGPU_MEMUSED_ACCUM), PM_TYPE_U64, GCARD_INDOM, PM_SEM_COUNTER,
+      PMDA_PMUNITS(1, 0, 0, PM_SPACE_BYTE, 0, 0) } },
+    {NULL,
+     {PMDA_PMID(0, AMDGPU_GPU_CLOCK), PM_TYPE_U32, GCARD_INDOM, PM_SEM_COUNTER,
+      PMDA_PMUNITS(1, 0, 0, 0, 0, 0) } },
```
natoscott, line 76: Same comment as above (and this repeats several more times below - this applies to all cases)

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,547 @@
...
+    {NULL,
+     {PMDA_PMID(0, AMDGPU_TEMPERATURE), PM_TYPE_U32, GCARD_INDOM, PM_SEM_COUNTER,
+      PMDA_PMUNITS(1, 0, 0, 0, 0, 0) } },
+    {NULL,
+     {PMDA_PMID(0, AMDGPU_GPU_LOAD), PM_TYPE_U32, GCARD_INDOM, PM_SEM_COUNTER,
+      PMDA_PMUNITS(1, 0, 0, 0, 0, 0) } },
+    {NULL,
+     {PMDA_PMID(0, AMDGPU_GPU_AVG_PWR), PM_TYPE_U32, GCARD_INDOM, PM_SEM_COUNTER,
+      PMDA_PMUNITS(1, 0, 0, 0, 0, 0) } },
+};
```
natoscott, line 95: Are these last three metrics really counters?  Seems. a little odd - I'd expect them to be PM_SEM_INSTANT.

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,547 @@
...
+static struct timeval interval;
+
+static int setup_gcard_indom(void) {
+  unsigned int device_count = 0;
+  pmdaIndom *idp = &indomtab[GCARD_INDOM];
+  char gpuname[32], *name;
+  int i, sts;
+
+  /* Initialize instance domain and instances. */
+  if ((sts = localDRMDeviceGetDevices(&pcp_amdgpuinfo.devs,
```
natoscott, line 136: s/local//g - I think this is a hangover from the nvidia agent right?

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,547 @@
...
+        free(pcp_amdgpuinfo.indom->it_set[i].i_name);
+      free(pcp_amdgpuinfo.devs);
+      free(pcp_amdgpuinfo.indom->it_set);
+      free(pcp_amdgpuinfo.info);
+      return -ENOMEM;
+    }
+    pcp_amdgpuinfo.indom->it_set[i].i_name = name;
+
+    /* Get static values */
+    if (localDRMDeviceGetGPUInfo(dev, &pcp_amdgpuinfo.info[i].gpu_info)) {
```
natoscott, line 182: s/local//g -> (throughout, as mentioned above)

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,547 @@
...
+    }
+    if (localDRMDeviceGetTemperature(dev, &info->temperature)) {
+      info->failed[AMDGPU_TEMPERATURE] = 1;
+    }
+    if (localDRMDeviceGetGPULoad(dev, &info->load)) {
+      info->failed[AMDGPU_GPU_LOAD] = 1;
+    }
+    if (localDRMDeviceGetGPUAveragePower(dev, &info->avg_pwr)) {
+      info->failed[AMDGPU_GPU_AVG_PWR] = 1;
+    }
```
natoscott, line 229: The code pattern used here equates to "for every fetch, fetch all the metrics available".  Sometimes, its not ideal (e.g. when sometimes only one or two metrics are fetched, but maybe some library calls are relatively expensive) - if this is not the desired behaviour (i.e. if you want to use the minimum number of library API calls) you can use the fetch/refresh pattern from src/pmdas/linux/pmda.c with a need_refresh array and carefully setup pmID clusters.
But, up to you - it may not be an issue worth spending time on here.

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,547 @@
...
+      /* The GPU max speed is the max_memory_clk.
+       * The GDDRx memory max speed is max_engine_clk
+       * In MHz
+       */
+      atom->ul = pcp_amdgpuinfo.info[inst].gpu_info.max_engine_clk;
+      break;
+    case AMDGPU_TEMPERATURE:
+      if (pcp_amdgpuinfo.info[inst].failed[AMDGPU_TEMPERATURE])
+        return PM_ERR_VALUE;
+      /* In millidegrees Celsius */
```
natoscott, line 339: One possibility here is to add a label to this metric describing the temperature units (since temperature this isn't a typical performance measure captured in regular PCP metrics).  src/pmdas/denki/denki.c has some examples of custom units labels for various power measures, for reference.

src/pmdas/amdgpu/help
```diff
@@ -0,0 +1,66 @@
...
+
+@ amdgpu.gpu_clock The GPU clock speed in MHz
+@ amdgpu.gpu_clock_max The maximum GPU clock speed in MHz
+
+@ amdgpu.temperature The GPU temperature in millidegrees Celsius
+The temperature provided is returned in millidegrees celcius, but the
+GPU may not have this precision.
+
+@ amdgpu.load The GPU load
+
```
natoscott, line 65: I'm curious as to what "load" is actually representing?  I see its a 32 bit integer ... possibly a counter ... but I'm not really following.  Is it a utilization (percent?) of some kind, or measure of time, or something else entirely?  Users will ask for sure, so we may as well describe it in details here.

src/pmdas/amdgpu/localdrm.c
```diff
@@ -0,0 +1,323 @@
+/*
```
natoscott, line 1: This file could be renamed drm.c (and likewise the accompanying .h) - we don't need to perform any shared library dlopen/shim gymnastics here like we do with nvidia.

src/pmdas/amdgpu/localdrm.c
```diff
@@ -0,0 +1,323 @@
...
+			     uint32_t *count) {
+  uint32_t amdgpu_count = 0;
+  drmDevicePtr *temp = NULL; /* Will contain all devices, including non-amd ones*/
+  drmDevicePtr *p = NULL; /* Helper to copy AMD device data */
+
+  /* First get the total count of devices */
+  int dev_count = drmGetDevices(NULL, 0);
+
+  if (dev_count <= 0) {
+      printf("No devices\n");
```
natoscott, line 55: The diagnostic printf calls in this file should all become __pmNotifyErr calls which interacts with the logging subsystem a bit more nicely (with timestamp prefixes, etc) - and ensures we write into the log and not stdout (which may even have pmcd on the other end of it waiting for a PDU).

src/pmdas/amdgpu/localdrm.c
```diff
@@ -0,0 +1,323 @@
...
+      if (strcmp(ver->name, "amdgpu")) {
+          drmFreeVersion(ver);
+          continue;
+      }
+
+      /* Copy the AMD GPU data */
+      memcpy(&p[amdgpu_count++], &temp[i], sizeof(drmDevicePtr));
+
+      /* Done with version */
+      drmFreeVersion(ver);
```
natoscott, line 124: Do we need to close the fd (local variable on-stack)?  Looks like it leaks an fd for each device here otherwise.

src/pmdas/amdgpu/localdrm.c
```diff
@@ -0,0 +1,323 @@
...
+
+  memory->total = mem.vram.total_heap_size;
+  memory->usable = mem.vram.usable_heap_size;
+  memory->used = mem.vram.heap_usage;
+
+  releaseAMDDevice(amdgpu_device, fd);
+
+  return DRM_SUCCESS;
+}
+
```
natoscott, line 303: There's alot of opening and closing of device files here, with the way these interfaces are setup (I think this is a bit based on nvidia again, and thats maybe made it more complex than necessary).

Going back to the earlier comment about need_refresh - currently we call all of these APIs on every fetch, for every GPU.  If we are going to do that, we can collapse all of these library calls into a single function that opens and closes each GPU fd a maximum of once per fetch, and avoid system time overheads associated with doing this "for every GPU for every metric" (i.e. ~10x less work if we have 10 metrics).

src/pmdas/amdgpu/pmns
```diff
@@ -0,0 +1,33 @@
...
+    memused			AMDGPU:0:2
+    memtotal			AMDGPU:0:3
+    memfree			AMDGPU:0:4
+    samples			AMDGPU:0:5
+    memusedaccum		AMDGPU:0:6
+    mem_clock			AMDGPU:0:7
+    mem_clock_max		AMDGPU:0:8
+    gpu_clock			AMDGPU:0:9
+    gpu_clock_max		AMDGPU:0:10
+    temperature			AMDGPU:0:11
```
natoscott, line 29: With several of the above metrics, consider a "memory" subtree in the PMNS, e.g. amdgpu.memory.clock and so on (total, free, etc).  Possibly same for amdgpu.gpu.clock and clock_max.

src/pmdas/amdgpu/pmns
```diff
@@ -0,0 +1,33 @@
...
+    load			AMDGPU:0:12
+    avg_pwr			AMDGPU:0:13
```
natoscott, line 31: I'd be inclined to expand avg_pwr to 'average_power' here.

natoscott (commented):
Looking good Fred!

Bunch of minor things described in individual comments.  Beyond that, here's a list of other files that will need changes too:

- configure.ac    (if libdrm is absent, we need to switch this PMDA off in the build - I see there'a a libdrm.pc so I recommend using PKG_CHECK_MODULES like e.g. libsasl)
- src/include/builddefs.in  (makefile macro(s) based on configure.ac mechanism)
- build/rpm/*.spec*  (we'll need a sub-package for this new PMDA)
- qa/xxxx[.out]  (regression test or two, see the apache PMDA test qa/755 as example)  
- qa/admin/package-lists  (list of rpm/deb packages that CI needs to install in order to build/test/release this).

If anything unclear (for sure some will be) - let's chat on slack.

src/pmdas/amdgpu/localdrm.c
fberat: Oops, these are development artefacts, I'll rework/cleanup.

src/pmdas/amdgpu/help
fberat: That seems to be a percentage, I'll add some information.

fberat:
I believe I addressed most of the review finidings, I'll need to go through the code at least one more time though.

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,668 @@
...
+}
+
+static int refresh(pcp_amdgpuinfo_t *amdgpuinfo, uint32_t to_refresh)
+{
+  int i;
+
+  if (!drm_initialized) {
+      setup_gcard_indom();
+      drm_initialized = 1;
+  }
```
natoscott, line 279: I think we should only set drm_initialized here if setup_gcard_indom does not fail?  If it fails, we can try it again next time right?  (maybe an issue has resolved in the meantime)

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,668 @@
...
+ * Wrapper for pmdaFetch which refresh the set of values once per fetch
+ * PDU.  The fetchCallback is then called once per-metric/instance pair
+ * to perform the actual filling of the pmResult (via each pmAtomValue).
+ */
+static int amdgpu_fetch(int numpmid, pmID pmidlist[], pmResult **resp,
+                        pmdaExt *pmda)
+{
+  uint32_t i = 0;
+
+  pmNotifyErr(LOG_ERR, "amdgpu_fetch");
```
natoscott, line 352: Leftover from debugging?  Alternatively, call it conditionally under a pmDebugOptions guard (off by default)

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,668 @@
...
+
+  for (i = 0; i <= numpmid; i++) {
+      unsigned int cluster = pmID_cluster(pmidlist[i]);
+      unsigned int item = pmID_item(pmidlist[i]);
+
+      pmNotifyErr(LOG_ERR, "amdgpu_fetch (%d/%d/%d)", cluster, item, numpmid);
```
natoscott, line 358: Likewise, too verbose by default.

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,668 @@
...
+
+      refresher_list[cluster][item]->needs_refresh = 1;
+  }
+
+  for (i = 0; i < AMDGPU_REFRESHER_COUNT; i++) {
+      if (!amd_refresher[i].needs_refresh)
+	continue;
+      amd_refresher[i].needs_refresh = 0;
+      refresh(&pcp_amdgpuinfo, i);
+      pmNotifyErr(LOG_ERR, "amdgpu_fetch refresh %d", i);
```
natoscott, line 373: Likewise, too verbose by default.

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,668 @@
...
+
+  return pmdaFetch(numpmid, pmidlist, resp, pmda);
+}
+
+static int amdgpu_fetchCallBack(pmdaMetric *mdesc, unsigned int inst,
+                                pmAtomValue *atom) {
+  unsigned int cluster = pmID_cluster(mdesc->m_desc.pmid);
+  unsigned int item = pmID_item(mdesc->m_desc.pmid);
+
+  pmNotifyErr(LOG_ERR, "amdgpu_fetchCallBack (%d/%d/%d)", cluster, item, inst);
```
natoscott, line 384: Too verbose - perhaps under "if (pmDebugOptions.xxx)" conditional?  Run pmdbg(1) --list for options.

src/pmdas/amdgpu/amdgpu.c
natoscott: I see the label callbacks now but should we add "units":"millidegrees celcius" as a label for this one?

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,668 @@
...
+        __pmAFunblock();
+        exit(1); /* fatal if we lose pmcd */
+      }
+      if (pmDebugOptions.appl0)
+        pmNotifyErr(LOG_DEBUG, "completed pmcd PDU [fd=%d]", pmcdfd);
+    }
+    if (autorefresh > 0) {
+      autorefresh = 0;
+      for (int i = 0;i < AMDGPU_REFRESHER_COUNT;i++) {
+	  pmNotifyErr(LOG_ERR, "Refreshing %d", i);
```
natoscott, line 608: Too verbose by default here, I think this is going to end up in a log file once every second?

src/pmdas/amdgpu/drm.c
```diff
@@ -0,0 +1,273 @@
...
+#endif
+#include "drm.h"
+
+#include <libdrm/amdgpu.h>
+#include <libdrm/amdgpu_drm.h>
+#include <xf86drm.h>
+
+#ifndef DSOSUFFIX
+#define DSOSUFFIX "so"
+#endif
```
natoscott, line 31: Pretty sure this agent is Linux-only, so safe to hard-code this if you like.

src/pmdas/amdgpu/drm.c
```diff
@@ -0,0 +1,273 @@
...
+      drmVersionPtr ver = drmGetVersion(fd);
+
+      if (!ver) {
+	  close(fd);
+	  continue;
+      }
+
+      if (strcmp(ver->name, "amdgpu")) {
+	  drmFreeVersion(ver);
+	  continue;
```
natoscott, line 119: I think we may leak an open fd here?

src/pmdas/amdgpu/drm.c
```diff
@@ -0,0 +1,273 @@
...
+      }
+
+      /* Copy the AMD GPU data */
+      memcpy(&p[amdgpu_count++], &temp[i], sizeof(drmDevicePtr));
+
+      /* Done with version */
+      drmFreeVersion(ver);
+  }
```
natoscott, line 127: And also here at the last part of the loop?  Could close it unconditionally right after drmGetVersion perhaps.

config.guess
```diff
@@ -1,12 +1,14 @@
-#! /bin/sh
+#!/usr/bin/sh
```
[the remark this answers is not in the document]
fberat: That actually came with the configure update.
In theory nowadays everything is in /usr. There is a trend to remove /bin and /sbin.

src/pmdas/amdgpu/amdgpu.c
fberat: I need to cleanup, at the end of the day, you were faster than me to review these changes :)

src/pmdas/amdgpu/amdgpu.c
fberat: Likewise, and some other places, removed.

src/pmdas/amdgpu/amdgpu.c
```diff
@@ -0,0 +1,668 @@
...
+      break;
+    default:
+      return PM_ERR_PMID;
+    }
+    break;
+  case MEMORY_CLUSTER: /* Memory related metrics */
+    switch (item) {
+    case AMDGPU_MEMORY_USED:
+      atom->ull = pcp_amdgpuinfo.info[inst].memory.used;
+      pmNotifyErr(LOG_ERR, "Getting used memory %lu", atom->ull);
```
fberat, line 413: Removed.

config.guess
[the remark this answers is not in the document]
fberat: Ok, that's annoying, I may need to check with upstream config.{sub,guess} repo.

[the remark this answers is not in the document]
fberat: I'll investigate further, there is something odd, the file on my system doesn't match the one in the redhat-rpm-config repository.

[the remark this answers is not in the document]
fberat: Ok, found the reason, it looks like there is a systemic modification of shebang in fedora when rpm are built. I'll revert the shebang back to the original value on my next update.

natoscott:
Last handful of small things @fberat ...

amdgpu_fetch()
```
     if (refresher_list[cluster][item] == NULL)
       continue;
```
We have a check that 'cluster' is within range, does 'item' need
a similar check to prevent possible out-of-bounds access? (i.e.
via a maliciously crafted PDU with pmID item >= MAX_ITEM_COUNT).

setup_gcard_indom()

```
pmNotifyErr(LOG_WARNING, "setup_gcard_indom: got %d cards" ...
```
INFO level might be more appropriate for this one?

```
pmNotifyErr(LOG_ERR, "Refreshed memory (%lx)", memory.used);
```
-> leftover temp diagnostic?  (or add pmDebugOptions.appl2 guard)


PMDA README file has the word Readme at the start - intentional?
Just looks a bit odd.  The heading underline could use one more
equals character as well.  :)

qa/1772 has template comment still.  [who are you?] -> Red Hat or
Fred (yourself).  Also need to change this line, e.g.:
```
# test for-some-thing || _notrun No support for some-thing
```
to
```
test -d $PCP_PMDAS_DIR/amdgpu || _notrun No support for AMD GPU metrics
```

fberat:
> Last handful of small things @fberat ...
> 
> amdgpu_fetch()
> 
> ```
>      if (refresher_list[cluster][item] == NULL)
>        continue;
> ```
> 
> We have a check that 'cluster' is within range, does 'item' need a similar check to prevent possible out-of-bounds access? (i.e. via a maliciously crafted PDU with pmID item >= MAX_ITEM_COUNT).

Agreed.

> 
> setup_gcard_indom()
> 
> ```
> pmNotifyErr(LOG_WARNING, "setup_gcard_indom: got %d cards" ...
> ```
> 
> INFO level might be more appropriate for this one?

Agreed.

> 
> ```
> pmNotifyErr(LOG_ERR, "Refreshed memory (%lx)", memory.used);
> ```
> 
> -> leftover temp diagnostic? (or add pmDebugOptions.appl2 guard)

Yes, removing.

> 
> PMDA README file has the word Readme at the start - intentional? Just looks a bit odd. The heading underline could use one more equals character as well. :)

Likely not intentional, removing. Equal character added.

> 
> qa/1772 has template comment still. [who are you?] -> Red Hat or Fred (yourself). Also need to change this line, e.g.:

Red Hat added.

> 
> ```
> # test for-some-thing || _notrun No support for some-thing
> ```
> 
> to
> 
> ```
> test -d $PCP_PMDAS_DIR/amdgpu || _notrun No support for AMD GPU metrics
> ```
Done.

Thanks for the review, I'll push the update.

PR #2009, opened 2024-07-08
proc: add amdgpu support

fberat:
This is a first shot for new AMD GPU process metrics.
This patch collects information retrieved from `/proc/<pid>/fdinfo`.

The current implementation accumulates data for each process, walking through all file descriptors and looking for drm and amd entries.

A future patch may be considered to separate the data per drm client ID.

### DIFF
--- a/qa/1222
+++ b/qa/1222
@@ -58,6 +58,8 @@ _filter()
 	-e '/^proc\.id\.container: Missing metric value(s)/d' \
 	-e '/^proc\.psinfo\.labels: No value(s) available/d' \
 	-e '/^proc\.psinfo\.ngid: Metric not supported by this version/d' \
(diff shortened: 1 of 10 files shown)


fberat:
Rebased on main, description amended.

natoscott:
Looking good Fred.  There's something not-quite-right causing the help text to go missing.  I haven't been able to pin it down though (there's some tweaks needed in root_proc but that's not enough it seems).  I'll send you a patch of what I have, maybe you can see what I'm missing here.

We'll also probably need some filtering in common QA routines, similar to the way [hot]proc.smaps.* is done.

natoscott:
@fberat thanks Fred - I've made some QA fixups on top of this (please double check), but there may be others I've missed (please keep an eye on CI results tomorrow).

PR #2018, opened 2024-07-12
proc: Fix resource leak in fdinfo processing

fberat:
Fix coverity reported issues:

- Close dirp appropriately when returning from refresh_proc_pid_fdinfo.
- Pass proc_pid_fdinfo_t struct as a pointer to accumulate_proc_fdinfo.


natoscott:
@fberat @kmcdonell I think you both fixed this issue, can you review & see if this PR is still needed?  taa.

kmcdonell:
The closedir() changes look the same in both commits, but we need some of my other changes (debug, readdir() loop issue) and some of @fberat 's changes (Coverity ones).
Let me try and resolve the conflicts in this PR.

PR #2033, opened 2024-07-30
libpcp: auto-load the proc PMDA in local context mode

natoscott:
We can treat pmda_proc.so the same as pmda_linux.so in the local context auto-loading code paths since its in exactly the same boat when it comes to default pmcd.conf state.

This resolves a problem report where pcp-htop reports some kernel metrics but the per-process stats are missing, when its fallback-to-local-without-pmcd mode occurs.


kmcdonell (approved):
LGTM

PR #2052, opened 2024-08-28
qa/valgrind-suppress: external libbpf.so does not fully initialize attr

wcohen:
The qa/1900 test makes use of libbpf.  When libbpf probes the linux kernel to check which features are available one of the probes does not fully initialize all the members of the attr struct. Adding suppression entries to qa/valgrind-suppress to catch those specific cases.  An upstream issue has been filed on libbpf to get this addressed:

   https://github.com/libbpf/libbpf/issues/845

### DIFF
--- a/qa/1900
+++ b/qa/1900
@@ -46,6 +46,39 @@ _filter_logs()
     # end
 }
 
+cat >$tmp.suppress <<End-of-File
+# The libbpf probes to test for whether the kernel supports certain features
+# do not fully initialize the attr struct passed in the syscall.
+# This is outside the control of PCP.  There is an upstream libbpf issue filed:
+#
+#   https://github.com/libbpf/libbpf/issues/845
+#
+# The following suppressions are for qa/1900
+{
+   libbpf missing attr missing expected_attach_type field initialization
+   Memcheck:Param
+   bpf(attr->expected_attach_type)
+   fun:syscall
+   fun:sys_bpf
+   fun:sys_bpf_fd
+   fun:sys_bpf_prog_load
+   fun:probe_kern_prog_name
+   fun:kernel_supports
+   ...
+}
+{
+   libbpf missing attr missing prog_ifindex field initialization
+   Memcheck:Param
+   bpf(attr->prog_ifindex)
+   fun:syscall
+   fun:sys_bpf
+   fun:sys_bpf_fd
+   fun:sys_bpf_prog_load
+   fun:probe_kern_prog_name
+   fun:kernel_supports
+   ...
+}
+End-of-File
 
 # preliminaries ...
 # - install bpf PMDA so we get bpf in the default PMNS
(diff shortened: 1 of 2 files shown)


natoscott:
@wcohen I'm wondering if we should localize this suppression to just qa/1900 if it only affects that test case?  the qa/valgrind-suppress file tends to be for things that need to cover all PCP tools (usually due to libs linked by libpcp for example).

I think (from a very quick look, maybe wrong) that an additional --suppressions argument can be passed to _run_valgrind - so perhaps a 'here' document could be used to keep all of this inside qa/1900?  (qa/828 does something similar, but not exactly the same as it uses valgrind directly)

natoscott:
valgrind man page says:

`       --suppressions=<filename> [default: $PREFIX/lib/valgrind/default.supp]
           Specifies an extra file from which to read descriptions of errors to
           suppress. You may use up to 100 extra suppression files.
`

wcohen:
@natoscott What would be the advantage of making a special valgrind-suppress file for qa/1900?  Looking through the qa/valgrind-suppress file there are several fixes for other libraries issues (nss, cyrus-sasl) already in there.

natoscott:
> What would be the advantage of making a special valgrind-suppress file for qa/1900?

It'd just be a temporary file (as in the example I gave) not a special file checked in to the git repo - the advantage is it keeps the workaround close to the problem its working around.  The next person to be debugging pmdabpf failures in this test in a year or two will then have more context.

> there are several fixes for other libraries issues (nss, cyrus-sasl) already in there.

These libraries are linked to libpcp.  So these valgrind suppressions are needed for every PCP binary, so their workarounds are centrally located and not duplicated throughout the testsuite.  This is different to libbpf which is used only by pmdabpf.

wcohen:
The _run_valgrind() function doesn't allow passing in arbitrary valgrind arguments.  Looks like qa/1900 needs to take an approach like qa/1290 which has some additional arguments over the regular _run_valgrind().  At one time 1290 had additional suppression, but that has been removed.

natoscott:
Yep, good point - we could add --suppressions as a third special case at the start of _run_valgrind to handle this.

kmcdonell:
@natoscott @wcohen There is already a hook for per-test valgrind suppressions it is $grind_extra.  It is already used in several tests, see for example qa/1316.

wcohen:
Thanks for the suggestions, @natoscott and @kmcdonell .  I have reworked the patch to have a local valgrind suppress file and use the $grind_extra.  I have put the revised patch on https://github.com/wcohen/pcp/tree/wcohen/valgrind_suppress1900a.  Is there an easy way to revise the patch in this existing pull request or should I close this pull request and open on on the other branch?

natoscott:
@wcohen easiest is to just add a second commit and git-push it into this same branch (same PR).  You can also change the original commit if you're comfortable with the git reset/rebase fun and games and then --force git-push here.  Either way, no need for a new PR.

wcohen:
@natoscott  the branch referenced in this pull request now has the revised patch in it that addresses the earlier comments.

PR #2058, opened 2024-09-05
pmdalinux: sysfs fallback from hinv.cpu.clock metric

### LINKED ISSUE #2047: [RFE] broader platform support for hinv.cpu.clock metric
@christianhorn noticed recently that non-x86_64 platform's /proc/cpuinfo file lacks the information that backs the popular clock rate metric.  However, sysfs files below /sys/devices/system/cpu/cpuN/cpufreq/scaling_cur_freq do contain this information for other platforms - and we already have code reading these sysfs files for other metrics - it'd be good to use these as an alternate source for this metric.

natoscott:
Resolves #2047


christianhorn:
Works great, thanks! Verified on Raspi4/Debian bookworm.

PR #2060, opened 2024-09-06
qa: relax the checks on parent directories permissions

natoscott:
Current releases of Fedora have tighter default settings for home directory permissions, which is tripping common script warning code when running QA tests; e.g.

Warning: parent directory /home/nathans (mode:drwx------.) not world readable and searchable

AFAICT this sseems to be too strict however and checking for owner read/exec suffices.

### DIFF
--- a/qa/common
+++ b/qa/common
@@ -937,10 +937,10 @@ else
 	__prot=`ls -ldL "$__x" | sed -e 's/[ 	].*//'`
 	case $__prot
 	in
-	    dr?xr?xr?[xt]*)
+	    dr?x*)
 		;;
 	    *)
-		echo "Warning: parent directory $__x (mode:$__prot) not world readable and searchable"
+		echo "Warning: parent directory $__x (mode:$__prot) not owner readable and searchable"
 		;;
 	esac
 	__x=`dirname $__x`


kmcdonell (commented):
Can't you just chmod $HOME?  This is a QA environment after all.
I don't recall the exact circumstances that prompted this, but I think it was something like ...
here=`pwd`
cd /some/place/else 
do qa stuff
$sudo -u someuser some command $here/someplace
Which will not work if $HOME (and all the subdirs down to to $here) is not searchable by someuser.

Of course, you can make the change and if stuff breaks we'd have a better *real* justification for the original code.

natoscott:
> Can't you just chmod $HOME? This is a QA environment after all.

Well, yep - in this case it's just my laptop.  However, its just one more thing in the way of people running QA for the first time.

> I don't recall the exact circumstances that prompted this, but I think it was something like ... here=`pwd` cd /some/place/else do qa stuff $sudo -u someuser some command $here/someplace Which will not work if $HOME (and all the subdirs down to to $here) is not searchable by someuser.
> 

Yeah, I figured its something like that but I cannot seem to get the failure to happen after trying various things.  And no sign of failed tests resulting from this (but could be one of the _notrun tests is affected of course).  It may be better to isolate the warning to just affected tests, if we do find a reproducer to the original issue.

> Of course, you can make the change and if stuff breaks we'd have a better _real_ justification for the original code.

+1

PR #2061, opened 2024-09-06
pcp-meminfo: added timestamp string to make it similar to other utility

orasagar:
changed timestamp format as well to print date along with time
Signed-off-by: sagar sagar <sagar.sagar@oracle.com>


natoscott (commented):
@orasagar looks good - does this affect any QA tests?

Issue #2062, opened 2024-09-09
pmdastatsd segfaults in pmdaRehash under a bigger load


pmoravec:
When sending a few thousands of metrics via `statsd` to PCP, I get random segfault of `pmdastatsd`. I tuned my reproducer to hit this every 30 minutes on average. Reproducer requires foreman+katello+pulp installed, so hard to reproduce on your own, I think. But I can provide full 250MB coredump if required.

Backtrace of the segfaulting thread:

    (gdb) bt full
    #0  pmdaRehash (pmda=pmda@entry=0x55997bf54910, metrics=<optimized out>, nmetrics=<optimized out>) at open.c:519
            extp = 0x55997bf549e0
            hashp = 0x55997bf549f0
            metric = 0x7f2d2ba5e010
            buf = " \221\370{\231U\000\000\360\216\366{\231U\000\000\360#\371{\231U\000\000\205\230\230?-\177\000"
            m = 4992
    #1  0x000055997a75926b in statsd_possible_reload (pmda=pmda@entry=0x55997bf54910) at pmda-callbacks.c:462
            data = 0x55997a961b00 <data>
            need_reload = <optimized out>
    #2  0x000055997a759b4b in statsd_fetch (num_pm_id=340, pm_id_list=0x55997bfb452c, resp=0x7ffeb0b69b98, pmda=0x55997bf54910) at pmda-callbacks.c:685
            data = 0x55997a961b00 <data>
    #3  0x00007f2d3f985d69 in __pmdaMainPDU (dispatch=0x7ffeb0b69c40) at mainloop.c:175
            pb = 0x55997bfb4510
            sts = 0
            psts = <optimized out>
            op_sts = 0
            pmid = 239076366
            desc = {pmid = 239076366, type = 5, indom = 239075330, sem = 1, units = {pad = 0, scaleCount = 0, scaleTime = 0, scaleSpace = 0, dimCount = 0, dimTime = 0, dimSpace = 0}}
            npmids = 340
            pmidlist = 0x55997bfb452c
            namelist = 0x0
            name = 0x55997bf89140 "\340\220\370{\231U"
            offspring = 0x0
            statuslist = 0x0
            subtype = 0
            result = 0x55997bfc6d20
            ctxnum = 0
            length = 0
            when = {tv_sec = 0, tv_usec = 0}
            ident = 239076700
            type = 6
            indom = 239075330
            inst = 0
            iname = 0x0
            inres = 0x55997bf92340
[...]

PR #2069, opened 2024-09-14
fix(statsd): prevent uncommitted metrics from being mapped to PCP

Erbenos:
Because of uninitialized "committed" flag of internal representation of statsd metric in aggregator-metrics.c `int create_metric(struct agent_config* config, struct statsd_datagram* datagram, struct metric** out)`, counterpart pmdaMetric* could have been created for it in pmda-callbacks.c "create_pcp_metric `static void create_pcp_metric(char* key, struct metric* item, pmdaExt* pmda)`, cause it may have passed the check at pmda-callbacks.c `static void map_metric(char* key, struct metric* item, void* pmda)`.

Internal representation of the statsd metric is the "struct metric* item" argument.

Processing a statsd metric happens in multiple steps, relevant for this is part of thread responsible for aggregation:

Screenshot 2024-09-15 at 1 14 40

add_metric, process_labeled_datagram, remove_metric and mark_metric_as_comitted (excuse grammar) are all synchronized by single mutex which is also contested by the thread handling the pcp callbacks. This mutex is unlocked before each of them returns.

In the thread handling pcp callbacks:

Screenshot 2024-09-15 at 1 20 46

need_reload read and statsd_map_stats contest same mutex mentioned above. Hence there may be the following sequence of calls that results in metrics being mapped into PCP representation (pmdaMetric*) whose internal representation is struct metric*, even though it never should have and since its data can be freed/removed at an arbitrary later point (with pmdaMetric pointing to some of it), I imagine this can cause issues. 

Such a sequence may be:

create_metric (aggregator) -> add_metric (aggregator) -> statsd_possible_reload determines that reload is needed (handling pcp callbacks) -> statsd_map_stats maps the internal metrics to pcp ones (handling pcp callbacks) -> process_labeled_datagram (agregator) -> remove_metric (aggregator)

datagram received which may trigger such a sequence is:

"metric_name,x=1:-2|c"

(parsing succeeds, but -2 is invalid value for a counter type, so process_labeled_datagram returns 0 and metric will be cleaned up with remove_metric)

I wasn't able to simulate a segmentation error in the pmdaRehash and I wasn't able to run 1720 qa test due to some issues with python, CI should run it and won't find an issue, hopefully.

Other changes are related to other errors I was able to simulate, eg.: metric above would lead to freeing a memory that wasn't alocated with malloc.

As noted in Slack @natoscott if you got some clue as to how to verify the correctness of the above, I would be grateful for a consultation as well as overall review, cause I have not written or read an ounce of C in quite a while.

Likely related #2062

### DIFF
--- a/src/pmdas/statsd/src/aggregator-metric-counter.c
+++ b/src/pmdas/statsd/src/aggregator-metric-counter.c
@@ -24,7 +24,11 @@
 #include "utils.h"
 
 /**
- * Creates counter value in given dest
+ * Creates counter value in given destination
+ * @arg config - Config
+ * @arg datagram - Data to extract value from
+ * @arg out - Dest pointer
+ * @return 1 on success, 0 on fail. Should NOT allocate "out" when 0 is returned
  */
 int
 create_counter_value(struct agent_config* config, struct statsd_datagram* datagram, void** out) {
--- a/src/pmdas/statsd/src/aggregator-metric-duration.c
+++ b/src/pmdas/statsd/src/aggregator-metric-duration.c
@@ -26,7 +26,11 @@
 #include "utils.h"
 
 /**
- * Creates duration value in given dest
+ * Creates duration value in given destination
+ * @arg config - Config from which we know what duration
+ * @arg datagram - Data to extract value from
+ * @arg out - Dest pointer
+ * @return 1 on success, 0 on fail. Should NOT allocate "out" when 0 is returned
  */
 int 
 create_duration_value(struct agent_config* config, struct statsd_datagram* datagram, void** out) {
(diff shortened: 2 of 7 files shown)


pmoravec:
I was testing a patched source delivered by @natoscott , but still I see segfaults:

    Sep 16 13:01:55 pmoravec-sat615 kernel: pmdastatsd[518864]: segfault at 7fb720c3f018 ip 00007fb728e968d6 sp 00007ffe0e48c4d0 error 4 in libpcp_pmda.so.3[7fb728e8d000+19000]

I will try to get backtrace from it, but so far running `pmdstatsd` under `gdb`, no coredump happened.. Anyway, I will edit the comment once I succeed.

Erbenos:
Even if this doesn't address the exact issue you were having, think it should be merged anyway as it fixes different ones for sure. If nothing else it narrows it down.

natoscott:
| [...] think it should be merged anyway
Will do.

PR #2072, opened 2024-09-17
fix(statsd): keep correct track of pmns mapped metric count

Erbenos:
Likely related #2062.

This is likely it, no race conditions, just me not being able to count well. I am kind of surprised that it took such large metric counts

> It seems tome the rehash was slowly happening at a time some other thread updated the hash table (the for (m = 0; m < pmda->e_nmetrics; m++) { was processing 4992th metric out of 5091)?

for it to segfault when it indexed out of bounds in pmdaRehash. Especially cause across multiple passes roughly speaking even after half of that number you would start reaching into memory I didn't allocate.

### DIFF
--- a/src/pmdas/statsd/src/pmda-callbacks.c
+++ b/src/pmdas/statsd/src/pmda-callbacks.c
@@ -134,6 +134,7 @@ create_pcp_metric(char* key, struct metric* item, pmdaExt* pmda) {
         "STATSD: adding metric %s %s from %s\n", item->meta->pcp_name, pmIDStr(item->meta->pmid), item->name
     );
     item->meta->pcp_metric_index = i;
+    data->pcp_metric_count += 1;
 }
 
 /***
@@ -366,7 +367,6 @@ map_metric(char* key, struct metric* item, void* pmda) {
     if (item->meta->pcp_instance_change_requested == 1) {
         update_pcp_metric_instance_domain(key, item, (pmdaExt*)pmda);
     }
-    data->pcp_metric_count += 1;
     process_stat(data->config, data->stats_storage, STAT_TRACKED_METRIC, (void*)item->type);
     VERBOSE_LOG(1, "Populated PMNS with %d, %s .", item->meta->pmid, item->meta->pcp_name);
     pmdaTreeInsert(data->pcp_pmns, item->meta->pmid, item->meta->pcp_name);


src/pmdas/statsd/src/pmda-callbacks.c
```diff
@@ -134,6 +134,7 @@ create_pcp_metric(char* key, struct metric* item, pmdaExt* pmda) {
         "STATSD: adding metric %s %s from %s\n", item->meta->pcp_name, pmIDStr(item->meta->pmid), item->name
     );
     item->meta->pcp_metric_index = i;
+    data->pcp_metric_count += 1;
```
Erbenos, line 137: This incremented total mapped metric count by 1, for each rebuild of PMNS, meaning that it also counted metrics that already had their counterpart pmdaMetric structures in `pmdaRehash(pmda, data->pcp_metrics, data->pcp_metric_count);`'s data->pcp_metrics.

src/pmdas/statsd/src/pmda-callbacks.c
```diff
@@ -366,7 +367,6 @@ map_metric(char* key, struct metric* item, void* pmda) {
     if (item->meta->pcp_instance_change_requested == 1) {
         update_pcp_metric_instance_domain(key, item, (pmdaExt*)pmda);
     }
-    data->pcp_metric_count += 1;
```
Erbenos, line 369: This incremented total mapped metric count by 1, for each rebuild of PMNS, meaning that it also counted metrics that already had their counterpart pmdaMetric structures in pmdaRehash(pmda, data->pcp_metrics, data->pcp_metric_count);'s data->pcp_metrics.

natoscott: Good find @Erbenos !  I'm out today but will try it tomorrow on the reproducer system we have here.

natoscott: @Erbenos this is looking good, the reproducer setup has been running for 10+ minutes now without failure 🥇 💯

natoscott: Only other thing I noticed here is that the statsd PMDA is pretty chatty - I'm not sure how normal a load this is but after 10-15 minutes the PMDA log file is already ~13MB large (over a day or two this is going to be causing some folks ENOSPC problems).  So we probably should reduce that ...I suggest moving some of the detail to under a debug flag, and just keep this one info-level line on reconfiguration:
```
[Thu Sep 19 02:18:24] pmdastatsd(1006338) Info: statsd: pmdastatsd: 920 metrics and 3 instance domains after reload
```

Here's what I'm seeing currently...
```
ls -lh /var/log/pcp/pmcd/statsd.log
-rw-r--r--. 1 root root 13M Sep 19 02:18 /var/log/pcp/pmcd/statsd.log
```

Erbenos: Depends on the verbosity set, can always be tweaked if necessary if a specific verbosity level feels like it has unjustifiable footprint

natoscott: Yeah, based on what I saw yesterday the defaults need tweaking - default should be more like one line of output on reconfiguration and verbose output is opt-in for when deeper diagnostics / debugging is required.

Issue #433, opened 2018-01-31
Interpolation not effective in some cases


myllynen:
As discussed in https://groups.io/g/pcp/topic/6111366, when creating an archive with pmlogger(1) the prologue is logged ~0.02 seconds before the actual data and we see for example with pmdumptext(1) shows results being 10 minutes off and misses some records in the end. I'm providing a copypaste and a test archive to demonstrate this issue below.

```
$ pmdumplog -z -a test                                          
Note: timezone set to local timezone of host "test" from archive

Log Label (Log Format Version 2)
Performance metrics from host test
    commencing Fri Oct 20 10:36:00.416347 2017
    ending     Fri Oct 20 11:26:00.436134 2017
Archive timezone: EEST-3
PID for pmlogger: 6479

Descriptions for Metrics in the Log ...
PMID: 60.1.2 (mem.util.free)
    Data Type: 64-bit unsigned int  InDom: PM_INDOM_NULL 0xffffffff
    Semantics: instant  Units: Kbyte
PMID: 2.3.2 (pmcd.pmlogger.archive)
    Data Type: string  InDom: 2.1 0x800001
    Semantics: discrete  Units: none
PMID: 2.3.0 (pmcd.pmlogger.port)
    Data Type: 32-bit unsigned int  InDom: 2.1 0x800001
    Semantics: discrete  Units: none
PMID: 2.3.3 (pmcd.pmlogger.host)
    Data Type: string  InDom: 2.1 0x800001
    Semantics: discrete  Units: none

Instance Domains in the Log ...
InDom: 2.1
10:36:00.416347 1 instances
   6479 or "6479"

Temporal Index
		Log Vol    end(meta)     end(log)
10:36:00.416347	      0          132          132
11:26:00.436134	      0          403          528

[136 bytes]
10:36:00.416347 3 metrics
    2.3.3 (pmcd.pmlogger.host): inst [6479 or "6479"] value "localhost"
    2.3.0 (pmcd.pmlogger.port): inst [6479 or "6479"] value 4330
    2.3.2 (pmcd.pmlogger.archive): inst [6479 or "6479"] value "/home/testuser/pcptest/logger/test"

[52 bytes]
10:36:00.435851 1 metric
    60.1.2 (mem.util.free): value 6925664

[52 bytes]
10:46:00.436619 1 metric
    60.1.2 (mem.util.free): value 3764672

[52 bytes]
10:56:00.436598 1 metric
[...]

PR #2078, opened 2024-09-20
pcp-buddyinfo: implementation of no interpolation option

orasagar:
Implemented -u option in pcp buddyinfo to disable interpolation to view archive data .

Following are enhance performance  observed 

In no interpolation mode 
time pcp buddyinfo -a /data/minimum_data/2hr_data/20240825_2hr_data.0 -u
real	0m0.917s
user	0m0.658s
sys	0m0.264s

In interpolation mode 
time pcp buddyinfo -a /data/minimum_data/2hr_data/20240825_2hr_data.0
real	0m7.284s
user	0m3.253s
sys	0m0.662s

Fixed broken pipe issue as well with pcp buddyinfo .


natoscott (changes requested):
@orasagar looks OK - an update for qa/1990 to exercise this new option too would be good.

orasagar:
@natoscott  I have added the no interpolation test case

kmcdonell:
@orasagar and @natoscott I understand the laudable goal of making things run faster, but I have some concerns here ...

- this changes the semantics of what is reported by buddyinfo, in some subtle ways that may not be obvious to the end user, specifically the "reporting" time is driven entirely by the timestamps in the archive, which may or may not be regular and for some archives this may produce misleading data by not reporting values for some instances and/or sample periods
- I think this sets a dangerous precedent ... in general terms, non-interpolation mode is NOT a faster version of interpolation mode

I've only looked quickly at pcp-buddyinfo, but it appears the metrics it needs are singular or from small fixed-sized instance domains, and these exclude all of the subcases where interpolation mode may be expensive, so I'd really like to understand what's going on here, because I think there is a lurking issue that needs to be fixed in libpcp, and -u for pcp-buddyinfo is wallpapering over that crack.

Is there any chance I can get my hands on your test case archive, or a "recipe" to recreate the archive locally?

@natoscott At first blush I don't think this is related to your htop performance issue with archives, as in that case I think the problem is more likely to lie with the dynamic instance domain under the proc.* metrics, and I've been thinking about that one too, but possible "fixes" there would not apply to pcp-buddyinfo unless there is something I'm missing entirely about the latter.

kmcdonell:
And one small nit, could the option be -U (upper case) rather than -u, so that it matches pmval, which is the only other tools that knows about non-interpolate mode for archives?

natoscott:
@kmcdonell there is one other precedent and that is pmrep, which uses -u/--no-interpol - I don't really mind regarding -U/-u except to note that with pmval -U an option is required (archive name), whereas pmrep sticks with -a/--archive.
This is all unrelated to pcp-htop, which does not yet support archives (remains a wish list to-do item).

kmcdonell:
Ok stick with lowercase -u

orasagar:
@kmcdonell thank you for your feedback. I appreciate the insights and concerns you’ve raised.

Here with non-interpolation mode we are trying to give user to view the data at the rate which it has been recorded in archive .But the data will be reported for for all instances.

We are not trying to portrait  non-interpolation mode as a faster mode but we are giving an option to user to view data at the rate which it has been recorded .

kmcdonell:
@orasagar understood, but still ...
```
Is there any chance I can get my hands on your test case archive, or a "recipe" to recreate the archive locally?
```

orasagar:
@kmcdonell the test archive is from large system where we have around 120 cpus and around 3000 processes ,in our pmlogger config all the proc related metrics are captured at 30 sec frequency and most other system metrics such as hinv, kernel are captured at 10 sec frequency.

I agree that in case of pcp-buddyinfo instances are of fixed domain size but actually something like for pcp-ps where the data is increasing with the number of processes, Interpolation vs non-interpolation might make a lot of difference.

I will be eager to know your opinion on this.

orasagar:
I implemented this option because I noticed pcp-iostat already has this option ...... and I'm planning to implement the same in pcp-ps tool as well.

Let me know your thoughts.

myllynen:
We're perhaps getting a bit sidetracked here but just few quick comments. I suspect the pmrep(1) -u option was chosen to differentiate from the pmval(1) -U option which indeed requires the archive name specified however users might remember that particular letter being related to interpolation with different tools.

More importantly, I just checked that the interpolation issue reported https://github.com/performancecopilot/pcp/issues/433 seems to be still affecting some tools. Not sure is there anything needed for pcp-buddyinfo or pcp-ps due to that but it might be a good idea to check if they are affected. Which I think is likely since AFAICS this is something on the libpcp level.

Here's a quick reproducer with 6.3.1, you can of course adjust the interval and end time to have the archive created quicker but I'm used 10min interval to really highlight the issue:

```
$ cat pmlogger.conf
log advisory on default { mem.util.free }
$ pmlogger -c ./pmlogger.conf -T 1h -t 10min test
$ pmlogdump -z -a test | grep -A 1 "1 metric"
08:15:13.411271635 1 metric
    60.1.2 (mem.util.free): value 3372220
--
08:25:13.411444328 1 metric
    60.1.2 (mem.util.free): value 3368616
--
08:35:13.411577350 1 metric
    60.1.2 (mem.util.free): value 3371636
--
08:45:13.411602286 1 metric
    60.1.2 (mem.util.free): value 3369240
--
08:55:13.411359464 1 metric
    60.1.2 (mem.util.free): value 3318828
--
09:05:13.411843555 1 metric
    60.1.2 (mem.util.free): value 3311652
$ pmval -z -a test -t 10min mem.util.free
Note: timezone set to local timezone of host "f40-server.example.com" from archive

metric:    mem.util.free
archive:   test
host:      f40-server.example.com
start:     Wed Sep 25 08:15:13 2024
end:       Wed Sep 25 09:15:13 2024
semantics: instantaneous value
units:     Kbyte
samples:   7
interval:  600.00 sec
08:15:13.400  No values available
08:25:13.400              3372220
08:35:13.400              3368616
08:45:13.400              3371636
08:55:13.400              3369240
09:05:13.400              3318828
09:15:13.400  No values available
$ pmval -z -U test -t 10min mem.util.free
Note: timezone set to local timezone of host "f40-server.example.com" from archive

metric:    mem.util.free
archive:   test
host:      f40-server.example.com
start:     Wed Sep 25 08:15:13 2024
end:       Wed Sep 25 09:15:13 2024
semantics: instantaneous value
units:     Kbyte
samples:   all
08:15:13.411              3372220
08:25:13.411              3368616
08:35:13.411              3371636
08:45:13.411              3369240
08:55:13.411              3318828
09:05:13.411              3311652
$ pmrep -z -a test -t 10min -p mem.util.free   
          m.u.free
             Kbyte
08:15:13       N/A
08:25:13   3372220
08:35:13   3368616
08:45:13   3371636
08:55:13   3369240
09:05:13   3318828
09:15:13       N/A
$ pmrep -z -a test -t 10min -p -u mem.util.free
          m.u.free
             Kbyte
08:15:13   3372220
08:25:13   3368616
08:35:13   3371636
08:45:13   3369240
08:55:13   3318828
09:05:13   3311652
```

Observe for example the metric value reported by pmlogdump(1) on 08:45 ending ..9240 and how that particular value is reported either at the correct time or off-by-interval depending on whether interpolation is used or not.

I think possible investigation of this issue should continue in https://github.com/performancecopilot/pcp/issues/433 but I wanted to point out it here as in some cases at least as of today there are benefits to check output without interpolation enabled. Thanks.

kmcdonell:
This extended example from @myllynen shows why -u|-U is so dangerous ... it allows misleading information to be reported without warning.  For more complicated archives and more complicated metric semantics, the situation is much worse.  So far in the issues reported, the difference between interpolation and non-interpolation are small because of the simplicity of the test archives and the metrics chosen, the real situations (especially with dynamic indoms, like the proc.* metrics) the potential difference in reported values is **much** more severe.

Remember, one of the design goals (that has not changed in 30 years) is that an archive source of metrics should be as close as possible to a host source of metrics in terms of semantics and data values ... interpolation is part of what makes that work, non-interpolation does not.

If the driver here is performance when replaying archives, let's focus on ways of addressing that while preserving the semantics, not weakening the semantics and potentially reporting misleading information for the sake of cycles.

I don't think #433 is a real issue, and I've posted a long explanation there.

I think we need a white paper to describe the issues around non-interpolated used of PCP archives, or perhaps another section added to https://pcp.io/papers/archive-interpolation.pdf where value interpolation is explained.

Also, any tool that wants to go down the -u path, needs IMHO to have a whacking big caveat in the man page warning that this may be faster, but the reported results maybe misleading.

orasagar:
@kmcdonell  now i understood the issue .
But from other perspective if you see -u options gives same information as pmdumplog but just in a cleaner way.

I do agree to your point that there is some discrepancy in values with or without interpolation.

natoscott:
@orasagar there's a merge conflict here - could you take a look?  Also, if we're going to go ahead with this I think last thing needed is the man page caveat (Ken, could you suggest some words?  These should be mirrored in pmrep man page too IMO).

| But from other perspective if you see -u options gives same information as pmdumplog but just in a cleaner way.

That's understating the problem a bit.  There are pathological situations (esp. for rate-converted counters) where values can exist in an archive, e.g. recorded at slightly times by being in different logging groups, but no/missing values will be reported by tools like pmrep in no-interpolate mode.

src/pcp/buddyinfo/pcp-buddyinfo.py
```diff
@@ -161,22 +167,47 @@ def report(self, manager):
...
+        if opt == 'u':
+            BuddyinfoOptions.uflag = True
+
     def __init__(self):
-        pmapi.pmOptions.__init__(self, "a:s:Z:zV?")
+        pmapi.pmOptions.__init__(self, "a:s:Z:uzV?")
         self.pmSetLongOptionHeader("General options")
+        self.pmSetOptionCallback(self.extraOptions)
+        self.pmSetLongOptionArchive()
+        self.pmSetLongOption("no-interpolation", 0, "u", "", "disable interpolation mode with archives")
```
natoscott, line 184: @orasagar could you make the long-option here match with pmrep too?  (--no-interpol)

natoscott (commented):
@orasagar I noticed also the commit message has a typo in "Implemanation"

src/pcp/buddyinfo/pcp-buddyinfo.py
orasagar: done

natoscott:
@orasagar thanks for the updates.  Re-reading this, I think there's still an outstanding request here to provide an archive exhibiting the problematic performance when replaying pcp-buddyinfo.  Did you have any success with pmlogredact since last we discussed it?  IIRC, there was a local configuration issue on your machine but that was then resolved - is that right?  Thanks.

orasagar:
@natoscott @kmcdonell Please find the attached archive 
[archive_data.tar.gz](https://github.com/user-attachments/files/17436873/archive_data.tar.gz)

Issue #1915, opened 2024-03-13
pmdabpf sets a too-low limit for rlimit


jasonk000:
Making the same mistake as `640KB ought to be enough for anybody`, I picked a 100MB rlimit setting during `bpf.c` setup. Now, when another (non-PCP) bpf application is loaded, we have a conflict.

Proposal:
- If the existing memlock rlimit value is greater than the 100MB target, do not try to change it.
- Alternatively, we could simply ignore rlimit and document the failure / push it to user responsibility (eg: systemd unit).

Reproduce:
- Start _other_ bpf applications that use > 100MB of space
- Set ulimit -l unlimited for the root user
- Start dbpmda and open the pmda
- - The PMDA sets rlimit 100MB
- dbpmda fails to start

With another program running, consuming ~350MB of maps and progs
```
(root) /home/user # ulimit -l unlimited

(root) /home/user # dbpmda
dbpmda> open dso /usr/lib/pcp/pmdas/bpf/pmda_bpf.so bpf_init 157
[Wed Mar 13 18:34:52] dbpmda(4849) Info: loading caches
[Wed Mar 13 18:34:52] dbpmda(4849) Info: loading configuration: /var/lib/pcp/pmdas/bpf/bpf.conf
[Wed Mar 13 18:34:52] dbpmda(4849) Info: loaded configuration: /var/lib/pcp/pmdas/bpf/bpf.conf
[Wed Mar 13 18:34:52] dbpmda(4849) Info: setrlimit RMLIMIT_MEMLOCK ok
[Wed Mar 13 18:34:52] dbpmda(4849) Info: loading modules
[Wed Mar 13 18:34:52] dbpmda(4849) Info: loading modules
[Wed Mar 13 18:34:52] dbpmda(4849) Info: loading runqlat.so
[Wed Mar 13 18:34:52] dbpmda(4849) Info: loading: runqlat.so from /var/lib/pcp/pmdas/bpf/modules/runqlat.so
[Wed Mar 13 18:34:52] dbpmda(4849) Info: initialising runqlat.so
[Wed Mar 13 18:34:52] dbpmda(4849) Info: booting: runqlat_bpf
[...]

PR #2091, opened 2024-10-15
Avoid too-low rlimit limit for pmdabpf

wcohen:
On older Linux kernels (pre 5.11) running multiple BPF applications may exceed the locked pages limit (RLIMIT_MEMLOCK) imposed in pmdabpf and prevent additional BPF applications from starting.  The pmdabpf has been changed to remove any locked page limit to elimitate this arbitrary limit interferring with BPF application startup.  This patch is based on the proposed patch from Nathan Scott in the issue.

Resolves: https://github.com/performancecopilot/pcp/issues/1915


natoscott (commented):
@wcohen LGTM, thanks.

PR #2094, opened 2024-10-17
pcp-ps: corrected column typo in error message

orasagar:


natoscott (commented):
@orasagar uhm, your change to the spelling of "column" here is incorrect.

Issue #1998, opened 2024-06-07
pcp2elasticsearch accesses a deprecated/removed API endpoint


jtrokel:
I am trying to use pcp2elasticsearch-5.2.5 to send PCP logs to an Elasticsearch-8.12.2 instance. However, trying to run it out of the box fails to upload anything to Elasticsearch but does not show any errors. Editing the source code for this program to add a few debugging print statements shows that a POST request is being made to <Elasticsearch-server>/pcp/pcp-metric. Using cURL to send a POST request constructed the same way gives back an error:
```
$ curl -i -X POST <es-server>/pcp/pcp-metric
    
  HTTP/1.1 400 Bad Request
  X-elastic-product: Elasticsearch
  content-type: application/json
  content-length: 72

  {"error":"no handler found for uri [/pcp/pcp-metric] and method [POST]"}
```

I could not find anything in the Elasticsearch API reference about an endpoint of the form /{index}/{type}, which is what this seems to be based on the .py source code. However, the API reference for Elasticsearch version 6.x and older shows that this was the correct endpoint to insert a document into an index. The reference page for 8.x (current) Elasticsearch lists a new endpoint along with the link [removal of mapping types](https://www.elastic.co/guide/en/elasticsearch/reference/8.14/removal-of-types.html), which explains that mapping types were slated to be removed in 5.x and have been phased out until their full removal in 8.x. However, pcp2elasticsearch seems to still rely on them. 

Sending a cURL POST request to the current endpoint works as expected: 
```
$ curl -i -X POST <es-server>/pcp/_doc
    
  HTTP/1.1 201 Created
  Location: /pcp/_doc/Q21s9I8BuerNNLwI2fY5
  X-elastic-product: Elasticsearch
  content-type: application/json
  content-length: 158

  {"_index":"pcp","_id":"Q21s9I8BuerNNLwI2fY5","_version":1,"result":"created","_shards":{"total":2,"successful":2,"failed":0},"_seq_no":4329,"_primary_term":1}
```
Changing the pcp2elasticsearch source code to use this updated endpoint also seems to fix the program, but the logic of it may rely on mapping types more deeply - I didn't spend very much time looking at it.

I am aware that there are more recent versions of PCP than 5.2.5 available which have updates to pcp2elasticsearch; however, these updates fix other problems. The current source code at pcp/src/pcp2elasticsearch/pcp2elasticsearch.py still uses the removed endpoint.

PR #2117, opened 2025-01-10
pcp2elasticsearch: support dynamic mapping

pauljevans:
Elasticsearch 7.X removed the concept of mapping types (initially this was depreciated in Elasticsearch 6.X).

Change over to support dynamic mapping by making use of the updated 'POST {server}/{index}/_doc' index API with auto-generated ids by default.

Users can still use the the search type '-p pcp-metric' command line argument on launch to make use of the original pcp-metric mapping name for environments with Elasticsearch 5.X and 6.X, ('POST {server}/{index}/{search_type}' API).

Note: These older versions are considered EOL by Elastic.

This fix addresses github issue #1998.

### DIFF
--- a/qa/1130
+++ b/qa/1130
@@ -2,7 +2,7 @@
 # PCP QA Test No. 1130
 # checks basic pcp2elasticsearch functionality
 #
-# Copyright (c) 2017-2019 Red Hat.
+# Copyright (c) 2017-2025 Red Hat.
 #
 seq=`basename $0`
 echo "QA output created by $seq"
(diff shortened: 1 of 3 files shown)


myllynen:
Thanks! I haven't followed Elasticsearch recently but based on your explanation this certainly makes sense.

However, looking at the CI runs were some of the related QA test cases skipped? Have you checked locally is there any QA impact with this change?

pauljevans:
Hi Marko,

Apologies for missing the QA test update, have provided in a further commit. 

I was originally running into an issue with QA/1130 skipping, i'm assuming that CI was complaining about also?  

It seems that 'socat' was failing on connection resets during the runs (giving "socat on this platform is not behaving as expected"). This behaviour was also being triggered for me with the current upstream version of the exporter also, I have commented out the 'socat' connection reset check in the test for now.

Cheers

myllynen:
Thank you! LGTM now but given that Ken spent a lot of time getting the QA running across different platform let's check with him if there would be any additional feedback - @kmcdonell ? Otherwise I think we could merge this. Thanks.

kmcdonell:
I don't have anything to add on the code change, this is Python after all.
To check for QA fallout, run ./check -g pmda.elasticsearch from the qa directory once the PMDA change is installed.

PR #2124, opened 2025-01-21
Remove all references to the lockdown class from SELinux policy

zpytela:
A comprehensive fix for all the problems caused by the lockdown SELinux class was rejected by Linus and for the lack of a better option, the consensus upstream was to just remove the class entirely and stop checking anything in the lockdown hook.

This commit is a follow-up to selinux-policy commit 12f821c731b2 ("Remove the lockdown-class rules from the policy").

### DIFF
--- a/src/selinux/pcp.te
+++ b/src/selinux/pcp.te
@@ -521,15 +521,6 @@ optional_policy(`
 
 optional_policy(`
     require {
-	class lockdown { integrity };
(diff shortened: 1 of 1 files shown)


zpytela:
This change should apply to RHEL 10 and Fedora 40+.
RHEL 9 and older did not have the lockdown class.

Issue #2089, opened 2024-10-14
[RFE] username authentication for pmdaproc


natoscott:
The proc metrics currently require UID-based authentication in order to allow fetch access - this is achieved using Unix-domain socket auto-authentication for most PCP tools, locally.  However users also want to access proc metric values using the REST API, and this provides USERNAME based authentication attribute to pmdaproc (which it ignores currently).

A scheme similar to pmdabpftrace - where an allowed_list of usernames can be specified - would be one way to solve this (stitching this into the logic behind the -A option and have_access.

An alternative (or additional) scheme could see pmdaproc attempt to associate the USERNAME attribute with a local UID, and if successful allow access in the same way the Unix-domain socket authentication functions.

PR #2135, opened 2025-02-18
pmcd, pmdaproc: allow non-local username connection attributes

natoscott:
Allows authenticated users to access the proc metric values even when there is no local account for the users name, via the /etc/pcp/proc/access.conf configuration file.

Related issue #2089

### DIFF
--- a/qa/1770
+++ b/qa/1770
@@ -142,10 +142,11 @@ $sudo chown pcp:pcp $PCP_SASLCONF_DIR/pmcd.conf
 ls -l $PCP_SASLCONF_DIR/pmcd.conf >>$seq_full
 $sudo -u pcp cat $PCP_SASLCONF_DIR/pmcd.conf >>$seq_full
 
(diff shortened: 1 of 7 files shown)


kmcdonell:
@natoscott by accident I was trying to get remote authentication working yesterday, and failed miserably. I found https://pcp.readthedocs.io/en/latest/QG/AuthenticatedConnections.html pretty confusing.  Would this be a good time to consider an editorial audit of that document, especially since this commit seems to move the ground underneath somewhat.

natoscott:
@kenj sounds like a good idea - I was interested from the REST API perspective from this change, and found the combination of reading QA tests and Andreas' write up from Grafana - https://grafana-pcp.readthedocs.io/en/latest/datasources/authentication.html - to be pretty close to the mark.  It's definitely not straightforward for folks coming across this functionality for the first time though, and there are a number of different scenarios for docs to consider.

PR #2156, opened 2025-03-14
Add filter parameter for the pmwebapi openmetrics 

wcohen:
The names parameter for pmwebapi get metrics operation pulls in all the names in the performance metric tree that match its comma separate list.  There may be cases where particular individual metrics that are found during the search should be excluded from the results.  For example one may want to exclude specific metrics that have instances for all the process running on the system, but include all the other related metrics in that area of the performance metric name tree.  Rather than enumerating the names in the names parameter one could just name the node in the tree that contains all the names and then use the filter parameter to exclude the unwanted metrics.

### DIFF
--- a/qa/1543.out
+++ b/qa/1543.out
@@ -1031,6 +1031,23 @@ sample_control{role="testing",agent="sample",hostname="HOSTNAME",cluster="zero",
 done full scrape
 == check pmproxy is running ==
 pmproxy check passed
+== good filter globbing ==
+# PCP5 sample.long.one 29.0.10 32 PM_INDOM_NULL instant none
+# HELP sample_long_one 1 as a 32-bit integer
+# TYPE sample_long_one gauge
+sample_long_one{role="testing",agent="sample",hostname="HOSTNAME",cluster="zero",domainname="DOMAINNAME",machineid="MACHINEID"} 1
+== good filter exact match ==
+# PCP5 sample.long.one 29.0.10 32 PM_INDOM_NULL instant none
+# HELP sample_long_one 1 as a 32-bit integer
+# TYPE sample_long_one gauge
+sample_long_one{role="testing",agent="sample",hostname="HOSTNAME",cluster="zero",domainname="DOMAINNAME",machineid="MACHINEID"} 1
+== good filter regex ==
+# PCP5 sample.long.one 29.0.10 32 PM_INDOM_NULL instant none
+# HELP sample_long_one 1 as a 32-bit integer
+# TYPE sample_long_one gauge
+sample_long_one{role="testing",agent="sample",hostname="HOSTNAME",cluster="zero",domainname="DOMAINNAME",machineid="MACHINEID"} 1
+== bad filter match param ==
+{"context":CONTEXT,"message":"dummy - invalid 'match' parameter value","success":false}
 
 *****   GET and POST /pmapi/derive   *****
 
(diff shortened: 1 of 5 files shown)


qa/1543.out
```diff
@@ -1031,6 +1031,23 @@ sample_control{role="testing",agent="sample",hostname="HOSTNAME",cluster="zero",
...
+# HELP sample_long_one 1 as a 32-bit integer
+# TYPE sample_long_one gauge
+sample_long_one{role="testing",agent="sample",hostname="HOSTNAME",cluster="zero",domainname="DOMAINNAME",machineid="MACHINEID"} 1
+== good filter regex ==
+# PCP5 sample.long.one 29.0.10 32 PM_INDOM_NULL instant none
+# HELP sample_long_one 1 as a 32-bit integer
+# TYPE sample_long_one gauge
+sample_long_one{role="testing",agent="sample",hostname="HOSTNAME",cluster="zero",domainname="DOMAINNAME",machineid="MACHINEID"} 1
+== bad filter match param ==
+{"context":562009464,"message":"dummy - invalid 'match' parameter value","success":false}
```
natoscott, line 1050: We'll need to add a sed filter on the context number here @wcohen (in test script qa/1543).

src/libpcp_web/src/webgroup.c
```diff
@@ -1867,6 +1868,39 @@ webgroup_scrape(pmWebGroupSettings *settings, context_t *cp,
...
+    enum matches	match;		/* type of matching to use for filters */
+    regex_t		*regex;		/* for filter compiled regular expressions */
+    void		*arg;
+} webscrape_t;
+
+static int
+filtered(struct webscrape *filter_info, const char *name)
+{
+    int i;
+    for (i = 0; i<filter_info->numfilters; ++i) {
```
natoscott, line 1891: Code style in this file IIRC is to have whitespace between 'i' and '<' and 'filter_info' here, and we tend to be the "i++" rather than "++i" style of for loop (no difference functionally, but that's the general coding style).

src/libpcp_web/src/webgroup.c
```diff
@@ -1867,6 +1868,39 @@ webgroup_scrape(pmWebGroupSettings *settings, context_t *cp,
...
+	if (filter_info->match == MATCH_EXACT && strcmp(filter_info->filters[i], name) == 0)
+	    return 1;
+	if (filter_info->match == MATCH_REGEX &&
+	    regexec(&(filter_info->regex[i]), name, 0, NULL, 0) == 0)
+	    return 1;
+	if (filter_info->match == MATCH_GLOB &&
+	    fnmatch(filter_info->filters[i], name, 0) == 0)
+	    return 1;
```
natoscott, line 1899: Small code consistency thing - can we make the first "if" branch line-splitting match the other two cases here?  (i.e. insert EOL after "&&").

src/libpcp_web/src/webgroup.c
```diff
@@ -1911,8 +1933,10 @@ webgroup_scrape_batch(const char *name, void *arg)
 	scrape->numnames = 0;
     }
 
-    scrape->names[scrape->numnames] = sdsnew(name);
-    scrape->numnames++;
+    if(!filtered(scrape, name)) {
```
natoscott, line 1936: Missing whitespace after "if" here.

natoscott (changes requested):
Looks good Will - couple of minor code style consistency things, and a tweak to the QA test is all I can suggest to improve this - functionally it is spot on already.  Thanks!

src/libpcp_web/src/webgroup.c
wcohen: I rebased on the current upstream master branch and updated the patches to address the comments above.  It should be ready to merge now.

PR #2157, opened 2025-03-14
qa/1747: fix of a typo in a variable name

kurik:
There is a typo in the qa/1747 test, causing this test to always fail.
$seq_full -> $seq.full


natoscott:
This is the correct variable name in the 'main' branch now - I'm guessing I've back-ported something, somewhere and botched it @kurik - let's chat further on slack

PR #2174, opened 2025-04-01
containers: update grafana time window in archive-analysis

sfeifer:
- Adds to the archive-import.py script to modify the grafana dashboard json based on the time window collected from the archives being imported
- Renames the Dockerfile to Containerfile to be more podman friendly


build/containers/archive-analysis/Containerfile
```diff
@@ -24,7 +24,7 @@ COPY --from=build /build /build
 COPY build/containers/archive-analysis/root /
 
 RUN dnf install -y --setopt=install_weak_deps=False --setopt=tsflags=nodocs \
-        python3 procps-ng gettext /build/*.rpm valkey grafana grafana-pcp && \
+        python3 python3-pcp procps-ng gettext /build/*.rpm valkey grafana grafana-pcp && \
```
natoscott, line 27: Just noticed "/build/*.rpm" here - I think this Containerfile should to be updated to not build its own PCP nowadays (see the RUN commands on line 10 and 11 - turf these) but rather use PCP from the distribution.
This build-it-locally I think harks back to the times we were still building pmproxy, pmseries, et al, and is no longer needed.

build/containers/archive-analysis/root/usr/bin/archive-import.py
```diff
@@ -43,6 +43,33 @@ def format_time(seconds: float):
     string = datetime.fromtimestamp(seconds, UTC).isoformat()
     return string.replace('+00:00', 'Z')
 
+def update_dashboard_time_window(begin, end):
```
natoscott, line 46: Rest of this file has python function types - can you add those in here?  (str in both cases)

build/containers/archive-analysis/root/usr/bin/archive-import.py
```diff
@@ -43,6 +43,33 @@ def format_time(seconds: float):
...
+    json_path="$(podman mount pcp-archive-analysis)/usr/local/var/lib/grafana/dashboards/pcp-archive-analysis.json"
+
+    # Check if the JSON file exists
+    if [ ! -f "$json_path" ]; then
+        echo "Error: JSON file not found" >&2
+        exit 1
+    fi
+
+    # Modify JSON using jq
+    jq '.time.from = "{begin}" | .time.to = "{end}"' "$json_path" > "$json_path.tmp" && mv "$json_path.tmp" "$json_path"
```
natoscott, line 62: Oh, hmm - could we use the python json library instead here?

build/containers/archive-analysis/root/usr/bin/archive-import.py
```diff
@@ -43,6 +43,33 @@ def format_time(seconds: float):
...
+
+    # script that gets the path to the json and updates the to and from fields
+    script = f"""
+    json_path="$(podman mount pcp-archive-analysis)/usr/local/var/lib/grafana/dashboards/pcp-archive-analysis.json"
+
+    # Check if the JSON file exists
+    if [ ! -f "$json_path" ]; then
+        echo "Error: JSON file not found" >&2
+        exit 1
+    fi
```
natoscott, line 59: Wait what?  isn't this a python script ... ? (not a shell script)

natoscott (changes requested):
I think there's some mixup between python/shell code in here to sort out, and as a followup (esp. since adding python3-pcp from outside the build) we should switch over to using distro packages IMO.  WDYT?

sfeifer:
No confusion :)....but it is a bit of a workaround. I initially tried using the python json library, but this did not work as it could not find the file to load and edit. I did some further investigation and found the json file was only accessible within a shell opened by `podman unshare`. I then updated the script to open a shell with `podman unshare`, get the json path, modify the from and to fields in the json file, and then exit the podman unshare shell automatically once the json file is done being updated. I'm open to suggestions on a better way of doing this, but this is what I found worked on my machine.

kurik:
Those changes to `archive-import.py` looks fine to me (even I have not tested it).
However I am a bit more concerned about those changes to the Dockerfile/Containerfile. I am not saying these changes are bad, but I just want to highlight some consequences:

- renaming  `Dockerfile` to `Containerfile` will break the current build setup on https://quay.io/repository/performancecopilot/archive-analysis - not a problem to reconfigure it, we just should not forget about it.
- rebuild of upstream PCP replaced with installation of PCP packages - there are two aspects affected by this:
  - build of the `archive-analysis` image is triggered by every commit to `main` branch of [PCP repository on github](https://github.com/performancecopilot/pcp). If we use build from PCP Fedora packages instead of rebuild from the latest upstream code, the triggering mechanism should be reworked. Perhaps the rebuild should be triggered by finished Koji build in Fedora or a Bodhi update. I am not sure, but I guess, to trigger a build in quay.io from Koji or Bodhi, we might need to integrate Packit to Fedora pipeline ?
  - Currently we already have PCP-7 in the  `archive-analysis` image, however when we install from Fedora packages, we will get PCP-6. This might be a benefit (stability, reliability) but can also be considered as a disadvantage (not an upstream version).
- two-stage build reduced to one-stage build - I am a bit concerned about size of the final image, as the two-stage build was there to reduce amount of ballast in the final image. But I have not tested this, so we might be still fine even with the one-stage build. This is something we should IMO explore a bit.

sfeifer:
I tried removing those two lines from the Containerfile, but both are still needed.

- chown: When I removed this line, I could not access localhost:3000
- sed: I thought I would be able to remove this line as the grafana.ini file in the most recent release is patched to allow both pcp-valkey-datasource and performancecopilot-valkey-datasource. When I tested with this change, though, the datasource and dashboard were not working in the container. I'm not sure exactly why this line is still needed as of now, other than needing it to get the dashboard to work.

Issue #2175, opened 2025-04-03
Discrete values should not always be reported for every fetch


myllynen:
PCP command line tools such as pmrep(1) using pmfg report discrete metric values for every fetch even when using the uninterpolated mode when there are new values for other metrics part of the fetchgroup:

```
$ pmlogdump -z -m sysbenchTEST | grep -c hinv.physmem              
1
$ pmrep -z -a sysbenchTEST -s 3 -p -b GB -u hinv.physmem
          h.physmem
                 GB
11:00:13    502.896
$ pmrep -z -a sysbenchTEST -s 3 -p -b GB -u hinv.physmem mem.util.free
          h.physmem  m.u.free
                 GB        GB
11:00:13    502.896       N/A
11:00:13    502.896   485.472
11:00:18    502.896   485.470
```

While this has not been and should not be an issue when reporting interactively on the command line this can be an issue e.g. with pcp2elasticsearch(1) as a potentially huge number of repetitive, non-existing metric values are being sent to an external system. This behavior can also be observed with pcp2xml(1) which reports the same as pcp2elasticsearch(1) would send out:

```
$ pcp2xml -z -a sysbenchTEST -s 3 hinv.physmem mem.util.free | grep -c physmem
3
```

It would be ideal to make this repeated reporting of discrete values optional, it would allow preserving and opting for the current behaviour in interactive use but when sending metric data to external systems only the existing and changed values would be sent over.

Looking under the hood, it turns out that in libpcp/pmfg there is special handling for discrete metrics that implements the current behaviour. IOW, pmfg always reports once seen discrete metrics but in the rare case new values for them would appear then those would be reported from there one. It seems uncontroversial that a) discrete metrics are being reported and b) in case they happen to change, for whatever reason, perhaps due to CPU/memory hot-plug or so, the new values are taken into account. The only question then seems to be about the above described behaviour, whether to report discrete metric values only once or "always" even if there were no values for them available as part of the latest fetch.

Since the pmfg code already both stores the previous results and also has special handling for discrete values it would seem more natural to consider changing this if so warranted in the libpcp/pmfg code as it is impossible for clients to know whether the discrete metric value was from a previous fetch or a newly available value. A client could of course store previous results and thus duplicate the effort with pmfg and check if the current are previous values are different which would probably lead to the right conclusion most of the time but any time a new unchanged discrete metric value is present in an archive there would be no way for a client to determine this reliably.

[...]

PR #2177, opened 2025-04-03
pmfg: allow for not preserving discrete metric values

myllynen:
Existing QA and C (pmclient_fg) and Python (pmrep et al) clients all work as before, the next commit will make the Python clients to use the feature and add appropriate QA.

Related to issue #2175

### DIFF
--- a/man/man3/pmfetchgroup.3
+++ b/man/man3/pmfetchgroup.3
@@ -481,6 +481,9 @@ pmValueSet.numval==0), then the last seen valid value (if any) is
 retained.
 This is intended to ease the processing of sets of archives with a
 mixture of once- and repeatedly-sampled metrics.
+The environment variable
+.B PCP_PMFG_NO_PRESERVE
+may be set to disable this feature.
 .SS Clearing a fetchgroup
 .ft 3
 .nf
--- a/src/libpcp/src/fetchgroup.c
+++ b/src/libpcp/src/fetchgroup.c
@@ -42,6 +42,7 @@ struct __pmInDomCache {
 struct __pmFetchGroup {
     int	ctx;			/* our pcp context */
     int wrap;			/* wrap-handling flag, set at fg-create-time */
+    int no_preserve;		/* do not preserve discrete values, set at fg-create-time */
     pmHighResResult *prevResult;
     struct __pmFetchGroupItem *items;
     struct __pmInDomCache *unique_indoms;
@@ -821,7 +822,7 @@ pmfg_fetch_item(pmFG pmfg, pmFGI item, pmHighResResult *newResult)
      * If we have some values, then DISCRETE preserved values should
      * be cleared now.
      */
-    if (item->u.item.metric_desc.sem == PM_SEM_DISCRETE) {
+    if (item->u.item.metric_desc.sem == PM_SEM_DISCRETE && !pmfg->no_preserve) {
 	for (i = 0; i < newResult->numpmid; i++) {
 	    if (newResult->vset[i]->pmid == item->u.item.metric_pmid) {
 		if (newResult->vset[i]->numval > 0) {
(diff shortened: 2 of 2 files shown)


src/include/pcp/pmapi.h
```diff
@@ -1199,6 +1199,7 @@ PCP_CALL extern int pmExtendFetchGroup_event(pmFG, const char *, const char *,
 			unsigned int, unsigned int *, int *);
 PCP_CALL extern int pmExtendFetchGroup_timestamp(pmFG, struct timeval *);
 PCP_CALL extern int pmFetchGroup(pmFG);
+PCP_CALL extern int pmFetchGroup_p(pmFG, int);
```
natoscott, line 1202: @myllynen I think it will be cleaner to not extend the API for special cases like this (else we end up with a bunch of APIs when next special case arrives, then the next, ...), but rather to use the same mechanism that special fetch handling for counter wrap is implemented using.

Have a look at the 'wrap' field in struct __pmFetchGroup - we could so similar here: add a 'preserve' field there, check an env var (once) on context creation, use the new boolean field internally, and not need an additional API.

natoscott (changes requested):
I think can be achieved more simply and more consistently with the 'wrap' handling in fetchgroup.c - see comment with pmapi.h update inline here.

myllynen:
Fair enough, this also simplifies the change itself, a new proposal pushed. Although I'm not a fan of the double-negation ("!pmfg->no_preserve") there I don't think we can or want to change the library default and using the same word in the variable name makes is perhaps slightly better than to come up with another name which wouldn't show up when searching for "preserve".

Do we need to add the env var description also the PCPIntro(1) or elsewhere?

Thanks.

natoscott:
> Fair enough, this also simplifies the change itself, a new proposal pushed. Although I'm not a fan of the double-negation ("!pmfg->no_preserve") there I don't think we can or want to change the library default and using the same word in the variable name makes is perhaps slightly better than to come up with another name which wouldn't show up when searching for "preserve".

OK.

> Do we need to add the env var description also the PCPIntro(1) or elsewhere?
> 
> Thanks.

Since it only affects fetchgroup.c (which affects relatively few PMAPI clients - mainly python tools) I think best location would be pmFetchGroup(3).

myllynen:
Yes, there's a note added in the pmFetchGroup(3) already. Let me know if this part looks acceptable and ready, I could then proceed updating the relevant Python clients, I don't see e.g. pmclient_fg needing to be updated due to this. Thanks.

natoscott:
> Fair enough, this also simplifies the change itself, a new proposal pushed. Although I'm not a fan of the double-negation ("!pmfg->no_preserve") there I don't think we can or want to change the library default and using the same word in the variable name makes is perhaps slightly better than to come up with another name which wouldn't show up when searching for "preserve".
> 

Reviewing the code again, I think we can improve it further.  The double negation problem can be solved by explicitly setting the flag to either one or zero in the constructor.  The variable name I'm finding a bit too cryptic as well ... how about this?

```
src/libpcp/src/fetchgroup.c
--- /tmp/git-blob-n9fASm/fetchgroup.c   2025-04-09 09:16:19.954846224 +1000
+++ src/libpcp/src/fetchgroup.c 2025-04-09 09:15:24.017766940 +1000
@@ -1291,6 +1291,10 @@ pmCreateFetchGroup(pmFG *ptr, int type,
     PM_LOCK(__pmLock_extcall);
     if (getenv("PCP_COUNTER_WRAP") != NULL)
        pmfg->wrap = 1;
+    if (getenv("PCP_DISCRETE_ONCE") != NULL)
+       pmfg->preserve = 0;
+    else
+       pmfg->preserve = 1;
     PM_UNLOCK(__pmLock_extcall);
 
     pmfg_clear_profile(pmfg);
```

Additionally, these fields (both 'wrap' and 'preserve') are boolean, so we can define them more like that and make the true/false-only semantics explicit via:

```
struct __pmFetchGroup {
    int ctx;                    /* the PMAPI context */
    unsigned int wrap : 1;      /* counter wrap handling */
    unsigned int preserve : 1;  /* discrete value preserving */
```

PR #2189, opened 2025-04-14
pmrep: re-enable pylint CI checking, fix reported issues

natoscott:


natoscott:
@myllynen one issue remains - any preference on how to best tackle this one:

```
pmrep.py:757:20: R1728: Consider using a generator instead 'sum(len(x[0]) for x in self.pmconfig.insts)' (consider-using-generator)
pmrep.py:813:28: R1728: Consider using a generator instead 'sum(len(i[0]) for i in self.pmconfig.insts)' (consider-using-generator)
pmrep.py:818:28: R1728: Consider using a generator instead 'sum(len(results[i]) for i in results)' (consider-using-generator)
```

myllynen:
Thanks, the suggestions indeed make sense here. So in this case pylint is  helpful enough to show how the more efficient code would look like. I already tried it and it passes all QA. Basically what's needed is to remove the []s on those three lines. If you want me to do it, just merge this and I'll tweak it right after that. Thanks.

PR #2191, opened 2025-04-16
ci: switch centos7-container over to a more recent Ubuntu

natoscott:
CI has started failing today with this Github runner message: "Ubuntu 20.04 LTS runner will be removed on 2025-04-15."

So we have no choice but to move now and revisit whatever the earlier reason for not moving forward for centos7 was.


natoscott:
Removed from CI elsewhere so this is no longer needed.

PR #2192, opened 2025-04-22
Fixed qa test 1988 for pcp-meminfo

orasagar:


natoscott:
This test is no longer failing in CI, nor when run locally for me with latest 'main' branch contents:

```
pcpqa@fedora:~$ ./check 1988
1988
Passed 1 tests
pcpqa@fedora:~$ 
```

So it seems this change is not needed?

orasagar:
I also do see latest main branch getting passed this QA test.
I think we can skip this.

Issue #2017, opened 2024-07-12
[RFE] pcp dstat utility configurable filtering


natoscott:
Issue #2016 describes a problem where root cause is the way pcp-dstat.py has hard-coded mappings between instance filtering command line options and the specific plugins they apply to.

A cleaner solution would allow the plugin configuration files to specify this mapping.  That way, as new config files are added (in PCP or by the user locally), the python code does not need to be changed to add the plugin names.

For example, the disk-util plugin from /etc/pcp/dstat/disk might gain a filter mapping specification something like:

```
[disk-util]
label = %I
width = 4
**filtertype = disk**
grouptype = 1
printtype = f
colorstep = 34
util = 100 * rate(disk.dev.avactive)
```

These are the all of the relevant command line options that provide "filtertype" values:

```
  -c, --cpu             enable cpu stats
     -C 0,3,total          include cpu0, cpu3 and total
  -d, --disk            enable disk stats
     -D total,sda          include sda and total
  --dm, --device-mapper enable device mapper stats
     -L root,home,total    include root, home and total
  --md, --multi-device  enable multi-device driver stats
     -M total,md-0         include md-0 and total
  --part, --partition   enable disk partition stats
     -P total,sdb2         include sdb2 and total
  -i, --int             enable interrupt stats
     -I 9,CAL              include int9 and function call interrupts
  -n, --net             enable network stats
     -N eth1,total         include eth1 and total
  -s, --swap            enable swap stats
     -S swap1,total        include swap1 and total
```

wcohen:
This has been resolved by the following git commits:
8d328da68ffd640428938e2fd994d6a42e98ecd1
9be342d78f7c0d7a4c77ecd013549dbc907b564f
425fab84249ca3f53b4220ef6b823960a670266c
406ab4b30473653d30ad3ffc92abdb10600ca799

Issue #2095, opened 2024-10-21
[RFE] add GPU support to dstat


natoscott:
Starting to get requests for this, not too surprisingly.  :)

It would be great to make GPUs a first class entity in dstat, just like CPUs, disks, network interfaces and so on, i.e. such that values for each individual card can be reports, filtered, etc.

Here's an example simple plugin that was not merged in the old dstat that could help as a starting point for nvidia GPUs (doesn't support per-card stats though): 
https://blog.datumbox.com/getting-the-gpu-usage-of-nvidia-cards-with-the-linux-dstat-tool/

We also need to support the AMD GPU metrics in PCP already, and ease the way for other GPU vendors too.

christianhorn:
Would also be cool to have these stats not only on GPU clocking or business, but consumption.. unfortunately, even if GPUs offer the metrics, it seems not nicely standardized so far, i.e. via standardized sysfs points.

wcohen:
Have a similar plugin for AMD GPU.

```
$ pcp dstat --amd-gpu
--gputotal-
util%  mem%
   24    34
   36    34 missed 2 ticks
   17    34
   11    34
    8    34
    9    34
```
It should be giving entries for two GPUs on this machine rather than one as there is a integrated GPU and a discrete GPU on it.

```
wcohen@e-bike:~$ pmval  amdgpu.numcards

metric:    amdgpu.numcards
host:      e-bike
semantics: discrete instantaneous value
units:     none
samples:   all
          2
^C
wcohen@e-bike:~$ pmval  amdgpu.cardname

metric:    amdgpu.cardname
host:      e-bike
semantics: discrete instantaneous value
units:     none
samples:   all

                 gpu0                  gpu1 
    "AMD Radeon 780M" "AMD Radeon RX 7700S" 

```

Issue #2016, opened 2024-07-10
Disk device filter does not work under disk-util and disk-wait.


NyBT-spec:
I installed pcp-dstat (5.3.7) from redhat 8 repo
The device filter does not work under --disk-util and --disk-wait.

working example:
root# dstat --disk-tps 1 1 -D scinia
-dsk/scinia
#read #writ
   0     0

non-working sample:

root# dstat --disk-util 1 1 -D scinia
sda--scin-scin-scin-scin-scin-scin-scin-scin-scin-scin-scin-scin-scin-scin-scin-scin-scin-scin
util:util:util:util:util:util:util:util:util:util:util:util:util:util:util:util:util:util:util
0.10:   0:   0:   0:   0:   0:   0:   0:   0:   0:   0:   0:   0:   0:   0:   0:   0:   0:   0

root# dstat --disk-wait 1 1 -D scinia
---sda------scinia----scinib----scinic----scinid----scinie----scinif----scinig----scinih----scinii----scinij----scinik----scinil----scinim----scinin----scinio----scinip----sciniq----scinir-
rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa:rawa wawa
   0 0.00:   0    0:   0    0:   0    0:   0    0:   0    0:   0    0:   0    0:   0    0:   0    0:   0    0:   0    0:   0    0:   0    0:   0    0:   0    0:   0    0:   0    0:   0    0

Is there any way I can fix this or do I have to wait for a fix?

natoscott:
@NyBT-spec thanks for reporting - this is a newly discovered issue, and I see the same behaviour in the latest version of the code too:

```
$ pcp --archive qa/archives/bozo-disk --start @10:32 dstat --time --disk-util 10 3 -D sda,sdb
----system---- sda--sdb--sdc--sdd--sde-
     time     |util:util:util:util:util
16-03 10:32:00|    :    :    :    :    
16-03 10:32:10|1.22:0.86:0.54:94.9:0.22
16-03 10:32:20|0.16:0.55:0.46:92.6:0.12
16-03 10:32:30|0.05:0.65:0.08:91.8:0.59
```

I'll take a closer look when I can, but to answer your questions: at this stage there is no fix available.

[...]

PR #2196, opened 2025-04-29
Removed hard-coded mappings in pcp dstat; resolves issue #2017

lmchilton:
Resolves git issue #2017

Updated pcp dstat tool to not rely on hard coded mappings between instance filtering command line options and the specific plugin they apply to

### DIFF
--- a/src/pcp/dstat/pcp-dstat.py
+++ b/src/pcp/dstat/pcp-dstat.py
@@ -444,7 +444,7 @@ class DstatTool(object):
 
         ### Add additional dstat metric specifiers
         dspec = (None, 'printtype', 'colorstep', 'grouptype', 'cullinsts',
-                'plugin', 'valuesets')
+                'plugin', 'valuesets', 'filter')
         mspec = self.pmconfig.metricspec + dspec
         self.pmconfig.metricspec = mspec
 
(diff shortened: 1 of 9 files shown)


wcohen:
Read #2017 and #2016 to get an understanding what the patch is addressing.  The patch looks sane.  It would be really good to have a test in qa directory to verify that the some filtering is working.  It looks like qa/1187 would be a likely place to include such a test.  Ideally, would check that #2016 is addressed, but that would require some additional archived data.  To keep it simple maybe filter existing archive data for cpu usage.

One thing I noticed is that the filtertype ended up being the plugin file name. Are there going to be any cases where the filtertype is going to be different that the plugin name?

natoscott:
| One thing I noticed is that the filtertype ended up being the plugin file name.
| Are there going to be any cases where the filtertype is going to be different that the plugin name?

I think this is a safe assumption to make at this stage.  Only future plugin/filter I can think of is for GPUs, but I expect they'll each require a separate file (a bit like disk vs dm vs md I guess).

wcohen:
For testing it would be good to build an RPM with this patch installed, install the generated RPMs, and test out on the dstat test group with:

```
cd /var/lib/pcp/testsuite/
sudo ./check -g dstat
```
Looking through the patch it seems odd that the plugins have "filtertype" but the pcp-dstat.py is looking for "filter".  Is that correct?  When trying to try this out for future extentions  to support dstat monitoring of GPUs (#2095) changed pcp-dstat.py to look for "filtertype", but got:

```
$ dstat -c
  File "/usr/bin/dstat", line 614, in prepare_plugins
    found = config.read(paths)
  File "/usr/lib64/python3.13/configparser.py", line 735, in read
    self._read(fp, filename)
    ~~~~~~~~~~^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/configparser.py", line 1050, in _read
    ParsingError._raise_all(self._read_inner(fp, fpname))
                            ~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/usr/lib64/python3.13/configparser.py", line 1079, in _read_inner
    self._handle_rest(st, line, fpname)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/configparser.py", line 1103, in _handle_rest
    raise MissingSectionHeaderError(fpname, st.lineno, line)
configparser.MissingSectionHeaderError: File contains no section headers.
file: '/etc/pcp/dstat/GNUmakefile', line: 15
'TOPDIR = ../../../..\n'

During handling of the above exception, another exception occurred:

  File "/usr/bin/dstat", line 1933, in <module>
    dstat = DstatTool(sys.argv[1:])
  File "/usr/bin/dstat", line 537, in __init__
    configs = self.prepare_plugins()
  File "/usr/bin/dstat", line 616, in prepare_plugins
    sys.stderr.write("Config parse failure: %s\n" % cfgerr.message())
                                                    ~~~~~~~~~~~~~~^^
TypeError: 'str' object is not callable

```

natoscott:
| sudo ./check -g dstat

Just for info, the testsuite installs a 'pcpqa' user and some tests do not correctly as root - I tend to "su - pcpqa" (which also conveniently lands you in /var/lib/pcp/testsuite) and from there it's just "./check ..."

lmchilton:
See new PR #2218

PR #2203, opened 2025-05-08
pmseries_import: use a datetime UTC interface from python3.6

natoscott:
Resolves an issue Ken reported running tests in the QA farm, which is due to use of an interface from more-recent-python: https://github.com/python/cpython/pull/91973


kmcdonell:
@natoscott LGTM.  I've installed your latest version of archive-import.py and run qa/1807 successfully on each of the following:
- bozo Ubuntu 24.04 Python 3.12.3
- vm04 CentOS Stream8 Python 3.6.8
- vm07 Debian 11.11 Python 3.9.2
- vm20 Ubuntu 20.04 Python 3.8.10

PR #2215, opened 2025-06-02
build: add configure option to enable/disable multilib headers

natoscott:
We want a very restricted set of places where the 32/64 suffix headers are used: only on RPM platforms opting-in for multilib.

Related to RHEL-93186


kmcdonell:
LGTM.  I've pulled the commit into my local tree and rebuilt and rerun QA for -g python -g sanity on 32-bit and 64-bit Ubuntu systems.

But I have a question ... I think the PM_PAD_* things in config32.h are wrong, as they should be #undef rather than #define ... 4

The attached C program (called x.c.txt because github is lame and won't upload a .c file) is based on the fragment of configure.ac that sets the "pad" thingies, augmented with lots more printf's.

[x.c.txt](https://github.com/user-attachments/files/20559097/x.c.txt)

I've installed the multilib toolchain rpms on a F42 system and these are the results.

For 64-bit
```
kenj@vm03:~$ cc x.c
kenj@vm03:~$ a.out
sizeof __pmTimestamp sec 8 nsec 4 total 16
sizeof __pmResult 24 numpmid offset 16
sizeof struct timespec tv_sec 8 tv_nsec 8 total 16
sizeof pmResult 24 numpmid offset 16
sizeof struct timeval tv_sec 8 tv_usec 8 total 16
sizeof pmResult_v2 24 numpmid offset 16
NO padding for __pmResult
NO padding for timespec
```
so NO padding and config64.h matches

For 32-bit
```
kenj@vm03:~$ cc -m32 x.c
kenj@vm03:~$ file a.out
a.out: ELF 32-bit LSB executable, Intel i386, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux.so.2, BuildID[sha1]=129a5a6358e04e2a5b380ab7eb0e3fa751bbb381, for GNU/Linux 3.2.0, not stripped
kenj@vm03:~$ a.out
sizeof __pmTimestamp sec 8 nsec 4 total 12
sizeof __pmResult 16 numpmid offset 12
sizeof struct timespec tv_sec 4 tv_nsec 4 total 8
sizeof pmResult 12 numpmid offset 8
sizeof struct timeval tv_sec 4 tv_usec 4 total 8
sizeof pmResult_v2 12 numpmid offset 8
NO padding for __pmResult
NO padding for timespec
```
so NO padding again, which does NOT match config32.h

natoscott:
@kmcdonell thanks Ken, I'll update config32.h and push shortly.

PR #2219, opened 2025-06-06
Addition of pcp2opentelemetry tool

lmchilton:
Supports translation of pcp metrics into opentelemetry format. Supports opentelemetry protocol http json.

Updated relevent qa tests; added one new one (1977) to test pcp2opentelemetry http functionality.

Added man page and makefile. Includes updates and
changes from review.

### DIFF
--- a/qa/1131
+++ b/qa/1131
@@ -1,8 +1,8 @@
 #!/bin/sh
 # PCP QA Test No. 1131
-# Exercise pcp2json and pcp2openmetrics.
+# Exercise pcp2json, pcp2openmetrics, and pcp2opentelemetry.
 #
-# Copyright (c) 2017 Red Hat.
+# Copyright (c) 2025 Red Hat.
 #
 
 seq=`basename $0`
@@ -17,6 +17,8 @@ $python -c "from collections import OrderedDict" >/dev/null 2>&1
 
 which pcp2json >/dev/null 2>&1 || _notrun "pcp2json not installed"
 which pcp2openmetrics >/dev/null 2>&1 || _notrun "pcp2openmetrics not installed"
+which pcp2opentelemetry >/dev/null 2>&1 || _notrun "pcp2opentelemetry not installed"
+
 
 status=1       # failure is the default!
 signal=$PCP_BINADM_DIR/pmsignal
@@ -46,9 +48,24 @@ _filter_pcp2openmetrics()
     -e '/machineid=/s/'$machineid'/MACHINEID/' \
     -e '/groupid=/s/groupid="*[0-9][0-9]*"*/groupid="GROUPID"/' \
     -e '/userid=/s/userid="*[0-9][0-9]*"*/userid="USERID"/' \
+    -e '/agent=/s/agent="*[^"]*"*/agent="AGENT"/' \
     -e '/hostname=/s/'$hostname'/HOST/' \
     -e 's/\(.*}\).*/\1 NCPU/' \
     | LC_COLLATE=POSIX sort
+}
+
+_filter_pcp2opentelemetry()
+{
+   tee -a $here/$seq.full \
+   | col -b \
+   | sed \
+       -e '/\"asInt\":/ s/[0-9][0-9]*/NCPU/' \
+       -e 's/\"'$machineid'\"/MACHINEID/' \
+       -e 's/\"'$hostname'\"/HOSTNAME/' \
+       -e 's/\"'$domainid'\"/DOMAINID/' \
+       -e 's/'$UID'/USER\/GROUP ID/' \
+       -e '/"key": "agent"/,/"stringValue"/s/"stringValue": "[^"]*"/"stringValue": AGENT/' \
+       -e 's/\"timeUnixNano\": .*/\"timeUnixNano\": TIME/'
 }
 
 # real QA test starts here
@@ -64,6 +81,8 @@ echo "---"
 pcp2openmetrics -s1 -z hinv.ncpu | _filter_pcp2openmetrics
 echo "---"
 pcp2openmetrics -s2 -x hinv.ncpu | _filter_pcp2openmetrics
+echo "---"
+pcp2opentelemetry -s1 -H -z hinv.ncpu | _filter_pcp2opentelemetry
 echo "---"
 
 cat $tmp.outfile | _archive_filter
--- a/qa/1977.out
+++ b/qa/1977.out
@@ -1,0 +1,137 @@
+QA output created by 1977
+pcp2opentelemetry invocation
+pcp2opentelemetry HTTP POST (sorted):
+INFO:root:Starting httpd...
+
+INFO:root:POST request,
+Path: /receive
+Headers:
+Host: localhost:PORT
+User-Agent: python-requests VERSION
+Accept: */*
+Content-Type: application/json
+Content-Length: SIZE
+
+
+
+Body:
+{
+  "resourceMetrics": [
+    {
+      "resource": {
+	"attributes": [
+	  {
+	    "key": "os.type",
+	    "value": {
+	      "stringValue": "linux"
+	    }
+	  },
+	  {
+	    "key": "telemetry.sdk.language",
+	    "value": {
+	      "stringValue": "python"
+	    }
+	  },
+	  {
+	    "key": "telemetry.sdk.name",
+	    "value": {
+	      "stringValue": "opentelemetry"
+	    }
+	  },
+	  {
+	    "key": "telemetry.sdk.version",
+	    "value": {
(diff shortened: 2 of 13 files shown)


natoscott:
@lmchilton this is failing in the CI build because the new pcp2opentelemetry tool subdir is missing from the src/GNUmakefile - this patch should fix it:
```
$ git diff
src/GNUmakefile
--- /tmp/git-blob-wXFv0g/GNUmakefile    2025-06-12 10:07:23.480332266 +1000
+++ src/GNUmakefile     2025-06-12 09:55:30.834563600 +1000
@@ -65,6 +65,7 @@ OTHER_SUBDIRS += \
        pcp2influxdb \
        pcp2json \
        pcp2openmetrics \
+       pcp2opentelemetry \
        pcp2spark \
        pcp2xlsx \
        pcp2xml \

```

wcohen:
Need an entry in qa/groups for the new test 1977.  Without the entry the the 1977 test doesn't get installed.
When looking through the test  qa/1131.out wondering why the change adding "age:" {...} on the pcp2openmetrics?

wcohen:
The "age": .... stuff doesn't look like it should be there in qa/1131.out

`qa/group` needs entry for pcp2openmetrics.

Need to have a entry in `qa/group` for 1977, something like:

`1977 pcp2openmetrics python pcp2xxx local`

Should test 1977 also check that -s option with an option greater than
1 works and doesn't hang (like the recent issue with qa/1131,
PR#2243).

There be some substitutions in `qa/1977.out`:

- for the "server.address" to replace the 54322 in "http://localhost:54322/receive". PROXYPORT?
- "groupid" 4209557 should have generic string replacing it. GID?
- "userid" 4209557 should have a generic string replacing it. UID?


Looking over the pcp2opentelemetry.py file based on the pcp2openmetric

Issues with the `PM_DEBUG_APPL1` in pcp2opentelemetry.py:

```
$ pcp2opentelemetry
  File "/usr/sbin/pcp2opentelemetry", line 33, in <module>
    from cpmapi import PM_CONTEXT_ARCHIVE, PM_INDOM_NULL, PM_DEBUG_APPL1, PM_TIME_SEC
ImportError: cannot import name 'PM_DEBUG_APPL1' from 'cpmapi' (/usr/lib64/python3.13/site-packages/cpmapi.cpython-313-x86_64-linux-gnu.so)

```

The PCP QA tests just compare the output to "GOOD" output.  Is there
tooling somewhere that will check that output generated is valide
opentelemetry JSON output?


Is the following function printing the correct thing for num >= 2 || num < 0?

```
            def dimCount_units(num):
                if num == 0:
                    return "count"
                elif num == 1:
                    return "count x 10"
                else:
                    return "count x 10%d" % num
```

natoscott:
> The PCP QA tests just compare the output to "GOOD" output. 

They also check the test return code FWIW.

> Is there tooling somewhere that will check that output generated is valid opentelemetry JSON output?

I don't think so, from a search and putting the question to a current AI.

> Is the following function printing the correct thing for num >= 2 || num < 0?

AIUI OpenTelemetry units must be in UCUM format - I think that whole function needs a rethink, from my understanding of the standard everything its producing is NQR.  I have some pmUnits->UCUM C code I'll put into a PR shortly that may provide a starting point for new python code.

qa/1977.out
```diff
@@ -0,0 +1,137 @@
...
+	  {
+	    "key": "service.name",
+	    "value": {
+	      "stringValue": "pcp"
+	    }
+	  },
+	  {
+	    "key": "server.address",
+	    "value": {
+	      "stringValue": "http://localhost:54322/receive"
```
natoscott, line 56: what is this server, port and /receive API OOC?

src/pcp2opentelemetry/pcp2opentelemetry.py
```diff
@@ -0,0 +1,675 @@
...
+                numdatapoints_list.append(datapoint_dict)
+            return numdatapoints_list
+
+        def attribute_function():
+            attribute_list = []
+            body = {}
+            body["os.type"] = platform.system().lower()
+            body["telemetry.sdk.language"] = "python"
+            body["telemetry.sdk.name"] = "opentelemetry"
+            body["telemetry.sdk.version"] = "1.24.0"
```
natoscott: Hmm?  We're not using the SDK though - so why put a version here?

We also should not hard-code the os.type like this, as a the tool may be sending data from an archive which was not created on Linux.

As mentioned elsewhere, I think we should treat the PCP context labels as the OTEL resource attributes - this seems like a close match, and means we can reduce the size of he generated JSON significantly (otherwise, these labels are re-produced for every instance of every metric).

src/pcp2opentelemetry/pcp2opentelemetry.py
```diff
@@ -0,0 +1,675 @@
...
+                elif num == 5:
+                    return "PiBy"
+                elif num == 6:
+                    return "EiBy"
+                elif num == 7:
+                    return "ZiBy"
+                elif num == 8:
+                    return "YiBy"
+                else:
+                    return "none"
```
natoscott: "none" is not valid as part of the UCUM specification.

src/pcp2opentelemetry/pcp2opentelemetry.py
```diff
@@ -0,0 +1,675 @@
...
+
+            def dimCount_units(num):
+                if num == 0:
+                    return "count"
+                elif num == 1:
+                    return "count x 10"
+                else:
+                    return "count x 10%d" % num
```
natoscott: I don't think this is valid UCUM format either is it?

src/pcp2opentelemetry/pcp2opentelemetry.py
```diff
@@ -0,0 +1,675 @@
...
+                if unit.dimSpace:
+                    return "%s / %s" % (dimSpace_units(unit.scaleSpace), dimTime_units(unit.scaleTime))
+                elif unit.dimCount:
+                    return "%s / %s" % (dimCount_units(unit.scaleSpace), dimTime_units(unit.scaleTime))
+            elif unit.dimCount < 0:
+                if unit.dimTime:
+                    return "%s / %s" % (dimTime_units(unit.scaleSpace), dimCount_units(unit.scaleTime))
+                elif unit.dimSpace:
+                    return "%s / %s" % (dimSpace_units(unit.scaleSpace), dimCount_units(unit.scaleTime))
+            elif unit.dimSpace:
```
natoscott: All of the above is not UCUM standard compliant AFAICS.

src/pcp2opentelemetry/pcp2opentelemetry.py
```diff
@@ -0,0 +1,675 @@
...
+                return dimSpace_units(unit.scaleSpace)
+            elif unit.dimTime:
+                return dimTime_units(unit.scaleTime)
+            elif unit.dimCount:
+                return dimCount_units(unit.scaleCount)
```
natoscott: These functions look pretty good though.

src/pcp2opentelemetry/pcp2opentelemetry.py
```diff
@@ -0,0 +1,675 @@
...
+            else:
+                return "none"
+
```
natoscott, line 638: Again, "none" here is not valid AIUI.

natoscott (commented):
Overall its getting closer, but I think the units handling and labels/resources need some updates and/or further discussion.

qa/1131
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 # PCP QA Test No. 1131
-# Exercise pcp2json and pcp2openmetrics.
+# Exercise pcp2json, pcp2openmetrics, and pcp2opentelemetry.
 #
 # Copyright (c) 2017 Red Hat.
 #
```
wcohen, line 6: There should be a check to verify that pcp2opentelemetry is available like the checks at lines 18 and 19 for pcp2json and pcp2openmetrics.

src/pcp2opentelemetry/pcp2opentelemetry.py
```diff
@@ -0,0 +1,675 @@
...
+# Our imports
+import requests
+import os
+import cpmapi
+import json
+
+# PCP Python PMAPI
+from pcp import pmapi, pmconfig
+from cpmapi import PM_CONTEXT_ARCHIVE, PM_INDOM_NULL, PM_DEBUG_APPL1, PM_TIME_SEC
```
wcohen: PM_DEBUG_APPL1 not available on some machines.  Looking at the pcp2openmetrics it eliminates this and just uses "appl1" in place of PM_DEBUG_APPL1 in the execute method.

PR #2224, opened 2025-06-22
pmdaopenmetrics metric removal bug fix

lmchilton:
Improvements to metric removal feature.
Added control metrics instance metric
removal.
PMDA also now supports refresh after an
existing config entry is modified
Resolves RHEL-54039

### DIFF
--- a/src/pmdas/openmetrics/pmdaopenmetrics.python
+++ b/src/pmdas/openmetrics/pmdaopenmetrics.python
@@ -32,6 +32,7 @@ import subprocess
 import sys
 from ctypes import c_int
 from socket import gethostname
+from urllib.parse import urlparse
 from stat import ST_MODE, S_IXUSR
 import requests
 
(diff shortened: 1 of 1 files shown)


src/pmdas/openmetrics/pmdaopenmetrics.python
```diff
@@ -1147,12 +1149,39 @@ class OpenMetricsPMDA(PMDA):
...
+                        m = os.path.getmtime(parsed.path)
+                        if mtime is None or m > mtime:
+                            mtime = m
             for d in subdirs:
                 m, _ = self.traverse(os.path.join(path, d), mtime)
                 if mtime is None or m > mtime:
                     mtime = m
         return mtime, ret
 
+    def initalize_controls(self, cluster):
```
wcohen, line 1166: Would be better to avoid the misspelling and call this "initialize_controls".

natoscott: I'll fix this up and merge so we can get this fix into Fedora with the pending update.

PR #2225, opened 2025-06-23
Fix of version check for a keyserver

kurik:
If a system has both Redis and Valkey available, the check for the key server needs to select the correct service to query.

### DIFF
--- a/qa/common.check
+++ b/qa/common.check
@@ -1616,11 +1616,11 @@ _check_series()
 #
 _check_key_server_version()
 {
-    if which valkey-cli >/dev/null 2>&1 && pidof valkey-server >/dev/null 2>&1
+    if which valkey-cli >/dev/null 2>&1 && [ -n "$(_get_pids_by_name valkey-server)" ]
     then
 	keys_cli=valkey-cli
 	keys_ver=valkey_version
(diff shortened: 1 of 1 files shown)


kurik:
For info: The [fedora-rawhide-container](https://github.com/performancecopilot/pcp/actions/runs/15826694022/job/44608636369?pr=2225#logs) is failing due to a known python issue on F43: [BZ 2371769](https://bugzilla.redhat.com/show_bug.cgi?id=2371769). It has nothing to do with this fix.

qa/common.check
```diff
@@ -1616,16 +1616,16 @@ _check_series()
 #
 _check_key_server_version()
 {
-    if which valkey-cli >/dev/null 2>&1
+    if which valkey-cli >/dev/null 2>&1 && pidof valkey-server >/dev/null 2>&1
```
natoscott: @kurik this is the first use of 'pidof' throughout the test suite ... are we a/ sure it exists on all platforms and b/ is always installed?

If not, there is a common function _get_pids_by_name that is used in quite a few places that may make a handy substitute.

kurik: ad a/ no, I am not sure  :-)
ad b/ at least on Linux and BSD systems it should be part of the core system, but I do not have a way how to verify it

I am going to rework the patch using the proposed function .... thanks for the comment.

PR #2226, opened 2025-06-23
batch sql statements to improve efficiency

sfeifer:
This PR looks to improve the efficiency of fetching metrics using pmdahdb by batching sql queries together for metrics from the same table rather than running a new query for each metric.

### DIFF
--- a/src/pmdas/hdb/pmdahdb.python
+++ b/src/pmdas/hdb/pmdahdb.python
@@ -25,7 +25,7 @@ import argparse
 import configparser
 import os.path
 import sys
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 from cpmapi import (
     PM_COUNT_ONE,
@@ -105,7 +105,7 @@ class HDBConnection:
             cursor.execute(sql, parameters)
         except dbapi.ProgrammingError as ex:
             raise dbapi.ProgrammingError(
-                f"ex={ex}, sql='{sql}', paramters={parameters}"
+                f"ex={ex}, sql='{sql}', parameters={parameters}"
             ) from ex
 
         result = cursor.fetchall()
@@ -216,8 +216,11 @@ class Metric:
         :param meta: pmdaMetric
             PCP specific meta data of the metric. See pmdaMetric for details.
         :param query: str
-            SQL query that returns the value of the metric. If the Metric has an associated instance domain,
+            Can be either:
+            a) SQL query that returns the value of the metric. If the Metric has an associated instance domain,
             use a parameterised SQL query.
+            b) If the metric's query has been batched (for efficiency) this field will store the index of the
+            metric's value in the associated table populated by hdb_refresh.
         :param help_text: str
             Long help text. If not set, @desc will be used.
         :param min_hana_revision: str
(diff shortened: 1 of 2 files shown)


src/pmdas/hdb/pmdahdb.python
```diff
@@ -238,11 +241,11 @@ class Metric:
...
-        if meta.m_desc.indom != PM_INDOM_NULL:
-            if " where " not in query.lower():
-                raise ValueError(
-                    "Metric with defined instance domain must have a WHERE clause in query"
-                )
+        #if meta.m_desc.indom != PM_INDOM_NULL:
+        #    if " where " not in query.lower():
+        #        raise ValueError(
+        #            "Metric with defined instance domain must have a WHERE clause in query"
+        #        )
```
sfeifer: @natoscott I'm unsure how to best handle this check now (if we want to handle it at all). The metrics that I batched together fail this check as they no longer have an SQL query in the "query" field, which is why it is commented out. What do you think about removing this check, or do you have a better idea for handling this?

natoscott: This check (and the associated comment above it) no longer makes sense given the nature of your performance optimisations - it can be removed.

Somewhat related, I notice the SYS.M_TRANSACTIONS table fetches still involve many separate SQL roundtrips - you might be able to use something like this SQL to get all those values at once with a single cluster refresh:
https://stackoverflow.com/questions/12789396/how-can-i-get-multiple-counts-with-one-sql-query

natoscott:
@sfeifer looks like some fresh pylint/whitespace issues.  Another thing I just remembered - there's a couple of missing bits here for a production quality PMDA.  If you look over in the postgresql PMDA directory, there is support for pmcheck(1) and pmlogconf(1) integration out-of-the-box - we should add this for HANA as well.  It involves installing two additional files, one a shell script service proble and the other a logging configuration based on metrics availability probe.

sfeifer:
@natoscott I created and pushed those 2 files, but I'd like some feedback on them as I'm not clear on a couple things.

- postgresql.pmcheck uses _ctrl_svc to get the state of the postgresql service. I removed this in hdb.pmcheck so the file looks nearly identical now to uwsgi.pmcheck. I think this is valid because a later line (probe) will fail if the HANA database is not running, but I want to make sure I'm not missing something here.
- I am currently testing hdb.pmcheck by running `./hdb.pmcheck -l`. Is this right, or is there another way I should be testing it to be sure it does what we want? EDIT: I found that running the script with different flags (i.e. -s to test the status) tests different things, but I'm still wondering the different ways I should run this to test functionality.
- I see the pmlogconf files usually pick a subset of metrics, but I'm not sure exactly how this subset was determined. Is there some general guidance on which metrics we want to start logging out of the box?

natoscott:
For pmcheck - yep, best to test using the various flags - also with the SAP HANA DB not running vs running, and the agent install in various states to match both modes ... bit of a matrix there to work through.

In terms of pmlogconf and selecting which metrics to record - its a bit of a dark art and tends to require domain knowledge.  Want to avoid default logging of things that may introduce extra overheads (e.g. some kernel metrics are known to be relatively expensive, so we choose not to log those by default) - if there's a very large or rapidly changing instance domain, we will often not log those by default, etc.  Again, its an art form - use your best judgement and we can always fine-tune it over time (best to be conservative initially, and add more metrics later than the other way around).

PR #2228, opened 2025-06-25
Fix of python3-pcp issue on Python-3.14

kurik:
Fix of a (de)reference of a tuple in python/pmda.c
Resolves bz2371769

### DIFF
--- a/src/python/pmda.c
+++ b/src/python/pmda.c
@@ -401,11 +401,11 @@ refresh_all_clusters(int numclusters, int *clusters)
 	PyList_SET_ITEM(list, i, num);
     }
 
-    arglist = Py_BuildValue("(N)", list);
+    arglist = Py_BuildValue("(O)", list);
+    Py_DECREF(list);
     if (arglist == NULL)
 	return -ENOMEM;
     result = PyObject_Call(refresh_all_func, arglist, NULL);
-    Py_DECREF(list);
     Py_DECREF(arglist);
     if (result == NULL)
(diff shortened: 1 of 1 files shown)


kurik:
For info: The [fedora-rawhide-container](https://github.com/performancecopilot/pcp/actions/runs/15869192250/job/44741780201?pr=2228#logs) is failing due to issue on F43: [BZ 2371769](https://bugzilla.redhat.com/show_bug.cgi?id=2371769). Once this fix is merged and a new build is provided, this should disappear.

natoscott:
@kurik I think there's several other python problems though - not directly related to PCP.  The rawhide CI build is still failing because it cannot install things like pylint and python3-elasticsearch.  (doesn't actually get to the point that it can test your fix due to these earlier issues)

kurik:
@natoscott yes, I know. I verified the fix on rawhide manually. The reason why I tried to push this through the CI is to be sure, there are no regressions on other/older distros. As many tests in the `pcp-testsuite` are covered by `valgrind`, I wanted to be sure the fix will not cause any memory leak or so...

kurik:
@natoscott , all the CI tests seems to be OK. We also have a confirmation from a python developer in https://bugzilla.redhat.com/show_bug.cgi?id=2371769#c15 that the fix should be OK. Can you please merge this and once you will be preparing a Fedora build, backport this patch to it as well ?
Thanks in advance.

natoscott:
@kurik will do - great job diagnosing this BTW, and your fix looks spot on.

PR #2231, opened 2025-07-03
src/selinux/pcp.te: Handle pmlogger perfmon AVCs

wcohen:
On Fedora running Linux 6.15 kernels started seeing AVCs reported by RHBZ2371004:

type=AVC msg=audit(N): avc:  denied  { perfmon } for  pid=PID comm="ps" capability=38  scontext=system_u:system_r:pcp_pmlogger_t:s0 tcontext=system_u:system_r:pcp_pmlogger_t:s0 tclass=capability2 permissive=0

Added a pcp.te entry to address these AVCs.

### DIFF
--- a/src/selinux/pcp.te
+++ b/src/selinux/pcp.te
@@ -352,6 +352,17 @@ optional_policy(`
     userdom_setattr_user_home_content_files(pcp_pmlogger_t)
 ')
 
+optional_policy(`
+    require {
+	type pcp_pmlogger_t;
+	class capability2 perfmon;
+    }
+
+    #RHBZ2371004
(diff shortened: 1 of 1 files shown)


kmcdonell (commented):
@wcohen  I'm no selinux expert, and the change per se LGTM.   So these comments are more at a superficial level.  And I know the suggestions below have not been followed consistently in pcp.te, but consistency will probably help in the long run.

1. I think it makes more sense for all the "optional_policy" clauses for a particular stanza like `pcp_pmlogger local  policy` to come after the non-optional policy clauses, rather than at the beginning in your commit.
2. I've always thought it helped to read the *.te files if there is a comment showing the essential parts of the relevant AVC ahead of an "allow" clause ... when the issue (or something like it) comes up again, it helps to have the original AVC to compare with any new AVC, and this detail is lost by time audit2allow(1) has munched on the AVC.

wcohen:
I have reworked the patch to address the two points mentioned in the earlier review. There was a mention of the AVC message in the git commit comment, but having that in the pcp.te will make it more obvious to people.  Grouping the change with the other option policy will makes things less scattered.

natoscott (commented):
LGTM

PR #2233, opened 2025-07-13
libpcp_web: remove accidental timespec nsec sign extension

natoscott:
QA tests 1617 and 1618 are sometimes failing as a result of this issue, which is a result of the tv_nsec field of struct timespec being sign extended when converted to u64 for a key (time) xstream conversion calculation within libpcp_web.

### DIFF
--- a/qa/1617
+++ b/qa/1617
@@ -155,7 +155,7 @@ pmsignal $pid
 
 # while that happens, check state of the key server
 echo "=== pmseries value check ==="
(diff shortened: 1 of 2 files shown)


kmcdonell:
LGTM, I'm going to merge so I can test in the QA Farm.

kmcdonell:
:+1:
Both qa/1617 and qa/1618 passed on bozo, where they have been hard failures forever.

PR #2238, opened 2025-07-17
pcp-dstat: Add plugins for basic AMD and NVIDIA GPU monitoring

wcohen:
Provide very basic plugin that replicates the overall GPU and memory utilizations provided by nvtop for AMD and NVIDIA GPUs. There is a new -G option in dstat to allow filtering specific GPUs.  The new plugins and the -G options are mentioned in the updated pcp-dstat man page.

Resolves https://github.com/performancecopilot/pcp/issues/2095


natoscott:
Looks great @wcohen - could we extend one of the existing dstat tests, or add a new one perhaps, to use the existing nvidia archive in QA?  (not sure we have an AMD recording, @fberat may know)
```
$ pminfo -a qa/archives/pcp-atop-nvidia.0.xz nvidia
nvidia.numcards
nvidia.cardname
nvidia.busid
nvidia.gpuactive
nvidia.memactive
nvidia.memused
nvidia.memtotal
nvidia.nprocs
nvidia.samples
nvidia.gpuutilaccum
nvidia.memutilaccum
nvidia.memusedaccum
nvidia.proc.all.samples
nvidia.proc.all.memused
nvidia.proc.all.memaccum
nvidia.proc.all.gpuactive
nvidia.proc.all.memactive
nvidia.proc.all.time
nvidia.proc.all.running
nvidia.proc.all.gpulist
nvidia.proc.all.ngpus
```

wcohen:
I looked around in qa/archives but missed the additional archive for nvidia atop metrics.  I agree that it would be good to have some qa test to make sure that plugins work.  Thanks for the pointer to the qa/archives/pcp-atop-nvidia.0.xz.  Is there a similar one with amd gpu data?

natoscott:
I can't see an archive with AMD GPU metrics anywhere as yet.

jhansonhpe:
I have access to nodes with AMD GPUs. If you can point to a How To I can gather the metrics.

natoscott:
@jhansonhpe I've added a shell script over in PR #2239 which should do what we need here - can you try it out?

jhansonhpe:
I will give it a try. HPE has tomorrow off so it will be sometime next week before I can get to it.

natoscott:
No problem.  If there was some workload using the GPUs at the time, such that the values aren't all zeroes, that would be ideal... but even all zeros is helpful.

jhansonhpe:
We have diags that can exercise them. Thanks for the reminder to do that.

wcohen:
I used the mk.amdgpu script from PR #2239  to collect data from a machine with an AMD GPU.  I also wrote a similar mk.nvidiagpu script to collect data on a machine with a machine with a NVIDIA GPU.  The collected data was added to qa/archives, test 1187 was updated to test the GPU plugins and added to this pull request. Does that look like it is good to check in now?

How do you want to handle the mk.amdgpu and mk.nvidia additions.  Pull in the mk.amdgpu from the existing pcp pull request and have me generate a similar pull request for the mk.nvidia?

natoscott:
@wcohen ah that's great Will.  I'll merge that other script - feel free to add your mk.nvidia script into this PR, maybe as a separate commit?  Or not, whatever works.  Consider also compressing those archives before you commit 'em because they all becomes part of the PCP source tarball, where smaller is always better.  Also have a look into qa/archives/GNUmakefile as you may need to list the new archive(s) there in order that they're included in the pcp-testsuite package when doing ./Makepkgs (and rpmbuild, etc).

PR #2242, opened 2025-07-22
pmdaopenmetrics metric removal qa bug fix

lmchilton:
addresses multiple qa test failures. all qa tests
now pass, and updated test #1976 checks for metric removal from a modified source that still persists in the config directory.

### DIFF
--- a/qa/1976
+++ b/qa/1976
@@ -2,7 +2,7 @@
 # PCP QA Test No. 1976
 # Test pmdaopenmetrics metric removal
 #
-# Copyright (c) 2017, 2025 Red Hat.  All Rights Reserved.
+# Copyright (c) 2025 Red Hat.  All Rights Reserved.
 #
 seq=`basename $0`
 echo "QA output created by $seq"
@@ -45,6 +45,15 @@ find $PCP_PMDAS_DIR/openmetrics/config.d -name GNU\* -exec rm -f {} ";"
 
 _pmdaopenmetrics_install
 
+iam=openmetrics
+# append -R option to pmcd config
+sed < $PCP_PMCDCONF_PATH \
+        -e "/^$iam.*/s/$/ -R 1/" \
+        > $tmp.conf
+$sudo cp $tmp.conf $PCP_PMCDCONF_PATH
+$sudo systemctl restart pmcd
+sleep 2
+
 if ! _pmdaopenmetrics_wait_for_metric openmetrics.thermostat
 then
     status=1
@@ -56,13 +65,21 @@ $sudo rm $PCP_PMDAS_DIR/openmetrics/config.d/simple_metric.url
 pminfo openmetrics.simple_metric
 echo
 
+echo "-- check control metrics disappeared --"
+pminfo -dfmt openmetrics.control.status_code
+echo
+
 echo "-- source re-addition --"
 # same access controls logic as above, user $PCP_USER needs to be
 # able to read the file at the end of the URL
 #
 cp $here/openmetrics/samples/simple_metric.txt $tmp.simple_metric.txt
-echo 'file:///'$tmp.simple_metric.txt >$PCP_PMDAS_DIR/openmetrics/config.d/simple_metric.url
+echo 'file:///'$tmp.simple_metric.txt > $PCP_PMDAS_DIR/openmetrics/config.d/simple_metric.url
 pminfo openmetrics.simple_metric
+echo
+
+echo "-- check control metrics reappeared --"
+pminfo -dfmt openmetrics.control.status_code
 echo
 
 echo "-- metric removal of recognized source/metric --"
@@ -76,6 +93,16 @@ echo 'file:///'$tmp.simple_metric.txt > $PCP_PMDAS_DIR/openmetrics/config.d/simp
 $sudo touch -t 197001010000 $PCP_PMDAS_DIR/openmetrics/config.d/simple_metric.url
 pminfo openmetrics.simple_metric
 echo
+
+echo "-- metric removal by modifying source file, source persists --"
+$sudo sed -i -e "/metric2/d" $tmp'.simple_metric.txt'
+
+echo "-- sleep to allow for old_enough_for_refresh() --"
+sleep 2
+echo
+
+echo "-- metric2 removed --"
+pminfo openmetrics.simple_metric
 
 _pmdaopenmetrics_remove >/dev/null 2>&1
 
--- a/qa/1976.out
+++ b/qa/1976.out
@@ -4,9 +4,84 @@ QA output created by 1976
 -- metric removal of new source/metric --
 Error: openmetrics.simple_metric: Unknown metric name
 
+-- check control metrics disappeared --
+
+openmetrics.control.status_code PMID: 144.0.6 [per-end-point source URL response status code after the most recent fetch]
+    Data Type: 32-bit int  InDom: 144.0 0x24000000
+    Semantics: discrete  Units: none
+    inst [0 or "control"] value 0
+    inst [1 or "awk_scripted"] value 0
+    inst [2 or "bad_summary_nometa"] value 0
+    inst [3 or "collectd_sample"] value 0
+    inst [4 or "curl.script"] value 0
+    inst [5 or "curl_filtered"] value 0
+    inst [6 or "curl_hostname_label"] value 0
+    inst [7 or "curl_scripted"] value 0
+    inst [8 or "duplicate_hostname_label"] value 0
+    inst [9 or "good_summary_nometa"] value 0
+    inst [10 or "jenkins_monitoring_javamelody"] value 0
+    inst [11 or "jenkins_prometheus_plugin"] value 0
+    inst [12 or "labelfiltering"] value 0
+    inst [13 or "multiple.namespace.levels"] value 0
+    inst [14 or "multiple.namespace.more"] value 0
+    inst [15 or "pmwebd_3_12_2"] value 0
+    inst [16 or "problematic_strings"] value 0
+    inst [17 or "prom_exposition_formats_example1"] value 0
+    inst [18 or "python_sample_client_server"] value 0
+    inst [19 or "python_scripted"] value 0
+    inst [20 or "reordered_labels"] value 0
+    inst [21 or "sample_pmda_3_12_2"] value 0
+    inst [22 or "sample_pmda_instname_5_0_0"] value 0
+    inst [23 or "sample_pmda_pcp5_metadata"] value 0
+    inst [24 or "sample_prometheus_metrics"] value 0
+    inst [25 or "script_failed"] value 0
+    inst [26 or "sh_script_no_suffix"] value 0
+    inst [27 or "sh_scripted"] value 0
+    inst [29 or "stderr_check"] value 0
+    inst [30 or "thermostat"] value 0
+    inst [31 or "vmware_exporter"] value 0
+
 -- source re-addition --
 openmetrics.simple_metric.metric2
 openmetrics.simple_metric.metric1
+
+-- check control metrics reappeared --
+
+openmetrics.control.status_code PMID: 144.0.6 [per-end-point source URL response status code after the most recent fetch]
+    Data Type: 32-bit int  InDom: 144.0 0x24000000
+    Semantics: discrete  Units: none
+    inst [0 or "control"] value 0
+    inst [1 or "awk_scripted"] value 0
+    inst [2 or "bad_summary_nometa"] value 0
+    inst [3 or "collectd_sample"] value 0
+    inst [4 or "curl.script"] value 0
+    inst [5 or "curl_filtered"] value 0
+    inst [6 or "curl_hostname_label"] value 0
+    inst [7 or "curl_scripted"] value 0
+    inst [8 or "duplicate_hostname_label"] value 0
+    inst [9 or "good_summary_nometa"] value 0
+    inst [10 or "jenkins_monitoring_javamelody"] value 0
+    inst [11 or "jenkins_prometheus_plugin"] value 0
+    inst [12 or "labelfiltering"] value 0
+    inst [13 or "multiple.namespace.levels"] value 0
+    inst [14 or "multiple.namespace.more"] value 0
+    inst [15 or "pmwebd_3_12_2"] value 0
+    inst [16 or "problematic_strings"] value 0
+    inst [17 or "prom_exposition_formats_example1"] value 0
+    inst [18 or "python_sample_client_server"] value 0
+    inst [19 or "python_scripted"] value 0
+    inst [20 or "reordered_labels"] value 0
+    inst [21 or "sample_pmda_3_12_2"] value 0
+    inst [22 or "sample_pmda_instname_5_0_0"] value 0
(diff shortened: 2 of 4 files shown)


qa/1976
```diff
@@ -78,6 +78,17 @@ $sudo touch -t 197001010000 $urlfile
 pminfo openmetrics.simple_metric
 echo
 
+echo "-- metric removal by modifying source file, source persists -- "
+path=$tmp'.simple_metric.txt'
+$sudo cp $here/openmetrics/samples/simple_metric.txt $tmp_file.tmp
+$sudo cp $here/openmetrics/samples/simple_metric.txt $tmp_file2.tmp2
```
natoscott: @lmchilton unless I'm missing something, the $tmp_file and $tmp_file2 shell variables are never created/initialized?  So, I think this will create files named ".tmp" and ".tmp2" which is probably not what you wanted?

qa/1976
```diff
@@ -78,6 +78,17 @@ $sudo touch -t 197001010000 $urlfile
...
+$sudo sed -i '7,9d' $path
+$sudo sed -i '7,9d' $here/openmetrics/samples/simple_metric.txt
+pminfo openmetrics.simple_metric
+$sudo mv $tmp_file.tmp $path
+$sudo mv $tmp_file2.tmp2 $here/openmetrics/samples/simple_metric.txt
+
```
natoscott, line 99: We seem to be doing alot of things twice here that doesn't seem necessary (or I'm misunderstanding some parts).  The sed invocation is also very cryptic ("delete lines 7-9") and may be brittle in future - if that simple_metric.txt file gains new content before line 7, the sed will break.

How about something like this as a more intuitive approach:

```
echo "-- metric removal by modifying source file, source persists --"
cp $here/qa/openmetrics/samples/simple_metric.txt $tmp.saved_metric2
sed -i -e "/metric2/d" $here/qa/openmetrics/samples/simple_metric.txt
pminfo openmetrics.simple_metric
mv $tmp.saved_metric2 $here/qa/openmetrics/samples/simple_metric.txt
```

natoscott: Oh, to clarify the sed a bit - it now means "delete any/all lines containing metric2" (no matter where in the file they are) and the -i means "in-place" so it operates directly on the file (which we'll later restore with the copy we kept via cp).

lmchilton: @natoscott Thanks for the feedback and sed help :) ! The reason I had to delete lines from two different paths is because the deletion from the $here/openmetrics triggers the PMDA to refresh because it picks up a change in a file, but the actual file it draws data from is the data in $path . Do you think there is a better work around for that?

natoscott: @lmchilton no problem - and ah, I see now - you may be able to just use the touch(1) command?  This can change the 'modified time' on the file that the PMDA is using to see if something changed, and this should trigger a refresh on the next fetch.

src/pmdas/openmetrics/pmdaopenmetrics.python
```diff
@@ -948,6 +948,24 @@ class Source(object):
                     # NB: the requests package automatically enables http keep-alive and compression
                     self.document = r.text
 
+            # ditch metrics no longer found in document
+            for name, metric in self.pmda.all_metrics.items():
+                split_metric = name.split(".")
+                stripped_metric = name.replace("openmetrics.%s." % self.name, "")
+                self.pmda.debug("stripped metric --> %s" % stripped_metric) if self.pmda.dbg else None
+                if stripped_metric == "metric2":
```
wcohen: The two lines for checking for "metric2" look to be development code that should be removed now.

qa/1976
wcohen: Looking through the pmdaopenmetrics.python code there is tracking of deleted metric that might be readded later.  The 1976 test doesn't appear to excercise that.  Is there any testing to verify that adding of deleted metric works as expected?

src/pmdas/openmetrics/pmdaopenmetrics.python
natoscott: @lmchilton the debug lines are still there BTW, marking 'unresolved' for now.

src/pmdas/openmetrics/pmdaopenmetrics.python
```diff
@@ -1147,12 +1169,40 @@ class OpenMetricsPMDA(PMDA):
                     if mtime is None or m > mtime:
                         mtime = m
                     ret.append(fname)
+                fname = os.path.join(path, f)
+                with open(fname, 'r') as name:
+                    f_path = name.readline().strip()
+                    if f_path.startswith("file:///"):
```
natoscott: @lmchilton @wcohen I think this approach may be incorrect - why keep mtime on these files?  We cannot treat files differently to a curl response from an endpoint.  The important part is whether the metrics are still there in the response (whether a file or http endpoint was used) - there should be no need to check filesystem state like this, right?

natoscott: This should suffice for QA (patch below).  It doesn't seem to work correctly yet - I think we still have some issues in the PMDA code to resolve - but this removes metric2 (without changing config file timestamps) and reinstates that metric2.  Once this works as shown in the .out change, we can have confidence in this PMDA code change.

```
$ git diff
qa/1976
--- /tmp/git-blob-7FfOb9/1976   2025-08-11 13:57:40.423237960 +1000
+++ qa/1976     2025-08-11 13:57:39.206463438 +1000
@@ -77,6 +77,16 @@ $sudo touch -t 197001010000 $PCP_PMDAS_D
 pminfo openmetrics.simple_metric
 echo
 
+echo "-- remove a metric from an existing source --"
+sed -e '/metric2/d' -i $tmp.simple_metric.txt
+pminfo openmetrics.simple_metric
+echo
+
+echo "-- reinstate metric for an existing source --"
+cp $here/openmetrics/samples/simple_metric.txt $tmp.simple_metric.txt
+pminfo openmetrics.simple_metric
+echo
+
 _pmdaopenmetrics_remove >/dev/null 2>&1
 
 # success, all done
qa/1976.out
--- /tmp/git-blob-B2e7Ur/1976.out       2025-08-11 14:03:04.574975077 +1000
+++ qa/1976.out 2025-08-11 14:03:00.664966764 +1000
@@ -15,3 +15,10 @@ Error: openmetrics.simple_metric: Unknow
 openmetrics.simple_metric.metric2
 openmetrics.simple_metric.metric1
 
+-- remove a metric from an existing source --
+openmetrics.simple_metric.metric1
+
+-- reinstate metric for an existing source --
+openmetrics.simple_metric.metric2
+openmetrics.simple_metric.metric1
+
```

lmchilton: @natoscott @wcohen  I just pushed an update with a working / simplifed qa #1976 test. 

Regarding the added code in function [traverse] keeping track of files mtime, the reasoning behind this is to signal a change in configuration when actual text files are modified and not just the URL holding the path to the file in the config directory. 

old traverse function code picks up a change when $PCP_PMDAS_DIR/openmetrics/config.d/metric.url is modified

new traverse function code picks up a change when $PCP_PMDAS_DIR/openmetrics/config.d/metric.url is modified AND picks up a change when the contents under metric.url (say /some/path/metric.txt) so the contents in metric.txt is modified

lmchilton: I see how we can't treat files differently from curl responses, but it seems imperative that the code picks up modifications in the text files?

natoscott: @lmchilton hi Lauren, yes that's right, we certainly need to detect modifications (in both the text files and network stream mode).  But, looking at the file modification time is not the way to go here - there is no modification time for network traffic - we need a mechanism that can tell from the data itself whether a metric has been removed.

The approach we use in other places (such as the pmdaCache(3) interfaces), is an algorithm more like this:
- initially mark all metrics as deleted for the source we're currently updating
- as we scan through and encounter each refreshed metric in the current update, mark that metric as present
- at the end of the scan, any metrics that remain not yet marked as present are removed from the namespace and no updated values returned

lmchilton: @natoscott OK got it- thank you for clarifying

qa/1976.out
```diff
@@ -4,14 +4,104 @@ QA output created by 1976
...
 openmetrics.simple_metric.metric2
 openmetrics.simple_metric.metric1
 
+-- metric removal by modifying source file, source persists --
+--- trigger a fetch ---
+
+openmetrics.simple_metric.metric2 PMID: 144.28.1 One-line Help: Error: One-line or help text is not available
+    Metadata unavailable
+Error: Unknown or illegal metric identifier
+
```
natoscott, line 95: @lmchilton this behavior (the preceding 4 lines) looks incorrect - pminfo is saying the metric name still exists but all other metadata and values do not.  I suspect the metric name should also no longer exist at this point - is that right?

lmchilton: @natoscott this was something I was struggling with.. I noticed the behavoir of the pmda in QA is slightly different than on my machine; On my machine the PMDA refreshes every 10 seconds (reflected in log file), but QA it doesn't seem to refresh unless metrics are explicitly fetched. I spent time to see if I could fix this behavoir somehow, but looking at other QA tests they seem to behave the same way. So, I decided to label in the test "trigger a fetch" (refresh). Do you think there is a better way to achieve this?

natoscott: Hmm - I don't think this is a QA problem, it looks more like a PMDA code problem.  There seems to be an internal state mismatch where the PMDA thinks a metric (partially) exists - it has a name for it, that it tells pminfo about, but it has no other metadata (pmID, pmDesc) nor values.

IOW, if pminfo isn't made aware of "openmetrics.simple_metric.metric2" (which should no longer exist at that point, AIUI) then it wont ask for the other metadata/values which is why we see these errors in the test output.

So, its probably an issue relating to not having removed that metric name (fully? properly?) from the PMDAs internal view (dicts) of what metrics exist at that time.

lmchilton: @natoscott If you look at the next output line, I call "pminfo openmetrics.simple_metric" and it only outputs metric1. So, the PMDA does correctly remove the metric (metric2); it just needs the intermediate step "pminfo -dfmt openmetrics.simple_metric" to trigger a fetch/refresh so the PMDA executes the code needed to remove the metric. Without that, the PMDA hangs with "clusters = []" when running in the function "refresh_clusters_for_fetch" so nothing happens. Hopefully that explanation makes sense?

lmchilton: @natoscott I also should mention this is a nonissue on my local system because the PMDA is constantly refreshing its clusters. So when metrics are removed from config files it is reflected seemlessly with a call to pminfo.

natoscott: @lmchilton the PMDA will only refresh when something asks it too - probably you have pmlogger running locally and its sampling every 10 seconds.  Try switching that off (systemctl stop pmlogger) and you should observe the same behavior QA sees.

There is definitely a bug in the PMDA here still though, with the way its handling the removed metric name (pminfo should not see this metric name once its been removed, but it is seeing it).

src/pmdas/openmetrics/pmdaopenmetrics.python
```diff
@@ -603,7 +605,7 @@ class Source(object):
         '''
         now = time.time()
         last_try_age = now - self.refresh_time
-        return len(self.metrics_by_name) == 0 and last_try_age > empty_source_pmns_poll
+        return len(self.metrics_by_name) == 0 or last_try_age > empty_source_pmns_poll
```
natoscott: @lmchilton Good catch!

30 seconds is a long time to wait in the middle of a test.  Could you make empty_source_pmns_poll configurable (another command line option), then set it to something small when running the test - e.g. 1 second or half a sec - and sleep just a short time like 2-3sec).

wcohen: Looking through the code it looks like there is the -t/--timeout option that sets the refresh.  By default that is set to 2 seconds.  Would turning that down to 1 second help? Is that "Scan Initial config" code at the bottom of pmdaopenmetrics.python actually waiting 10x of the timeout and then turning it down to the actual value or does it effective just refresh whatever the time out is?  Might just try using a sleep of greater than the default 2 seconds and see if the test works (sleep 3).

lmchilton: @wcohen the -t --timeout option is used in the refresh functions to make sure the get requests for network sources have a timeout, so they do not wait indefinitely. I do not think it has any correlation to the refresh time of clusters. 

@natoscott that makes sense & I added the option to the PMDA. Do you know how to add options to the pmda in QA tests. I know on my system that configuring pmda options lives in /etc/pcp/pmcd/pmcd.conf . Is it the same place for QA?

natoscott: @lmchilton yes, pmcd.conf is the place that needs to be updated.  Will need to be done after an initial ./Install - agent options are always a bit of a pain to manage, historically the Install scripts were more interactive and allowed command line options to be manually entered, but these days everything needs to be done in an automated fashion, so these options only tend to get used in QA settings and Install tends to no longer support 'em.

natoscott: @lmchilton qa/725 is an example of a test that temporarily modifies an agents parameters in pmcd.conf

src/pmdas/openmetrics/pmdaopenmetrics.python
```diff
@@ -1583,6 +1622,11 @@ if __name__ == '__main__':
         type=int,
         default=2,
         help='HTTP GET timeout for each end-point URL (default 2 seconds)')
+    parser.add_argument(
+        '-rf', '--refresh',
```
wcohen: Generally a single '-' is associated with single letter options like "-l", "-r", etc.  Maybe use a case sensitive "-R" to avoid collision with the existing "-r" "--root" option.

qa/1976
```diff
@@ -45,6 +45,15 @@ find $PCP_PMDAS_DIR/openmetrics/config.d -name GNU\* -exec rm -f {} ";"
 
 _pmdaopenmetrics_install
 
+iam=openmetrics
+# append option to look from /proc/fs/jbd2
+sed < $PCP_PMCDCONF_PATH \
+        -e "/^$iam.*/s/$/ -R 1/" \
+        > $tmp.conf
+$sudo cp $tmp.conf $PCP_PMCDCONF_PATH
```
natoscott, line 53: @lmchilton we'll need to undo this change before the test exits.
See the use of "_save_config $PCP_PMCDCONF_PATH", and _restore_config, in other tests for examples.

lmchilton: @natoscott the function _prepare_pmda openmetrics calls _save_config & saves the original pmcd.conf file then the call in cleanup to _pmdaopenmetrics_cleanup calls  _cleanup_pmda openmetrics which in turn calls _restore_config 

So, I think it's taken care of?

wcohen:
I have taken a look at the Oct 3 version of the patch.  It looks reasonable to me.

qa/1976
natoscott: @lmchilton right you are - excellent - nothing to do here 👍

PR #2243, opened 2025-07-23
pcp2openmetrics tool causing hanging

lmchilton:
bug found of tool outputting one metric
then hanging until manually exiting with
control c. deletion of line setting new
context to UTC fixed bug and allowed tool
to continuously output metrics or up to
the sample size.

### DIFF
--- a/src/pcp2openmetrics/pcp2openmetrics.py
+++ b/src/pcp2openmetrics/pcp2openmetrics.py
@@ -39,11 +39,14 @@ CONFVER = 1
 INDENT = 2
 TIMEFMT = "%Y-%m-%d %H:%M:%S"
 TIMEOUT = 2.5 # seconds
+os.environ["TZ"] = "UTC"
+time.tzset()
 
 class PCP2OPENMETRICS(object):
     """ PCP to OPENMETRICS """
     def __init__(self):
         """ Construct object, prepare for command line handling """
+
         self.context = None
         self.daemonize = 0
         self.pmconfig = pmconfig.pmConfig(self)
@@ -390,7 +393,6 @@ class PCP2OPENMETRICS(object):
             # Silent goodbye, close in finalize()
             return
 
-        self.context.pmNewZone("UTC")
         ts = self.context.datetime_to_secs(self.pmfg_ts(), PM_TIME_SEC)
 
         if self.prev_ts is None:


natoscott:
How do we guarantee the timestamps are in UTC timezone without something like this?  (AIUI this is mandatory for OpenTelemetry, no?).

natoscott:
> How do we guarantee the timestamps are in UTC timezone without something like this? (AIUI this is mandatory for OpenTelemetry, no?).

It may be possible to do this as the first thing the program does:
```
os.environ["TZ"] = "UTC"
```

Setting the context timezone (your removed line) after command line arg processing (and setting defaults) may be the root cause of the "hang" - its possible the tool was waiting until a start time (defaults to "now", but ended up shifted several hours) - maybe.

wcohen:
Completely removing the code to set the time correct zone doesn't seem like the correct approach. Thinking it would be better to move the `self.context.pmNewZone("UTC")` to the `connect` method right after `self.context` is initialized.  Like the following which allowed `pcp2openmetrics  -t1 -s2 -x hinv.ncpu` to work properly:

```
$ git diff
diff --git a/src/pcp2openmetrics/pcp2openmetrics.py b/src/pcp2openmetrics/pcp2openmetrics.py
index 8d8b62066..3a1c8d97e 100755
--- a/src/pcp2openmetrics/pcp2openmetrics.py
+++ b/src/pcp2openmetrics/pcp2openmetrics.py
@@ -312,6 +312,7 @@ class PCP2OPENMETRICS(object):
         self.pmfg_ts = self.pmfg.extend_timeval()
         self.context = self.pmfg.get_context()
 
+        self.context.pmNewZone("UTC")
         if pmapi.c_api.pmSetContextOptions(self.context.ctx, self.opts.mode, self.opts.delta):
             raise pmapi.pmUsageErr()
 
@@ -390,7 +391,6 @@ class PCP2OPENMETRICS(object):
             # Silent goodbye, close in finalize()
             return
 
-        self.context.pmNewZone("UTC")
         ts = self.context.datetime_to_secs(self.pmfg_ts(), PM_TIME_SEC)
 
         if self.prev_ts is None:


```

lmchilton:
@natoscott @wcohen Thank you both. Thanks Nathan for that explanation, that would make sense. I remember we added that line, but I forgot the exact reasoning for the addition, so I thought maybe it was not needed anymore. But now I remember we needed to ensure that no matter the timezone the tool processed time in a unified way.  Thanks

natoscott:
@lmchilton @wcohen I have a vague memory that this maybe caused a problem in the past too - so, worth checking the git logs (but maybe I'm misremembering, could have been a pcp2opentelemetry discussion).

wcohen:
@natoscott I tried doing some digging around with gitk to see whether `pmNewZone` had be in other places in `pcp2openmetrics.py`.  It was moved around 2024-07-02 in git commit 2cdde889bb17a5c54909669e398fddf87c61c5d1 to right after the `pmSetContextOptions()` then reverted 2024-07-4 in git commit f9a75c4e473558101ced660c6448a6e505f5d007 saying that it was an unintended.  However, there wasn't anything describing the specific problems.

One difference between the earlier patches and this patch is that the `pmNewZone("UTC")` is called before `pmSetContextOptions()`.  I don't know if there might be some interaction between `pmNewZone()` and `pmSetContextOptions()`

There is also 6cefac3c6560487abad80498240545e6835e503f from 2024-6-19 that adds the pmNewZone("UTC").

wcohen:
This updated patch is really close.  However, looks like need to have a `time.tzset()` in there to update the timezone info. This can be demoed with a a littlt python script:

```
#!/usr/bin/env pmpython

# importing time and os module
import time, os

# Define TZ environment variable
os.environ['TZ'] = 'UTC'

# print time
print(time.strftime('%X %x %Z'))

# reset the time conversion rules
time.tzset()

# print time
print(time.strftime('%X %x %Z'))
```

Note that the first print out is the local timezone, while the second is in the expected UTC time zone:

```
$ ./tz.py
11:54:55 07/29/25 EDT
15:54:55 07/29/25 UTC

```

PR #2256, opened 2025-08-04
pcp-atop: failure of cpu stat accumulation

lmchilton:
fixes failure of thread cpu utilization
accumulation. Fix includes sorting processes
based on thread ID and accumulation of cpu
stats into the 'parent' process; where
parent process is defined by PID=TGID

resolves RHEL-79767


src/pcp/atop/photoproc.c
```diff
@@ -34,6 +34,54 @@ extern regex_t  envregex;
...
+}
+
+static void
+cpu_accumulation(struct tstat **tasks, int count)
+{
+	qsort(*tasks, count, sizeof(struct tstat), thread_sorter);
+
+	struct tstat *parent = NULL;
+	struct tstat *current;
+	int64_t cpu_use_u, cpu_use_s = 0;
```
natoscott: Perhaps change '_u' to '_utime' and '_s' to '_stime' here for clarity.

src/pcp/atop/photoproc.c
```diff
@@ -34,6 +34,54 @@ extern regex_t  envregex;
...
+		if (parent == NULL)
+		{
+			parent = current;
+			cpu_use_u = current->cpu.utime;
+			cpu_use_s = current->cpu.stime;
+		}
+		else if (parent->gen.tgid != current->gen.tgid)
+		{
+			(&(*tasks)[p_index])->cpu.utime = cpu_use_u;
+			(&(*tasks)[p_index])->cpu.stime = cpu_use_s;
```
natoscott: Instead of using p_index and this "(&(*tasks)[p_index])" expression, use a "task *previous;" - that'd be alot more readable I think.

src/pcp/atop/photoproc.c
```diff
@@ -34,6 +34,54 @@ extern regex_t  envregex;
...
+			cpu_use_s = current->cpu.stime;
+		}
+		else
+		{
+			cpu_use_u += current->cpu.utime;
+			cpu_use_s += current->cpu.stime;
+		}
+		if (current->gen.pid == current->gen.tgid)
+			p_index = i;
+	}
```
natoscott, line 98: There's a bug here in that the last group doesn't get correctly accounted (i.e. the case where we reach the end of the array - 'i == count' - we don't go through the code path that accumulates the CPU usage.  We'll need an update outside the loop to handle that case.

src/pcp/atop/photoproc.c
```diff
@@ -365,6 +413,8 @@ photoproc(struct tstat **tasks, unsigned long *taskslen)
 		update_task(&(*tasks)[i], pids[i], insts[i], result, descs, offset);
 	}
 
+	cpu_accumulation(tasks, count);
```
natoscott: Suggest renaming this to 'thread_cpu_accumulation' so its more clear what it does and that its associated with the qsort callback which is named that way.

natoscott (changes requested):
This PR is missing any regression test updates.  It'd be worth running the new binary with the existing ~15 or so pcp-atop[sar] tests because its likely they'll be exercising these code paths already (and will possibly be failing as a result of this change?).

lmchilton:
Does this bug need code addition to QA testsuite to verify cpu accumulation?

natoscott:
> Does this bug need code addition to QA testsuite to verify cpu accumulation?

I don't think you'll need code - you might be able to observe the effects of the change in the current tests?  If so, check the output makes sense, then "./remake" the expected test output.  If not, see if there's some other way to observe the effect of your fix via a QA test (new test, or additional case in an existing atop test).  If its not possible at all, mention it in the commit message and describe the manual verification (e.g. to assist other folk doing RHEL QA of the fix subsequently :)

src/pcp/atop/photoproc.c
wcohen: Yes, I noticed that the last set of threads aren't accumulated also.  Probably need something like the following immediately after the loop:

```
  if (parent != NULL)
     parent->cpu.utime = cpu_use_utime;
     parent->cpu.stime = cpu_use_stime;

```

wcohen:
The patch certainly changes qa/1748.out and qa/1978.out, so there is something testable there.  Looking through the  SYSCPU and USRCPU values they are increased  for various PIDs which would be expected if the patch is now adding in the previously unaccounted threads times.  Looking through qa/1748.out and qa/1978.out the new order of the pids seems to make sense.

The thread_cpu_accumultaion function now addresses the accumulation of the threads for the last PID group.

The current patch looks good to me.

PR #2257, opened 2025-08-05
pmda.smart Add smart.wwid.* metric branch which uses drive uuid for i…

pauljevans:
…nstance

Add support to track drive instances by their wwid identifiers through the smart.wwid.* branch.

Also mask zram devices from being added as instances.


natoscott:
@pauljevans looks like QA test 1397 is failing on all the daily CI hosts with this change - could you take a look?  Taa.  

https://performancecopilot.github.io/qa-reports/reports/20250805_212914-501c7247/#suites/324a1b8ff0a59f4b9c5518c86b42ac3c/7a85cfcdaadb9d1c/

natoscott:
@pauljevans additionally, the coverity scan is reporting a TOCTOU security-related issue here:
```
        /* Get disk UUID if not already populated */
        n = pmsprintf(buffer, sizeof(buffer), "/sys/block/%s/device/wwid", name);
        if (n <= 0 || access(buffer, F_OK) != 0) /* try alternate path */
                n = pmsprintf(buffer, sizeof(buffer), "/sys/block/%s/wwid", name);
        if (n > 0 && (fd = open(buffer, O_RDONLY)) >= 0) {
```
Its largely benign - there is a tiny (theoretical?) race condition where access() succeeds then the open() fails. 
However the code would be cleaner if it was rearranged to not use access(2) and instead directly open(2) the first path, and failing that falling back to a direct open call on the second path.

The scan also found a small resource leak on an error path in uuid_instance_refresh - that one's trivial, I'll push through a fix there shortly.

PR #2266, opened 2025-08-08
libpcp: fix memleak on indom instlist seen in pmproxy QA

natoscott:
Picked up by valgrind in pmproxy SIGHUP memory leak tests.


kmcdonell:
LGTM, the my original code is missing the conditional free.

PR #2269, opened 2025-08-09
rc_scripts: remove legacy rc_pcp (pmcd+pmlogger) script

natoscott:
This has not been used since systemd times, and even before it was not the preferred way since PCP circa v4.

Removing it resolves systemd warnings like these:

> systemd-sysv-generator[PID]: SysV service '/etc/init.d/pcp' lacks a native systemd unit file, automatically generating a unit file for compatibility.
> systemd-sysv-generator[PID]: Please update package to include a native systemd unit file.


kmcdonell (commented):
source code changes all LGTM

For the QA changes, I wonder if we need to get rid of _service pcp [stop|start|restart] because there is a lot of logic in _service() to handle this using the base pmcd and pmlogger services|scripts (not the pcp init.d script).  We could just leave all this alone and it will still "just work" (tm) when /etc/init.d/pcp is no longer there.

If you do want to want to whack all the _service pcp uses, the I think the _service() code in common.check should be cleaned up to remove and error for _service pcp &lt;anything&gt;, and in the individual tests, I think template below is a better replacement and reduces the risk of races.

_service pcp stop
```
_service pmlogger stop || _exit 1
_wait_pmlogger_end || _exit 1
_service pmcd stop || _exit 1
_wait_pmcd_end || _exit 1
```
[_wait_pmlogger_end will need a small tweak to treat the missing pid argument as the pid of the primary pmlogger (if it is running), but I have that already done]

And similar (but slightly different) templates for start (and restart).
```
_service pmcd start || _exit 1
_wait_for_pmcd || _exit 1
_service pmlogger start || _exit 1
_wait_for_pmlogger || _exit 1
```
I know it is a lot of changes, and late release cycle, so I'd suggest merging the source code changes and we (you or me) cycle back after the 7.0 release and clean up the QA scripts.

Of course you may have evidence of QA failures with just the source change, but I don't think such a massive QA refactor (in terms of the number of files, not scope) is necessarily needed.

natoscott:
Good points, one and all - I'll undo the QA changes, that certainly simplifies things alot.

PR #2283, opened 2025-08-18
pmproxy, libpcp_web: fix QA valgrind issue with new source leak

natoscott:
When new context labels arrive after establishing a metrics source, SHA1 identifiers change; handling this correctly requires different mechanisms for resetting a context vs setting up a new one.

While there, improve the locking around lookup for a new archive context by removing a small race between what was previously two hash table lookups in loggroup_new_archive.


natoscott:
This needs additional work (fixes one issue but adds another), will reopen when I get time to return to it.

PR #2284, opened 2025-08-18
build: add Debian13 into CI/QA now it has released

natoscott:


christianhorn:
Looks like something early, special to the build process here..
FWIW, on Trixie/amd64, I can build with "configure/make", but Makepkgs fails with:
```
[..]
dh_compress -a
dh_installdirs -a
dh_install -a --sourcedir=debian/pcp
dh_install: warning: Cannot find (any matches for) "usr/lib/sysusers.d/pcp-testsuite.conf" (tried in debian/pcp, debian/tmp)

dh_install: warning: pcp-testsuite missing files: usr/lib/sysusers.d/pcp-testsuite.conf
dh_install: error: missing files, aborting
[..]
```
Comparing with Bookworm, I see that on Trixie file pcp-testsuite.conf is not installed in the first place, seeing that earlier in the log. Maybe some conditional in debian/* area which is not executed on Trixie? Full logs for Debian12/13 at https://fluxcoil.net/files/tmp/deblogs/ .

natoscott:
@christianhorn OK, we're reproducing the issue you saw locally now in CI.  The issue is a side-effect of this from ./configure:
```
checking if systemd should be used... no
```
which ultimately comes from this snippet in configure.ac:
```
    PKG_CHECK_VAR([SYSTEMD_SYSTEMUNITDIR], [systemd], [systemdsystemunitdir], 
        [pcp_systemdunit_dir=$SYSTEMD_SYSTEMUNITDIR], [enable_systemd=false])
    AC_SUBST(pcp_systemdunit_dir)
    if test -z "$pcp_systemdunit_dir"
    then
        if test "$do_systemd" = "yes"
        then
            AC_MSG_ERROR([cannot enable systemd support - no systemunitdir path])
        fi
        enable_systemd=false
    fi
```

... so, either the systemd pkg-config path identifiers (particularly "systemdsystemunitdir") have different names, or they're not present at all, or something along those lines - note sure what yet, needs further investigation.

christianhorn:
Interestingly, configure/make on Trixies also produces something for me - I have not tried it out though.
Makes sense though, I see that part as "no" in the logs of plain configure and also Makepkgs on Trixie.

natoscott:
I think what's happened is the pkg-config file has moved from systemd package into a new systemd-dev package - trying a CI build with that installed now.

Issue #2272, opened 2025-08-10
On the Debian package pcp, pmda-postfix is not shipped


christianhorn:
On Debian we ship PMDAs as part of the pcp package, instead of having separate packages for the PMDAs, like i.e. on Fedora.
PCP build scripts check for existence of the qshape script at build time. On Debian, it's not one of the build requirements, so pmda-postfix is not getting built.

On Debian, qshape is part of the Postfix package. If we require postfix for building the package, then later at runtime we

a) Could have package pcp to depend on package postfix. This would enforce postfix on all users, not acceptable. It would also mean this solution would apply to all other PMDAs.. 
b) would install pmda-postfix as part of pcp, but we would not have a proper way to tell the user "you need to install package postfix to use this pmda".

Is there an other solution?
qshape is a 13kb Perl script, shipping it in Postfix package allows it to be tight down to the installed Postfix version.
On Fedora, there is the separate package pcp-pmda-postfix package. It's also not depending on full blown Postfix package, but on the more lean postfix-perl-scripts, which then contains script qshape.

Right now, as a user on Debian, to use pmda-postfix I have to
* understand the above
* install package Postfix
* compile own PCP packages
* and then do the above again when either PCP gets updates I want to catch, or build dependencies get updates which I want. Or abstract the above with an own pmda-postfix implementation via pmda-openmetrics, that would allow me to stick with the Debian pcp package.

Not unlikely that we have seen this situation already?

natoscott:
> On Debian we ship PMDAs as part of the pcp package, instead of having separate packages for the PMDAs, like i.e. on Fedora. PCP build scripts check for existence of the qshape script at build time. On Debian, it's not one of the build requirements, so pmda-postfix is not getting built.
> 
> On Debian, qshape is part of the Postfix package. If we require postfix for building the package, then later at runtime we

We should add a build time dependency on this package, using the existing configure.ac macros and additions to the scripts below ./debian/

> a) Could have package pcp to depend on package postfix. This would enforce postfix on all users, not acceptable.

Yeah, this is not the way to go.

> Is there an other solution? qshape is a 13kb Perl script, shipping it in Postfix package allows it to be tight down to the installed Postfix version. On Fedora, there is the separate package pcp-pmda-postfix package. It's also not depending on full blown Postfix package, but on the more lean postfix-perl-scripts, which then contains script qshape.
> 

[...]

PR #2286, opened 2025-08-18
Add postfix as build-dependency on Debian

christianhorn:
Add postfix as build-dependency on Debian. 
That ensures pmda-postfix is getting built and shipped with package 'pcp'.
If the Debian system installing 'pcp' has then package 'postfix' installed, then script qshape is available, and the pmda can be used.

In the unlikely event that a system has not 'postfix' installed but the installation of pmda-postfix is attempted, this is output:
```
root@debian12:/var/lib/pcp/pmdas/postfix# ./Install
Postfix qshape utility is not installed
```
Tested on Debian Bookworm.
Fixes https://github.com/performancecopilot/pcp/issues/2272 .


natoscott:
postfix support should already be in place (grep -r postfix debian) without this - I think there's something else gone wrong here, probably its not installed on the host where the source upload is done, so doesn't get stamped into the build as something we want - I'll check.

natoscott:
The build seems to be working OK (without these changes), once postfix is installed on the source build machine - I'll ensure this is the case for the next upload.  For your own local ./Makepkgs builds @christianhorn you'll need to do the same and should see this (Makepkgs is a funny case because it runs ./configure prior to doing any platform-specific rpm/deb/... builds, and the results of configure create the rpm/deb/... build deps to some extent - and in this case, that can filter out postfix):

```
$ dpkg --listfiles pcp | grep postfix
/etc/pcp/dstat/postfix
/usr/lib/pcp/pmdas/postfix
/usr/lib/pcp/pmdas/postfix/Install
/usr/lib/pcp/pmdas/postfix/Remove
/usr/lib/pcp/pmdas/postfix/pmdapostfix.pl
/usr/share/man/man1/pmdapostfix.1.gz
/usr/share/pcp/htop/meters/postfix
/var/lib/pcp/pmdas/postfix
/var/lib/pcp/pmdas/postfix/Install
/var/lib/pcp/pmdas/postfix/Remove
/var/lib/pcp/pmdas/postfix/pmdapostfix.pl
```

natoscott:
@christianhorn I'm gonna close this as I think we have deb postfix support already, as per earlier notes, and the underlying problem is at my end when I update Debian via the source upload.  I've made changes now to ensure that issue doesn't bite again.

PR #2296, opened 2025-08-20
PMDAopentelemetry: parsing structure change

lmchilton:
changed the parsing structure to use python json module

WIP: Metric Class: unit function & metric removal is not complete due to pmdaopenmetric metric removal devel work still in progress

### DIFF
--- a/src/pmdas/opentelemetry/pmdaopentelemetry.python
+++ b/src/pmdas/opentelemetry/pmdaopentelemetry.python
@@ -35,6 +35,7 @@ from ctypes import c_int
 from socket import gethostname
 from stat import ST_MODE, S_IXUSR, ST_CTIME
 import requests
+import json
 
 from pcp.pmapi import pmUnits, pmContext
 from pcp.pmda import PMDA, pmdaMetric, pmdaIndom, pmdaInstid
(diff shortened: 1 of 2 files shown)


src/pmdas/opentelemetry/pmdaopentelemetry.python
```diff
@@ -445,19 +507,30 @@ class Source(object):
...
+        labels = {}
+        for item in attribute_list:
+            key = item["key"]
+            value = item["value"]["stringValue"]
+            labels[key] = value
+        return labels
+    
+    def pcp_name(self, name):
+        return name.replace(".", "_")
+
```
natoscott, line 520: @lmchilton let's drop this pcp_name routine - reading more about OpenTelemetry, I think we can use PCP names unchanged and dotted notation is in fact a more fitting style there.  I updated pmproxy earlier in this way also.

natoscott (commented):
@lmchilton looking good!  See note inline re metric names.  Also, can we drop the example.txt (it doesn't seem to match the code currently anyway?)?

There's some CI failures relating to pylint - small stuff to clean up, looks like.

OOC, can this version parse the output from pmproxy and pcp2opentelemtry now?  This could provide some neat QA opportunities too.

natoscott:
@lmchilton I resolved the branch conflict, pylint issues and qa/1947 failure, then merged.  So closing this one now.
I guess at this point the correct OTel UCUM -> PCP units handling is going to have to wait for after pcp-7.0.0 - we're well within code freeze now, so looking only for critical fixes at this time.

PR #2313, opened 2025-08-28
pcp-mpstat: refactor to use vuser/vnice metrics and unify CPU utilization calculation

orasagar:
Switch all references in MPSTAT_METRICS from cpu.user/cpu.nice to the newer cpu.vuser/cpu.vnice metrics for both kernel.all and kernel.percpu to be inline with mpstat tool. 
Refactor all per-metric calculation methods in CoreCpuUtil to delegate to a new _compute_metric helper, eliminating redundant code.
Optimize CPU count retrieval by caching the result of hinv.ncpu in the CoreCpuUtil class. 
Remove unused or redundant calls to cpu_online, clarifying the core class interface. 
Simplify and centralize the per-metric computation logic by using self._all_or_percpu() for proper metric path construction and handling instance (per-core vs global) logic in a single location.


natoscott:
@orasagar thanks!  do we need updates to QA tests?  And/or to the src/pmlogconf/tools/mpstat config file?

orasagar:
@natoscott yes i think we need to update those,let me do that.

PR #2316, opened 2025-09-03
qa: Enable pylint installation in F43 test env.

kurik:
pylint is already working in F43 distro. Let's enable it for the test environment (this is a partial revert of the pull request #2235).


natoscott:
@kurik do you know if rawhide is OK too?  (there's a Fedora+44+x86_64 file now too)

kurik:
> @kurik do you know if rawhide is OK too? (there's a Fedora+44+x86_64 file now too)

Thanks @natoscott for the reminder. Yes, it is working on F44 as well. I just updated the pull request.

PR #2317, opened 2025-09-03
pmdagfs2: Add support for glockfd debugfs metric collection

pauljevans:
Add support for collecting metrics from the GFS2 glockfd debugfs file which keeps track of glocks "held" open by file descriptors.


qa/654.out
```diff
@@ -196,6 +196,29 @@ the lock_time metrics or not. The machine must have the gfs2 trace-points
...
+    inst [1 or "loop0::holder_1"] value 9
+
+gfs2.glockfd.process PMID: 115.8.0 [List of process PID with glock holders per filesystem]
+    Data Type: 64-bit unsigned int  InDom: 115.1 0x1cc00001
+    Semantics: instant  Units: count
+Help:
+List of process PID with glock holders per filesystem
+    inst [0 or "loop0::holder_0"] value 4923
+    inst [1 or "loop0::holder_1"] value 4960
+
```
natoscott, line 214: File descriptors and process identifiers can be represented by signed 32 bit values (e.g. see return code of open(2), I don't think these need to be 64 bits.

src/pmdas/gfs2/glockfd.c
```diff
@@ -0,0 +1,145 @@
+/*
+ * GFS2 glockfd file statistics.
+ *
+ * Copyright (c) 2013-25 Red Hat.
```
natoscott: Can you make that 2025 like in all other places?  thanks :)

src/pmdas/gfs2/glockfd.h
```diff
@@ -0,0 +1,51 @@
...
+enum {
+    GLOCKFD_PROCESS = 0,
+    GLOCKFD_FILE_DESCRIPTOR,
+    NUM_PER_HOLDER_STATS
+};
+
+struct glockfd {
+    __uint64_t total;
+    __uint64_t process[MAX_NUMBER_OF_HOLDERS];
+    __uint64_t file_descriptor[MAX_NUMBER_OF_HOLDERS];
```
natoscott: 32 bits for the second two fields please.  And, hmm, is this a long-lived structure?  Seems slightly wasteful of memory if so (alternatively, may be a target for buffer overrun) - can we dynamically size these arrays.

src/pmdas/gfs2/glockfd.h
```diff
@@ -0,0 +1,51 @@
...
+};
+
+struct glockfd_per_instance {
+    __uint64_t holder_id;
+    __uint64_t process;
+    __uint64_t file_descriptor;
+};
```
natoscott, line 41: 32 bits for the second two fields please, not sure what holder_id is though.

src/pmdas/gfs2/pmda.c
```diff
@@ -267,6 +266,19 @@ pmdaMetric metrictable[] = {
...
         PM_TYPE_U64, GFS_FS_INDOM, PM_SEM_INSTANT,
         PMDA_PMUNITS(0,0,1,0,0,PM_COUNT_ONE) }, },
+    /* GLOCKFD*/
+    { .m_desc = {
+        PMDA_PMID(CLUSTER_GLOCKFD, GLOCKFD_TOTAL),
+        PM_TYPE_U64, GFS_FS_INDOM, PM_SEM_INSTANT,
+        PMDA_PMUNITS(0,0,1,0,0,PM_COUNT_ONE) }, },
+    { .m_desc = {
+        PMDA_PMID(CLUSTER_HOLDER_STATS, GLOCKFD_PROCESS),
+        PM_TYPE_U64, GFS_HOLDER_INDOM, PM_SEM_INSTANT,
```
natoscott: PM_TYPE_32 here (and below).  And I think the units should be all zeros, not with 1/PM_COUNT_1 anywhere (again below as well).

src/pmdas/gfs2/pmns
```diff
@@ -92,6 +93,12 @@ gfs2.glstats {
     journal		GFS2:2:9
 }
 
+gfs2.glockfd {
+    total		GFS2:7:0
+    process		GFS2:8:0
+    file_descriptor	GFS2:8:1
```
natoscott: we've tended to use the short-hands "fds" in other places, e.g.

pmcd.openfds
openmetrics.grafana.process_open_fds
openmetrics.grafana.process_max_fds

Similarly, we could more descriptively use "pids" instead of "process" here too.

pauljevans:
Hi Nathan,

Thanks for the review, I've made the suggested changes and updated the PR.

PR #2318, opened 2025-09-04
aarch64 containers

natoscott:


kurik:
I will not comment on `.github/workflows/main.yml`, I will leave it to Sam :-)

Re the change in Dockerfile:
```
RUN dnf remove -y selinux-policy selinux-policy-targeted || true
```
This does not work for me as the `selinux-policy-targeted` is a protected package. To make it more versatile I would propose to remove all potential selinux policies by using
```
RUN dnf remove -y --setopt=protected_packages= selinux-policy\* || true
```

Otherwise I am fine with those changes.

natoscott:
Hmm.  Currently, when building the container without this change, I see:

```
[1/2] STEP 6/8: RUN dnf remove -y selinux-policy || true
Failed to resolve the transaction:
Problem: installed package selinux-policy-targeted-42.7-1.fc42.noarch requires selinux-policy = 42.7-1.fc42, but none of the providers can be installed
  - conflicting requests
  - problem with installed package
--> bc3bfdde837c
```

And you're right - with the change I see:

```
[1/2] STEP 6/8: RUN dnf remove -y selinux-policy selinux-policy-targeted || true
Failed to resolve the transaction:
Problem: The operation would result in removing the following protected packages: selinux-policy-targeted
```

So, I'll try your suggestion :) - cos currently, everything is fail.  (but it always just carries on with the build regardless)

.github/workflows/main.yml
```diff
@@ -48,4 +52,4 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-          
+          platforms: ${{ matrix.arch }}
```
sfeifer: I think this should be `linux/${{ matrix.arch }}` based on the docker documentation [here](https://docs.docker.com/reference/cli/docker/buildx/build/#platform). This sentence in particular: "For example, linux/amd64 or linux/arm/v7."

sfeifer:
This change results in main.yml essentially running 2 jobs: one for amd64 and one for arm64. As I understand it, this results in 2 separately tagged images (one for each architecture). There is a way to have a single tag with both amd64 and arm64 images, but I’m not sure if this is what we want as they would not be easily distinguishable from each other by their names (i.e. the only difference in their names would be some hash). The link [here](https://docs.docker.com/build/ci/github-actions/multi-platform/) describes this functionality.

natoscott:
@kurik your final suggestion seems to be the only one that works, so I'll go with that - thanks!

PR #2320, opened 2025-09-04
build: add zstd package as a hard dependency

natoscott:
A colleague encountered a "corrupt archive" - turned out it was an archive copied from one container to another and the second had PCP but didn't have zstd installed:

```
# pmrep -a 20250902.18.35 kernel.all.load
__pmProcessPipe: child pid=530309 execvp(zstd, ...) failed pmrep: Corrupted record in a PCP archive
```

Which is somewhat cryptic; took me a little while to figure out the root cause, despite it being in the error message - the "corrupted record" bit distracted me I guess.

This commit makes zstd a hard dependency because I think we will see more of this scenario now on (remote) replay hosts since we choose to use this form of compression more often, in recent releases.

### DIFF
--- a/build/rpm/pcp.spec.in
+++ b/build/rpm/pcp.spec.in
@@ -132,7 +132,7 @@ BuildRequires: ncurses-devel
 BuildRequires: readline-devel
 
 # Utilities used indirectly e.g. by scripts we install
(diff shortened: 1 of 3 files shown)


kmcdonell (approved):
All look good to me.
And I agree with the argument for making this change.

PR #2335, opened 2025-09-17
Rename the container build files

natoscott:


natoscott:
Thanks @sfeifer I've added a README describing the situation.

Issue #2305, opened 2025-08-27
Add macOS into the Release process


tallpsmith:
When a release tag is created, the current OS distributions are automatically built under GitHub CI workflows and uploaded to PackageCloud.  We should include macOS into this process.

@natoscott mentioned in Slack that it might be better done as a standalone GitHub action workflow to keep them separated.

PR #2337, opened 2025-09-18
DRAFT: Improve Mac CI workflow to include automatic Release process on Tag detection

tallpsmith:
Add automatic detection of a release-numbered-tag as the git reference and automatically upload the generated DMG into the Github releases area.  

This leverages `action-gh-release` module which automatically sets things up and would re-use an existing Github Release object if it's already there, so it shouldn't destroy anything created either manually, or by the other existing Linux-based distribution release steps in `release.yml`.

Addresses #2305


natoscott:
| PLEASE for the love of all that is holy, squash all these commits... :-D

You can do that locally ("git rebase -i main") , right?  Then "git push --force ..." into this branch FTW, updating this PR.

tallpsmith:
> | PLEASE for the love of all that is holy, squash all these commits... :-D
> 
> You can do that locally ("git rebase -i main") , right? Then "git push --force ..." into this branch FTW, updating this PR.

Yeah ,but you can also do it from this interface when merging, right?  as a reviewer, I would always like to see the 'story' of how the changes came and then merge into a single commit (if appropriate).  I hate force pushing, and I prefer to have the history of the branch right up until merge time.

build/mac/build-installer
```diff
@@ -114,6 +114,8 @@ _do mv $PKG $PACKAGE
 _do $HDIUTIL makehybrid -quiet -o $PACKAGE.tmp $PACKAGE \
 	-hfs-openfolder dmg.tmp/$DMG -hfs
 _do $HDIUTIL convert -quiet $TEMPDMG -format UDRO -o $DMG
+# get rid of a tmp file that is generated as part of the DMG build, as this complicates the Release uploaod process
```
natoscott: typo on "uploaod"

.github/workflows/macOS.yml
```diff
@@ -53,3 +56,26 @@ jobs:
...
+      - name: Generate Github Summary - no version tag
+        if: steps.check_tag.outputs.match == ''
+        run: |
+          echo "Not a Release!" >> $GITHUB_STEP_SUMMARY
+      - name: Release
+        uses: softprops/action-gh-release@v2
+        if: steps.check_tag.outputs.match != ''
+        with:
+          files: |
+              pcp-**/build/mac/pcp-*.dmg
```
natoscott, line 81: Is that pattern matching supposed to be two stars up front or would just one do? (like in the second part)

natoscott (approved):
Fabulous, thanks @tallpsmith - few minor suggestions inline, but all good.

.github/workflows/macOS.yml
tallpsmith: This is Javascript 'glob' library which will match multiple segments (e.g. directories, but also by habit I use it for covering off dotted numbers).  I think either way works.

PR #2348, opened 2025-09-23
Some minor fixes to eliminate some warning on the Code Scans

wcohen:
There were a few xz_debug() statements in libpcp and libpcp3 that didn't have quite the right format specifications for arguments being printed out.  These arguments caused some warnings on the [upstream pcp security scan](https://github.com/performancecopilot/pcp/security/code-scanning).  A couple minor patches to make the format specification and the printed value match in type are in this pull request.

### DIFF
--- a/src/libpcp3/src/io_xz.c
+++ b/src/libpcp3/src/io_xz.c
@@ -251,7 +251,8 @@ parse_indexes(FILE *f, size_t *nr_streams)
       pos -= LZMA_STREAM_HEADER_SIZE;
       (*nr_streams)++;
 
-      xz_debug("%s(%d, ...): decode stream footer at pos = %d", __func__, fileno(f), pos);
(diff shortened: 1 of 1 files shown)


natoscott:
@wcohen ah, we need to be careful using %ld as long changes sizes between 64/32 bit platforms (I suspect the warning will move from one to the other with this change).  An approach we've used in other places is to use %lld in the format string, and cast the corresponding argument to (long long) - this is 64-bit on all PCP platforms.

wcohen:
Ah, that's a good point about about this patch causing issues on some platforms. I will rework the patch.

wcohen:
Reworked the patches to use %lld and (long long) casts.

natoscott:
> Reworked the patches to use %lld and (long long) casts.

LGTM.

PR #2350, opened 2025-09-24
Update QA 821.out with new mem.numa.util metrics

orasagar:
Add output for additional mem.numa.util.* in qa/821.out to reflect recent changes and expanded coverage, ensuring test output aligns with updated PCP metric sources.


orasagar:
@natoscott 
I’ve collected NUMA metrics from the multi-node system and packaged them into a new .tar file as recommended. However, while running test 821, I noticed unexpected occurrences of the string "scan linux table" appearing randomly in the output. I couldn’t pinpoint the root cause—it might be related to the dev system environment.

Although i have manually removed them ,if you feel and our build servers has an issue we can drop this PR.

natoscott:
@orasagar I'm not seeing that diagnostic here - could it be be a local change you made?  there is a pmdalinux source file named src/pmdas/linux/linux_table.c which is related to this work you've been doing, but it doesn't contain that exact string.

orasagar:
@natoscott yeah probably my local issue.

PR #2367, opened 2025-10-10
libpcp_web: add testcase and fix some bugs

kimullaa:
I added some libpcp_web test case, and fixed bug.


qa/run.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+socat UNIX-LISTEN:/tmp/httpcache.sock,fork TCP-CONNECT:localhost:44323 &
+pid=$!
+trap "kill $pid" EXIT
+exit
```
natoscott: @kimullaa it looks like this file is some leftover diagnostics/testing that's been incorporated into the main test now - can it be removed?

kimullaa: oops.... thank you for a comment. I fixed it .

PR #2368, opened 2025-10-10
Fixes hugepagesize metric value from bytes to KB as per proc/meminfo

sourav-sharma796:
Before:
[localhost@vbox ~]$ pcp meminfo | grep Hugepagesize ; cat /proc/meminfo | grep Hugepagesize
Hugepagesize      : 2097152 kB
Hugepagesize:       2048 kB

After:
[local@localhost ~]$ pcp meminfo | grep Hugepagesize ; cat /proc/meminfo | grep Hugepagesize
Hugepagesize      : 2048 kB
Hugepagesize:       2048 kB

### DIFF
--- a/src/pmdas/linux/help
+++ b/src/pmdas/linux/help
@@ -1745,7 +1745,7 @@ of the number of fragments by combining them as they are received.
 @ hinv.pagesize Memory page size
 The memory page size of the running kernel in bytes.
 @ hinv.hugepagesize Huge page size from /proc/meminfo
-The memory huge page size of the running kernel in bytes.
+The memory huge page size of the running kernel in kilobytes.
 @ hinv.ncpu number of CPUs in the system
 @ hinv.ndisk number of disks in the system
 @ hinv.nfilesys number of (local) file systems currently mounted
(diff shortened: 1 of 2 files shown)


natoscott:
I think this may be a pcp-meminfo bug, not a pmdalinux bug ... the units look correct here, no?  Or have I missed some subtlety?

```
$ pminfo -df mem.hugepages.pagesize

mem.hugepages.pagesize
    Data Type: 64-bit unsigned int  InDom: 60.43 0xf00002b
    Semantics: instant  Units: Kbyte
    inst [0 or "2048kB"] value 2048
    inst [1 or "32768kB"] value 32768
    inst [2 or "64kB"] value 64
    inst [3 or "1048576kB"] value 1048576
```

natoscott:
And same for the older hugepagesize metric mentioned here:
```
pcpqa@fedora:~$ pminfo -tT hinv.hugepagesize -fd

hinv.hugepagesize [Huge page size from /proc/meminfo]
    Data Type: 32-bit unsigned int  InDom: PM_INDOM_NULL 0xffffffff
    Semantics: discrete  Units: byte
Help:
The memory huge page size of the running kernel in bytes.
    value 2097152
```

this looks correct to me - I guess pcp-meminfo is making an incorrect assumption, no?

natoscott:
Fixed via #2371

PR #2371, opened 2025-10-13
pcp-meminfo: fix handling of hinv.hugepagesize which uses bytes

natoscott:
Related #2368


natoscott:
I'm going to merge this as I'm pretty sure the units are correct as-is, its a simple fix in pcp-meminfo, and we have a release pending in a day or two so I'm keen to get this resolved asap.

PR #2375, opened 2025-10-17
Fix null pointer dereference in pmproxy

kurik:
A null pointer dereference in libpcp_web when handling battons causing pmproxy to crash (core dump). The crash happens from time to time when pmproxy is under load and serving data in paralel to more pmproxy clients. This patch fixes the null pointer dereference.

Resolves: RHEL-70195
Resolves: bz2401579

### DIFF
--- a/src/libpcp_web/src/batons.c
+++ b/src/libpcp_web/src/batons.c
@@ -129,13 +129,11 @@ seriesPassBaton(seriesBatonPhase **head, void *arg, const char *caller)
     seriesBatonMagic	*baton = (seriesBatonMagic *)arg;
 
     if (baton == NULL) {
-        if (pmDebugOptions.series)
-            fprintf(stderr, "%s: NULL baton from %s\n", __FUNCTION__, caller);
+        pmNotifyErr(LOG_ERR, "%s: NULL baton from %s\n", __FUNCTION__, caller);
         return;
     }
     if (head == NULL) {
-        if (pmDebugOptions.series)
-            fprintf(stderr, "%s: NULL head from %s\n", __FUNCTION__, caller);
+        pmNotifyErr(LOG_ERR, "%s: NULL baton from %s\n", __FUNCTION__, caller);
         return;
     }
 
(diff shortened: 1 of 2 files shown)


kurik:
I did quite extensive testing of this patch and so far, no crashes (core dumps) were observed with this patch applied. Unfortunately there is no direct way how to reliably reproduce this null pointer dereference, so it is not possible to test it with 100% certainty.

src/libpcp/src/logutil.c
```diff
@@ -623,18 +628,21 @@ logFreeHashText(__pmHashCtl *type_ctl)
...
+	    if (ident_ctl->hash != NULL)
+		free(ident_ctl->hash);
 	    free(ident_ctl);
 	    free(curr_type_node);
 	}
     }
-    free(type_ctl->hash);
+    if (type_ctl->hash != NULL)
+	free(type_ctl->hash);
 }
```
natoscott, line 638: All of the changes in this file are no-ops, from the free(3) man page section on free (2nd paragraph):
"If ptr is NULL, no operation is performed."

src/libpcp_web/src/batons.c
```diff
@@ -128,13 +128,28 @@ seriesPassBaton(seriesBatonPhase **head, void *arg, const char *caller)
     seriesBatonPhase	*next;
     seriesBatonMagic	*baton = (seriesBatonMagic *)arg;
 
+    if (baton == NULL) {
+        if (pmDebugOptions.series)
+            fprintf(stderr, "seriesPassBaton: NULL baton from %s\n", caller);
```
natoscott: this is fine, but better to use __FUNCTION__ with"%s: ... from %s\n" here - means we don't ever get mismatched diagnostics if code is refactored, and generally shorter error strings embedded in libraries.

src/libpcp_web/src/load.c
```diff
@@ -1162,6 +1162,11 @@ freeSeriesLoadBaton(seriesLoadBaton *baton)
 void
 doneSeriesLoadBaton(seriesLoadBaton *baton, const char *caller)
 {
+	if (baton == NULL) {
+		if (pmDebugOptions.series)
+	        fprintf(stderr, "doneSeriesLoadBaton: NULL baton from %s\n", caller);
+        return;
```
natoscott, line 1169: same comment as in the other file (__FUNCTION__ use in diagnostics).

kurik:
@natoscott  Makes sense to me - reworked.

PR #2379, opened 2025-10-22
QA: fix qa 867

kimullaa:
follow the QA 983 style(cleanup the socat using pmsignal)

It works in my environment, but I haven't confirmed whether it works in the CI environment.

```
$ ./check 867
Warning: changing umask from 0002 to 0022
867 0s ...
Passed 1 tests
$ echo $?
0
$ ps -ef | grep socat
kimullaa 2558431 2497823  0 21:01 pts/2    00:00:00 grep --color=auto socat
```

### DIFF
--- a/qa/867
+++ b/qa/867
@@ -17,46 +17,78 @@ _check_series
 
 _cleanup()
 {
+    [ -z "$socat_pid" ] || $sudo $signal $socat_pid
+    [ -f "$tmp.socat.out" ] && cat $tmp.socat.out >> $seq_full
+    [ -f "$tmp.socat.err" ] && cat $tmp.socat.err >> $seq_full
+    [ -z "$proxy_pid" ] || $sudo $signal $proxy_pid
+    [ -n "$options" ] && $keys_cli $options shutdown
     cd $here
     $sudo rm -rf \$tmp \$tmp.*
 }
 
 status=1	# failure is the default!
+signal=$PCP_BINADM_DIR/pmsignal
 trap "cd $here; _cleanup; exit \$status" 0 1 2 3 15
 
 unset http_proxy
 unset HTTP_PROXY
 
 # real QA test starts here
-if ! _service pmproxy restart >/dev/null 2>&1; then _exit 1; fi
+echo "=== Start test key server ..."
+key_server_port=`_find_free_port`
+options="-p $key_server_port"
+$key_server --port $key_server_port --save "" > $tmp.keys 2>&1 &
+_check_key_server_ping $key_server_port
+echo
+echo "=== Start test pmproxy ..."
+proxyport=`_find_free_port`
+username=`id -u -n`
+cat > $tmp.conf <<EOF
+[pmproxy]
+pcp.enabled = true
+http.enabled = true
+resp.enabled = true
+[pmseries]
+enabled = true
+servers = localhost:$key_server_port
+EOF
+$PCP_BINADM_DIR/pmproxy -f -p $proxyport -s $tmp.socket \
+        -U $username -l $tmp.log -c $tmp.conf &
+proxy_pid=$!
+pmsleep 0.5
+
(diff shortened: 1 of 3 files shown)


kmcdonell (changes requested):
Hi, @kimullaa ... this is still not working for me, I'm afraid.

1. you need to rm -f /tmp/httpcache.sock before running socat
2. the test hangs in httpcache

Here is some debugging I've been able to collect ...

$ sh -x 867
\+ basename 867
\+ seq=867
\+ echo QA output created by 867
...
\+ rm -f /var/tmp/867-19435.socat.out /var/tmp/867-19435.socat.err /tmp/httpcache.sock
\+ socat_pid=19586
\+ socat -d -d unix-listen:/tmp/httpcache.sock,fork tcp-connect:localhost:44323
\+ echo == Check HTTP cache
\+ tee -a /home/kenj/src/pcp/qa/867.full
== Check HTTP cache
\+ /home/kenj/src/pcp/qa/src/httpcache

and httpcache never exits.

$ cat /var/tmp/867-19435.socat.err
2025/10/23 07:45:00 socat[19586] N listening on AF=1 "/tmp/httpcache.sock"

$ cat /var/tmp/867-19435.socat.out
<empty>

$ pstree -p 19435
sh(19435)─┬─httpcache(19589)
          └─socat(19586)

$ gdb -p 19589 ~/src/pcp/qa/src/httpcache
...
(gdb) where
#0  0x000075589f32bbfd in __libc_recv (fd=3, buf=0x7ffe848bf910, len=8192, 
    flags=0) at ../sysdeps/unix/sysv/linux/recv.c:28
#1  0x000075589f63844a in __pmRecv () from /lib/libpcp.so.4
#2  0x000075589f5476ca in ?? () from /lib/libpcp_web.so.1
#3  0x000075589f54864a in pmhttpClientGet () from /lib/libpcp_web.so.1
#4  0x000055c0640b729c in call_http_get (
    url=0x55c0640b8070 "http://localhost:44323/series/metrics?match=kernel.all.load") at httpcache.c:19
#5  0x000055c0640b7463 in main (argc=1, argv=0x7ffe848c1b38) at httpcache.c:65

Hope that's enough for you to figure out the issue(s).

kmcdonell:
Sorry @kimullaa this is still not working for me.  It now hangs in the first iteration of the added socat block.

kmcdonell:
@kimullaa Still no luck for me ... it is back to hanging in httpcache.
I think the issue may be you're pmproxy config which is not default.
If this is the case, your test needs to take control and start a new pmproxy with the config you need, then cleanup afterwards.  qa/1214 shows how to start a private pmproxy and key server (if you need that).

kimullaa:
@kmcdonell thank you for a advice!That’s very helpful.

Hmm, I don’t think I’ve changed the configuration.

```
diff -u /etc/pcp/pmproxy/pmproxy.conf src/pmproxy/pmproxy.conf
```

> If this is the case, your test needs to take control and start a new pmproxy with the config you need, then cleanup afterwards. qa/1214 shows how to start a private pmproxy and key server (if you need that).

It is a good practice, so I’ll incorporate that.

I hope this fix resolves the issue, but if it doesn’t,I'd like to reproduce the hang myself — could you tell me about your environment, such as which distribution you're using?

kmcdonell:
G'day @kimullaa

Latest version works 100% for me ... :beers:

Just one small nit in httpcache.c ... please use pmsprintf() in place of sprintf() ... the extra (2nd) argument protects against buffer overrun (it is not really an issue here, but various compilers and code analysers we use will complain and require the change to be made anyway to keep our "no compiler warnings or code analyser issues" goal).

For reference my main failing system is a 64-bit Ubuntu 24.04, but I think taking control of your own key server and pmproxy instance was the key (excuse the pun), not the platform per se.

Cheers, Ken.

PR #2383, opened 2025-10-27
pmlogconf: Updates to pmlogconf configurations reflecting filesystem and storage updates

pauljevans:
Updates to add some default pmlogconf configurations covering some of the more recent filesystem and storage PMDA changes.


natoscott:
Could we put all of these in the src/pmlogconfs/storage subdirectory @pauljevans ?

pauljevans:
Sounds like a plan, have moved the newly added logging configurations into the storage directory.

PR #2386, opened 2025-10-28
pmdafarm: Extend support for SCSI (SAS) FARM log

pauljevans:
Extend support for SCSI (SAS) FARM log to also include per-actuator stats for a given drive from the log.
    
Updated QA/1540 to reflect these changes.

Further efficiency gains by reducing the number of calls to strtrim per refresh of FARM log data.

### DIFF
--- a/qa/1540.out
+++ b/qa/1540.out
@@ -853,28 +853,28 @@ Help:
 The total number of recorded logical sectors written on the drive
     inst [0 or "sda"] value 148199221061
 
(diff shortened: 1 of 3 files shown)


natoscott (commented):
@pauljevans looking good.  With the units, can you clarify - I see many metric names that relate to counts involving "sectors", and I see those metrics are exported in bytes - should there be a multiplication somewhere (I may have missed it) that converts from sectors (physical? logical?) into bytes?

pauljevans:
Hi Nathan,

Thanks for the review, have made corrections to the units for the additional metrics, these are all basic "number of" counters.

PR #2392, opened 2025-10-30
pmlogconf: Fix configuration for FARM metrics logging

pauljevans:
Fix logging configuration for Farm metrics in pmlogconf.

Correct first probe statement mistake.

Remove second probe statement (fixes fail with QA/1430)

### DIFF
--- a/src/pmlogconf/storage/smart
+++ b/src/pmlogconf/storage/smart
@@ -1,6 +1,4 @@
 #pmlogconf-setup 2.0
 ident	S.M.A.R.T drive statistics [Linux]
-probe	smart.info values ? include : exclude
-	smart.info
-	smart.health
-	smart.attributes
+probe	smart.info.device_model values ? include : exclude
+	smart


src/pmlogconf/storage/farm
```diff
@@ -1,5 +1,4 @@
 #pmlogconf-setup 2.0
 ident	Seagate Field Accessible Reliability Metrics (FARM) Log [Linux]
-probe	smart.ata values ? include : exclude
-probe	smart.scsi values ? include : exclude
+probe	farm.ata
```
natoscott: I think this must be a leaf node metric too @pauljevans - from pmlogconf(1):
```
       •  probe control lines have the format:
                                      probe metric [condition [state_rule] ]
          where metric is the name of a PCP metric (must be a leaf node in the PMNS, no instance specifi‐
          cation is allowed, and it must not be a derived metric) and the optional condition is [...]
```

pauljevans:
Follow-up updates to the patcheset. 

Have split the FARM log logging configuration into separate ATA and SCSI configurations. 

Also added a further commit correcting the probe statement for the SMART logging configuration.

QA tests 364, 894, 1097, 1430, 1681 and 1940 are all passing on my side now.

PR #2400, opened 2025-11-04
libpcp: remove local strchrnul definition entirely

natoscott:
This is problematic on macOS platforms, as it is there for some versions but not others, and we only build the one DMG image. Conflicts exist within pcp-htop where htop tackles the problem in a similar way (creating a local strchrnul) which complicates the problem further if libpcp also defines it.

Fortunately, there are almost no uses and we can easily refactor the two discover.c cases.


Explorer09:
Um, no. The CI error in htop is about pcp's header (pmapi.h) defines HAVE_STRCHRNUL that's would conflict with htop's config.h when building. Having a local definition of strchrnul in pcp code is fine. But to prevent further problems, a better approach for pcp is to define it as an internal function (something like `_pm_strchrnul`), not exposed to pmapi.h, and cannot be linked from outside the library. That would solve it.

natoscott:
This approach is best for avoiding future conflicts between PCP & htop on this symbol IMO, most importantly there is *no need* for a local strchrnul - which is 2/3 lines of code anyway! - and now zero users on non-Linux platforms in PCP.

Explorer09:
> This approach is best for avoiding future conflicts between PCP & htop on this symbol IMO

Sure, but not worth repeating the same idiom across different places in the codebase. (That is what inline functions are for.)

natoscott:
What different places?  There are none, after so many years of PCP - that's my point.  We can safely remove it.

Explorer09:
> What different places? There are none, after so many years of PCP - that's my point. We can safely remove it.

What I see is three calls that would lose the benefit of code size if you force-inline them.

Anyway, it's your decision, not mine.

natoscott:
> What I see is three calls [...]

You're looking at Linux kernel metrics code there, which is unaffected by this change (strchrnul is always available on Linux, so Linux-specific code can continue to use it).  Its core code - like libpcp - that is under the spotlight here.

PR #2403, opened 2025-11-05
pmdaopenmetrics: fix benchmark performance bug

lmchilton:
Resolves RHEL-124769. Clusters were being refreshed hundreds of times / second because the configuration directory was populated with sources that were not connected to the PMDA. This caused the length of the metric dictionary to always equal 0 and the conditional in the function old_enough_for_refresh to always return true. The solution now checks if the source metric dict is equal to 0 AND if its the first time connecting to the source. Otherwise call for refresh every timeout seconds.

### DIFF
--- a/src/pmdas/openmetrics/pmdaopenmetrics.python
+++ b/src/pmdas/openmetrics/pmdaopenmetrics.python
@@ -565,6 +565,7 @@ class Source(object):
         self.document = None
 
         self.refresh_time = 0 # "never"
+
         if not is_scripted:
             # source is a URL.  Create a session for it and initialize a few things
             self.requests = self.pmda.requests # allow persistent connections etc.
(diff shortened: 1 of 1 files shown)


src/pmdas/openmetrics/pmdaopenmetrics.python
```diff
@@ -601,7 +603,13 @@ class Source(object):
...
         now = time.time()
         last_try_age = now - self.refresh_time
-        return len(self.metrics_by_name) == 0 or last_try_age > self.pmda.refresh_timeout
+
+        if len(self.metrics_by_name) == 0 and self.first_refresh is True:
+            refresh = True
+        else:
+            refresh = last_try_age > self.pmda.refresh_timeout
+
+        return refresh
```
natoscott: This change would be a little simpler not adding a new 'refresh' variable...?  (also don't think "is True" is needed)

```
        if len(self.metrics_by_name) == 0 and self.first_refresh:
            return True
        return last_try_age > self.pmda.refresh_timeout
```

Also, hmm, isn't  len(self.metrics_by_name) always zero on the first refresh?

natoscott: @lmchilton still has the 'refresh' variable ... can that be removed?

natoscott:
@lmchilton nice that's slimmed down even further.  Looks like this fix is needed in opentelemetry too, so I've made that change there and fixed up the method comment.  Can you double check it please?

PR #2404, opened 2025-11-05
pmlogger selinux fix for observed denials

sfeifer:
This PR adds a rule to pcp.te to fix an observed AVC denied message while testing.

### DIFF
--- a/src/selinux/pcp.te
+++ b/src/selinux/pcp.te
@@ -307,6 +307,7 @@ allow pcp_pmlogger_t self:unix_dgram_socket create_socket_perms;
 
 allow pcp_pmlogger_t pcp_pmlogger_exec_t:file execute_no_trans;
 allow pcp_pmlogger_t ldconfig_exec_t:file { execute execute_no_trans };
+allow pcp_pmlogger_t pcp_pmie_exec_t:file execute;
 
 dontaudit pcp_pmlogger_t self:cap_userns { sys_ptrace };
 


kurik:
LGTM

src/selinux/pcp.te
```diff
@@ -307,6 +307,7 @@ allow pcp_pmlogger_t self:unix_dgram_socket create_socket_perms;
 
 allow pcp_pmlogger_t pcp_pmlogger_exec_t:file execute_no_trans;
 allow pcp_pmlogger_t ldconfig_exec_t:file { execute execute_no_trans };
+allow pcp_pmlogger_t pcp_pmie_exec_t:file execute;
```
natoscott, line 310: this seems to be mixing pmlogger and pmie ... ?

sfeifer: We observed a new AVC denied message in the automated testing for fedora builds. The change introduced in the build was additional selinux rules....so we think something elsewhere, like the selinux-policy, has changed causing this new denial. The rule was created based on the AVC message: 
```
type=AVC msg=audit(N) : avc:  denied  { execute } for  pid=PID comm=pmiectl name=pmie dev="nvme0n1p4" ino=INO scontext=system_u:system_r:pcp_pmlogger_t:s0 tcontext=system_u:object_r:pcp_pmie_exec_t:s0 tclass=file permissive=0 
```

natoscott: This explanation doesn't make sense.  The tools involved here are pmiectl, pmie and (I guess?) pmie_check / pmie_daily, etc.  These are all types from pcp policy, not core selinux policy.  And there should be no pmlogger labels involved here - this seems more like a labeling problem on the filesystem where the tests are being run - can you try a restorecon(1) there?

This is definitely not the right fix though - there should be no overlap between pmie and pmlogger labelling, they are completely separate services.

sfeifer: @kurik Could this just be an issue with the test environment? I recall you were able to reproduce this on the latest fedora rawhide compose.

kurik: At first, I am sorry for being too convincing when we discussed this issue with @sfeifer  and I pointed @sfeifer to a wrong direction. Thanks @natoscott for reviewing this, which forced me to do a proper investigation of the issue.

I was not able to find exactly what has changed either in PCP or in Fedora system and why we have not seen this issue before. However the root cause is in `/usr/bin/pmiectl` resp. `/usr/bin/pmlogctl` which is triggered by `pmie_check.service` resp.  by `pmlogger_check.service`. These two scripts are in fact just one script being hard-linked to two scripts. The script at the beginning identifies it self if it was started as `pmiectl` or as `pmlogctl` and then it is taking care of `pmie` or `pmlogger ` respectivelly.

This is OK, however selinux is not happy about this as to manage `pmie` the script needs to be running in `pmie` context while to manage `pmlogger` it needs `pmlogger` context. This command shows the issue, where we can see that even `pmiectl` is having pmlogger context:
```
# ll -iZ /usr/bin/pm*ctl
183108 -rwxr-xr-x. 2 root root system_u:object_r:pcp_pmlogger_exec_t:s0 53600 Nov  6 07:38 /usr/bin/pmiectl
183108 -rwxr-xr-x. 2 root root system_u:object_r:pcp_pmlogger_exec_t:s0 53600 Nov  6 07:38 /usr/bin/pmlogctl
```

As a workaround we can split the hard-link and set proper security context to `/usr/bin/pmiectl`
```
# cp /usr/bin/pmiectl /usr/bin/pmiectl.x; rm /usr/bin/pmiectl; mv /usr/bin/pmiectl.x /usr/bin/pmiectl
# chcon -u system_u -r object_r -t pcp_pmie_exec_t /usr/bin/pmiectl
# ll -iZ /usr/bin/pm*ctl
145455 -rwxr-xr-x. 1 root root system_u:object_r:pcp_pmie_exec_t:s0     53600 Nov  6 11:57 /usr/bin/pmiectl
183108 -rwxr-xr-x. 1 root root system_u:object_r:pcp_pmlogger_exec_t:s0 53600 Nov  6 07:38 /usr/bin/pmlogctl
```
This fixes the issue and no AVC is triggered by those "*_check" services then.

I like the concept of sharing one script for two different purposes, however I do not know how to set the proper security context then. Having two identical scrips with different security contexts (as the workaround) does not seem to be very elegant. Any thoughts how solve this in a better way ?

natoscott:
@kurik I believe your analysis is correct.  This commit from last year should tackle the issue for pmiectl + pmlogctl, but looks like its not in whatever version you have there?

```
$ git show e85172532d00cd6f4ea50ee7442baa7f33cafeb1
commit e85172532d00cd6f4ea50ee7442baa7f33cafeb1
Author: Nathan Scott <nathans@redhat.com>
Date:   Wed Oct 23 11:30:31 2024 +1100

    selinux: install pmiectl as a script for correct selinux label

diff --git a/src/pmlogctl/GNUmakefile b/src/pmlogctl/GNUmakefile
index acc3c40eb6..efc4cac21a 100644
--- a/src/pmlogctl/GNUmakefile
+++ b/src/pmlogctl/GNUmakefile
@@ -23,9 +23,10 @@ include $(BUILDRULES)
 
 install : default
        $(INSTALL) -m 755 pmlogctl.sh $(PCP_BIN_DIR)/pmlogctl$(SHELLSUFFIX)
+       # not installed as a symlink so that selinux file context applies
+       $(INSTALL) -m 755 pmlogctl.sh $(PCP_BIN_DIR)/pmiectl$(SHELLSUFFIX)
        $(INSTALL) -m 755 -d $(PCP_ETC_DIR)/pcp/pmlogger/class.d
        $(INSTALL) -m 644 pmlogger-pmfind.class $(PCP_ETC_DIR)/pcp/pmlogger/class.d/pmfind
-       $(INSTALL) -S $(PCP_BIN_DIR)/pmlogctl$(SHELLSUFFIX) $(PCP_BIN_DIR)/pmiectl$(SHELLSUFFIX)
        $(INSTALL) -m 755 -d $(PCP_ETC_DIR)/pcp/pmie/class.d
        $(INSTALL) -m 644 pmie-pmfind.class $(PCP_ETC_DIR)/pcp/pmie/class.d/pmfind
 ```

**Comment by kurik** [comment]
I am using the latest build of PCP ( `pcp-7.0.2-1.fc44.x86_64`)  on the latest Fedora rawhide (compose 2025-11-07-06-26). The fix is there (in source rpm). However the problem is slightly different, than the fix. The fix is replacing symlink with a copy of the file, while the problem we have is hardlink between `pmiectl` and `pmlogctl`.

**Comment by kurik** [comment]
I guess I found the real root cause.

When I download RPMs manually (from koji) and install these manually using `rpm`, all works great. `pmiectl` and `pmlogctl` are independent files, each having its own i-node and own selinux context.
```
# rpm -i pcp-7.0.2-1.fc44.x86_64.rpm pcp-conf-7.0.2-1.fc44.x86_64.rpm pcp-libs-7.0.2-1.fc44.x86_64.rpm pcp-selinux-7.0.2-1.fc44.x86_64.rpm
# ls -ilZ /usr/bin/pm*ctl
145091 -rwxr-xr-x. 1 root root system_u:object_r:pcp_pmie_exec_t:s0     53604 Oct 14 20:00 /usr/bin/pmiectl
165529 -rwxr-xr-x. 1 root root system_u:object_r:pcp_pmlogger_exec_t:s0 53604 Oct 14 20:00 /usr/bin/pmlogctl
```

However when I install PCP rpms using `yum`/`dnf`/`dnf5` (which are all just aliases to `dnf5`) then `pmiectl` and `pmlogctl` are hard-linked together.
```
# dnf5 install -y pcp
# ls -ilZ /usr/bin/pm*ctl
165678 -rwxr-xr-x. 2 root root system_u:object_r:pcp_pmlogger_exec_t:s0 53604 Oct 14 20:00 /usr/bin/pmiectl
165678 -rwxr-xr-x. 2 root root system_u:object_r:pcp_pmlogger_exec_t:s0 53604 Oct 14 20:00 /usr/bin/pmlogctl
```

It seems like `dnf` started to do some optimization and is now hard-linking binary identical files together. I will try to reach DNF people inside of RedHat and verify with them what is going on .....

**Comment by kurik** [comment]
I was pointed to https://fedoraproject.org/wiki/Changes/Hardlink_identical_files_in_packages_by_default which was implemented approx. a month ago. That is the root cause.

Reported regression in Fedora Rawhide: https://bugzilla.redhat.com/show_bug.cgi?id=2413314

**Comment by kurik** [comment]
From the https://bugzilla.redhat.com/show_bug.cgi?id=2413314
```
So I think there are two choices for pcp:
1. add 'BuildRequires: pcp-selinux'
2. opt out of the hardlinking by adding '%undefine __os_install_post_hardlink'
```

What do you think about it guys ?

natoscott:
A third option would be to split pmlogctl into a shared script with common code, and two separate front-end scripts.  This would remote all special-handling here (incl. the little comment I added to the makefile when removing the symlink).  I think that would fix this permanently and without the (slight) fragility of the other approaches.

natoscott:
Resolved by https://github.com/performancecopilot/pcp/pull/2405

PR #2411, opened 2025-11-17
Libvalkey migration

sfeifer:
With the change from using redis to valkey, we also want to change the underlying libraries. The libvalkey library is the valkey friendly replacement for the redis and hireds-cluster libraries we previously used.

### DIFF
--- a/qa/1543.out
+++ b/qa/1543.out
@@ -507,8 +507,8 @@ Series checked (indom)
     "text-help": "Universally 9 instances numbered 100 .. 900 in steps of 100, and named\u000A\u0022bin-100\u0022 .. \u0022bin-900\u0022",
     "instances": [
         {
-            "instance": 900,
-            "name": "bin-900",
+            "instance": 300,
+            "name": "bin-300",
             "labels": {
                 "domainname": "DOMAINNAME"
                 "hostname": "HOSTNAME"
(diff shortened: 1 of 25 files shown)


natoscott:
With valkey-8.1.5-1.fc43.aarch64 I'm seeing test 1603 as the only concerning one after running all pmproxy & pmseries tests.  Looks like its encountering a valkey-server failure in cluster mode though, which gives a cascading pmseries failure.

```
=== VALKEY BUG REPORT START: Cut & paste starting from here ===
2096478:M 21 Jan 2026 09:36:56.106 # valkey 8.1.5 crashed by signal: 6, si_code: 0
2096478:M 21 Jan 2026 09:36:56.107 # Killed by PID: 2096341, UID: 974
2096478:M 21 Jan 2026 09:36:56.107 # Crashed running the instruction at: 0xffff812192ac

------ STACK TRACE ------
EIP:
/lib64/libc.so.6(+0x992ac) [0xffff812192ac]

2096482 bio_close_file
/lib64/libc.so.6(+0x992ac) [0xffff812192ac]
/lib64/libc.so.6(+0x8bb7c) [0xffff8120bb7c]
/lib64/libc.so.6(+0x8bfd4) [0xffff8120bfd4]
/lib64/libc.so.6(pthread_cond_wait+0x164) [0xffff8120ec68]
valkey-server *:54323 [cluster](bioProcessBackgroundJobs+0x17c) [0xaaaabd28627c]
[...]
```

AIUI "bio" in this context is referring to one of the "background IO" threads.  @sfeifer can you follow up with upstream valkey folk?  I see there is a 9.0.1 now (in rawhide), but 8.1.5 is latest stable version in v8 series.

I'll merge this when I get through a final round of review, but we'll need to follow up on that Valkey issue and also see if the cascading pmseries failure can be handled better.

sfeifer:
Took some time to investigate and create a reproducer, but I just created a new [issue](https://github.com/valkey-io/libvalkey/issues/276) for the libvalkey project. It looks like there's an issue with how the command gets parsed by libvalkey.

PR #2415, opened 2025-11-22
Fix of python_scripted.python and related tests

kurik:
Fix of syntax issues in the `python_scripted.python` script from the testsuite as well as corresponding tests using this script.


kurik:
Pull request from a wrong branch - sorry for this.
Cancelling this PR.

PR #2416, opened 2025-11-22
Fix of `python_scripted.python` and related tests

kurik:
Fix of syntax issues in the `python_scripted.python` script from the testsuite as well as corresponding tests using this script.


kurik:
The debian-11 failed with error message:
``E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?`` so it is an infra issue, not an issue in the code.
Otherwise it seems to be all fine.

@natoscott , @lmchilton  may I ask you please for a review and merge (if all is OK) ?

PR #2420, opened 2025-12-03
pmdadm: dm-multipath fixes

pauljevans:
Update dm-multipath support to better support situations where multipathd does not exist or when multipathd reports no current maps/paths via stderr through it's output.

This fixes issues seen with QA/1188.

### DIFF
--- a/qa/1188
+++ b/qa/1188
@@ -37,6 +37,8 @@ _filter()
     sed \
 	-e "s,$PCP_PMDAS_DIR,PCP_PMDAS_DIR,g" \
 	-e '/Warning: built without support for dmstats metrics/d' \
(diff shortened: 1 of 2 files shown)


pauljevans:
Hopefully this patch also fixes the issue Nathan reported with Coverity Scan, I cannot get the warning to reproduce myself using a local Coverity instance on my end (I do wonder if there's a different version being used with the online scanner).

pauljevans:
Hi Ken,

Your right, it's worth re-adding the pmNotifyErr() messages in the log file on a popen() fail or no found multipath configuration. Will update the PR with these changes

kmcdonell:
Excellent, @pauljevans.
Thanks, merging now.

Issue #2303, opened 2025-08-27
Some metrics not available on macOS/Darwin


tallpsmith:
When running `pmstat` on macOS, several metrics are not available under this OS:

1. `swapd`
	2. `memory`
		1. `buff`
		2. `cache`
	2. `swap`
		1. `in`
		2. `out`
	3. `system`
		1. `in`
		2. `cs`

During a discussion on Slack, @natoscott  mentioned:

> it'd be good to add "Implement process and thread metrics via a new darwin_proc PMDA" - based on linux_proc, and can look into how htop extracts this info

tallpsmith:
I'm closing this as a duplicate/resolved-by #2442  (which won't get all of them, but is getting what we can from this set).

PR #2421, opened 2025-12-04
Allow pmstat to auto-hide unavailable columns

tallpsmith:
At the moment, `pmstat` on macOS is quite lean on what it can show because some metrics are not available (yet).

I've modified `pmstat` so that it (optionally) hides columns that are not available.

I held several sessions with Claude Code to get here because it's certainly better at C than I am...  After some back and forth I narrowed in on the `-C` and `--compact` mode for `pmstat` which seemed to not conflict with any other pm-tool that Claude looked at (and my fuzzy memory).  I elected to make this compact mode optional, and not default, just to maintain backwards compatibility.  There's certainly a 'brain freeze' moment when one runs `pmstat` and sees a lot of columns missing, so I wondered if it was better to maintain the original layout (with missing metrics) as the default for us old folks...

This goes towards #2303

### DIFF
--- a/src/pmstat/pmstat.c
+++ b/src/pmstat/pmstat.c
@@ -113,6 +113,7 @@ pmLongOptions longopts[] = {
     PMOPT_HOSTSFILE,
     PMOPT_LOCALPMDA,
     PMAPI_OPTIONS_HEADER("Reporting options"),
+    { "compact", 0, 'C', 0, "suppress columns for unavailable metrics" },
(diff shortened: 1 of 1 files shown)


tallpsmith:
@natoscott  I don't mind if this is rejected and we leave it as existing behaviour - I initially thought this might be 'simple' and an interim usefulness until the bigger Darwin PMDA upgrade happened (which is nearly in), so if you're not really convinced (and I'm not entirely sure I am either now) then lets close this one.

PR #2422, opened 2025-12-04
pmdanfsclient: fix regex to correctly parse NFS op stats

sourav-sharma796:
The regex used to match NFS operation statistics in /proc/self/mountstats was missing a capture group, causing parsing failures when an additional field was present in newer kernel formats. Updated the regex to include the extra numeric field so that all opstats lines are parsed correctly.

### DIFF
--- a/qa/798.out.64
+++ b/qa/798.out.64
@@ -684,29 +684,29 @@ PMID(s): 62.8.181
 __pmResult ... numpmid: 1
   62.8.181 (<noname>): numval: 10 valfmt: 1 vlist[]:
     inst [N or ???] value 0
-    inst [N or ???] value 3432
-    inst [N or ???] value 3432
-    inst [N or ???] value 3432
-    inst [N or ???] value 3432
-    inst [N or ???] value 3432
-    inst [N or ???] value 3432
-    inst [N or ???] value 3432
-    inst [N or ???] value 3432
-    inst [N or ???] value 3432
+    inst [N or ???] value 0
+    inst [N or ???] value 0
+    inst [N or ???] value 0
+    inst [N or ???] value 0
+    inst [N or ???] value 0
+    inst [N or ???] value 0
+    inst [N or ???] value 0
+    inst [N or ???] value 0
+    inst [N or ???] value 0
 dbpmda> fetch nfsclient.ops.readdirplus.ntrans
 PMID(s): 62.8.182
 __pmResult ... numpmid: 1
   62.8.182 (<noname>): numval: 10 valfmt: 1 vlist[]:
     inst [N or ???] value 0
-    inst [N or ???] value 3432
-    inst [N or ???] value 3432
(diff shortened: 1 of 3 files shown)


natoscott:
@sourav-sharma796 thanks Sourav!  Could you add a new file below qa/nfsclient/ that reproduces the issue as well please?  (this should then automatically get used by qa/798)

sourav-sharma796:
Hi @natoscott , I've added a new file mounstat5.qa. Please take a look.

Thanks!

sourav-sharma796:
Hi @natoscott, can you please take a look?

Thanks!

sourav-sharma796:
Hi @natoscott , @kmcdonell  can anyone please take a look?

Thanks!

sourav-sharma796:
Hi, @kmcdonell, can you please take a look at this one?

Thanks!

kmcdonell:
@sourav-sharma796 OK, thanks for adding the extra test data, but now you need to run qa/798 which will fail because of the additional testing ... then in the qa directory, you'll find a ./remake script, run this as follows:
```./remake 798```
and when you're sure the (additional) output is correct, commit the changes to `798.out` as part of this PR.
I'll merge as soon as that is done.
Cheers, Ken

**Comment by sourav-sharma796** [comment]
Hi @kmcdonell, I've modified qa/798.out.64 output file after running qa/798 script.

I don't have 32 bit system, therefore couldn't update qa/798.out.32.

Please take a look, if this is fine.

Thanks!

**Review by kmcdonell — COMMENTED** [review]
@sourav-sharma796 Are you sure these QA results are right?
All of the values (except the exports and mountdir ones) are 0.
But there are some non-zero values in the data file.
To increase confidence that the PMDA is actually returning expected values, could I please suggest that you edit the qa/nfsclient/mountstats5-rpc-v1.1.qa file and make some of the 0 values there something else, e.g.
```
LOCKU: 1 2 3 4 5 6 7 8 9
```
and then confirm that some (at least) of the matching values appear in the .out file.  Of course doing this for more than just the LOCKU line would increase coverage and confidence.  Note the data does not need to make "sense", we're concerned that if some value N appears in the stats file, the PMDA will correctly pluck N for the corresponding PCP metric.

I'm also concerned about the spaces being stripped from the dbpmda> prompt ... that's not right.

I can fix the 32-bit version of the .out once we have resolved the issues with the 64-bit version.

sourav-sharma796:
Hi @kmcdonell ,

I checked and found that there's already QA file mountstats-4.18.0-105.el8.rpc-v1.1.qa that has 9 values under per-op statistics.

Therefore, I've updated 798.out.64 file after running 798 script and now values are correctly parsed.

In the 798 script only errors column values are fetched, therefore, I've changed those values in the qa file only.

Thanks!

kmcdonell (approved):
Thanks, we're done here.

PR #2425, opened 2025-12-05
Fix macOS launchctl pmcd fork issue

### LINKED ISSUE #2307: Cannot detect `pmcd` crash on macOS - `launchctl` is unable to detect failure and auto-restart
Because of the way `pmcd` starts up, the macOS `launchctl` system is unable to detect that it has crashed and auto-restarts.

Because of the way `pmcd` forks from its start script, If we try to configure launchctl to monitor the process it had launched, it thinks it has crashed and tries to restart it, even though it has successfully started (but in a separate PID).  This causes constant `pmcd` restarts.

Instead, we are currently only able to launch `pmcd` on startup and 'hope for the best' that it stays up.

tallpsmith:
## Summary
Fixes #2307 - pmcd now runs in foreground mode when launched by launchctl on macOS, preventing the parent process exit after fork from being interpreted as a crash.

## Changes
- **build/mac/io.pcp.pmcd.plist**: Added `PMCD_LAUNCHED_BY_LAUNCHD` environment variable and enabled `KeepAlive`
- **src/pmcd/rc_pmcd**: Detect launchctl environment and add `-f` flag to run pmcd in foreground mode

## Benefits
- Enables proper launchctl process monitoring and crash recovery
- Uses existing pmcd foreground mode (`-f` flag)
- Platform-specific solution that doesn't affect other platforms
- Preserves all rc_pmcd setup logic (PMNS rebuild, PMDA setup, etc.)

## Test Plan
See test plan below - validation needed on macOS with launchctl.

### DIFF
--- a/build/mac/qa/TESTING.md
+++ b/build/mac/qa/TESTING.md
@@ -1,0 +1,244 @@
+# macOS launchctl pmcd Test Plan
+
+## Automated Testing (Recommended)
+
+The automated test script handles all timing issues for slow VMs:
+
+```bash
+# Make executable and run
+chmod +x test-pmcd-launchctl.sh
+sudo ./test-pmcd-launchctl.sh
+```
+
+**Note**: The script is configured for slow VMs with 3-minute timeouts. You can adjust the timeout values at the top of the script if needed.
+
+---
+
+## Manual Testing
+
+For manual validation on slow VMs, use these commands with proper wait strategies:
+
+### Prerequisites
+```bash
+# Build and install PCP with the fix
+cd /path/to/pcp-pmcd-daemon-launchctl
+./configure --prefix=/usr --libexecdir=/usr/lib --sysconfdir=/etc --localstatedir=/var
+make
+sudo make install
+```
+
+### Helper Function for Waiting
+
+Add this to your shell session for easier testing:
+
+```bash
+# Wait for pmcd to be ready (use PCP's built-in tool)
+wait_pmcd() {
+    echo "Waiting for pmcd to respond..."
+    if command -v pmcd_wait >/dev/null 2>&1; then
+        # Use PCP's pmcd_wait with 3 minute timeout
+        pmcd_wait -t 180
+    else
+        # Fallback: poll until pmcd responds
+        local count=0
+        while ! pminfo -f pmcd.version >/dev/null 2>&1; do
+            sleep 5
+            count=$((count + 1))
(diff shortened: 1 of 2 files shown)


natoscott (commented):
pmproxy will need exactly the same change, it starts up very similarly to pmcd.

I'd prefer the QA to move below qa/macos for now - in time we need to fix the growing problem of divergent QA suites for mac and everyone else.

tallpsmith:
@natoscott You are correct, but I don't actually think macOS even supports pmproxy working just yet at all (I should raise an issue for that while I remember).

I'll also create an issue regarding the macOS QA setup.  My issue is that hardly any of the existing QA suite works on the Mac, which maybe because it is legitimately broken, but it makes it hard to see the forest for the trees.

kmcdonell:
> My issue is that hardly any of the existing QA suite works on the Mac, which maybe because it is legitimately broken, but it makes it hard to see the forest for the trees.

@tallpsmith were you able to follow my developer workflow notes and if so, can you send me (in email, not here) some examples of tests in the qa dir that are not working, e.g. what does `check -g sanity` report?

tallpsmith:
@kmcdonell I'll definitely run through your developer workflow notes and gather up what I find and post any issues I see. (just want to finalise any of these Darwin PMDA PRs first if I can.

natoscott:
@tallpsmith re "forest for the trees" - much of the initial effort will be in the qa/common* scripts I guess - missing launchd and general macos support in there.  It may be best to focus on just one or two very basic tests (qa/000 and qa/001) - once that common code is working, many other tests will just start working, and the forest will hopefully be much thinner.

tallpsmith:
@natoscott  can we get this one in for the next release?  I'd love to have this in to compliment the Homebrew one (https://github.com/performancecopilot/homebrew-performancecopilot - this is just a website documentation update on how to homebrew install).

natoscott:
@tallpsmith looks good.  Just for when you come back to this with pmproxy, you might want to use a common variable like "PCP_USING_LAUNCHD" for both places, perhaps.  Also at that time, consider making this into a one-liner:
```
+           if [ -n "$PMCD_LAUNCHED_BY_LAUNCHD" ]
+           then
+               OPTS="$OPTS -f"
+           fi
```

as:  [ -n "$PMCD_LAUNCHED_BY_LAUNCHD" ] && OPTS="$OPTS -f"

PR #2427, opened 2025-12-07
Fix Python deprecation warnings during build

### LINKED ISSUE #2306: Python-warning during build
```
        ********************************************************************************
        Please avoid running ``setup.py`` directly.
        Instead, use pypa/build, pypa/installer or other
        standards-based tools.

        See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
        ********************************************************************************

```

tallpsmith:
## Summary

Eliminates Python deprecation warnings that appear during the build process by modernizing the Python packaging approach.

## Changes

- Replace deprecated `setup.py install` with `pip install` in the build system
- Remove deprecated `codecs.open()` import from setup.py
- Add `python3-pip` to QA package manifest for all supported platforms (CI build dependency)

## Why These Changes Are Safe

**codecs.open() removal:**
The `codecs.open()` import was added in 2016 when PCP supported Python 2, where the built-in `open()` function lacked an `encoding` parameter. In Python 3 (the only version PCP now supports), the built-in `open()` natively supports encoding and is the recommended approach. Both functions behave identically for UTF-8 files.

**pip install migration:**
The `setup.py install` command has been deprecated by setuptools in favor of standards-based tools like pip. The new approach uses `pip install` with equivalent flags to achieve the same installation behavior.

**python3-pip in manifest:**
The pip module is now required for the build process but wasn't installed in most CI environments. Added python3-pip entries to `qa/admin/other-packages/manifest` for all package managers (dpkg, rpm, emerge, etc.) to ensure pip is available during builds. Verified availability on all supported platforms: Ubuntu 18.04+, Debian 11+, CentOS Stream 8+, Fedora 42+, and others.

## Testing

- ✅ Build completes successfully with no deprecation warnings
- ✅ All Python extensions compile correctly
- ✅ No functional changes to the installation process

Fixes #2306

### DIFF
--- a/src/python/GNUmakefile
+++ b/src/python/GNUmakefile
@@ -65,7 +65,7 @@ build_python3: $(SETUP_PY) $(CFILES)
 	touch build_python3
 
 install_python3: build_python3
-	export $(ENV); $(PYTHON3) -m pip install --no-build-isolation --root="$${DIST_ROOT:-/}" --prefix=$(PYTHON_PREFIX) .
+	export $(ENV); $(PYTHON3) -m pip install --no-build-isolation --root="$${DIST_ROOT:-/}" --prefix=$(PYTHON_PREFIX) --record=$(TOPDIR)/python3-pcp.list .
 	export $(ENV); $(PYTHON3_INSTALL)
 else
 build_python3:


tallpsmith:
ok, this is working on the Mac because we include python3-pip as a dependency. To fix this we'd need to add that as a dependency to all operating systems.  This probably needs discussion @natoscott .

kmcdonell:
@tallpsmith No --record for `/usr/bin/python3 -m pip install` on Ubuntu.

tallpsmith:
Yeah @kmcdonell  this is looking hairier than Hairy Maclary right now as I dig in.  I also have to solve the "how do I run the CI podman setup locally" so that I can easily iterate locally, however I ran into a container architecture mismatch with Apple Silicon that might be another rabbit hole.

I may end up resorting to an Option Claude Code suggested earlier - simply suppress the warnings! :)

natoscott:
@tallpsmith @kmcdonell see related changes in https://github.com/performancecopilot/pcp/pull/2412 - also not yet complete, but very likely we have alot of overlap ... hopefully orthogonal parts of the problem are solved in each.

tallpsmith:
> @tallpsmith @kmcdonell see related changes in #2412 - also not yet complete, but very likely we have alot of overlap ... hopefully orthogonal parts of the problem are solved in each.

Ok, that's good news, maybe your work will supersede this one (I kinda gave up to focus on some other things anyway).  Happy to withdraw this (I'll mark it as draft anyway).

PR #2428, opened 2025-12-08
Add AmazonLinux 2023 to CI Pipleline

kdvalin:
# Summary
Adds Amazon Linux 2023 (both x86_64 and aarch64) to the CI pipeline and eventual publishing to packagecloud.

# Changes
`ci.yml`: Add support for different arch's, since previously crun downloads were hardcoded to `amd64` arch.
`whatami`: Detect `/etc/amazon-release` before defaulting to `/etc/os-release`

### DIFF
--- a/build/ci/platforms/amazonlinux2023-aarch64-container.yml
+++ b/build/ci/platforms/amazonlinux2023-aarch64-container.yml
@@ -2,7 +2,7 @@ type: container
 
 artifactory:
   package_type: rpm
-  deploy_path: /amazon/2023/x86_64/pcp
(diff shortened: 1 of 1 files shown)


kdvalin:
Thanks for the feedback!  I can separate out the ci.yml changes and move it to something like an experimental QA pipeline if desired.  If it helps reduce some hesitations, AmazonLinux is essentially a stripped down Fedora (their repos don't include things like libvirt).

Ultimately what I'm looking for is the ability to install pcp rpms on AmazonLinux from PackageCloud, since my team is looking to utilize PCP to help standardize metric collection across multiple Distros (AmazonLinux being one of them).

Issue #2432, opened 2025-12-10
Injecting functions for Solaris 10 build


SeanBurfordT:
I'm attempting to build PCP on Solaris 10.

Yes, I realize that this is potentially a complicated endeavor.

I have promising progress, however would like to add a local implementation of strcasestr().  Normally I would add a compat_strcasestr.c and compat_strcasestr.h file with a weak symbol for strcasestr(), however it's not obvious how to do this in the PCP build system.

```
bash ./Makepkgs CFLAGS='' CXXFLAGS='-std=gnu++11' AR='/usr/sfw/sparc-sun-solaris2.10/bin/ar'

== Cleaning build tree for packaging build
Wrote: /export/home/shared/src/pcp/build/tar/pcp-7.1.0.src.tar.gz

== Preparing fresh build tree for packaging build

== Configuring pcp, log is in /export/home/shared/src/pcp/Logs/pcp  (--sysconfdir=/etc --localstatedir=/var --prefix=/usr --libdir=/usr/lib/64  CFLAGS= CXXFLAGS=-std=gnu++11 AR=/usr/sfw/sparc-sun-solaris2.10/bin/ar)

== Building pcp, log is in /export/home/shared/src/pcp/Logs/pcp
Make default_pcp failed, see log in /export/home/shared/src/pcp/Logs/pcp
/opt/csw/bin/gcc -D_POSIX_PTHREAD_SEMANTICS -I../../src/include -I../../src/include/pcp -DPCP_VERSION=\"7.1.0\" -fPIC -fno-strict-aliasing -D_XPG4_2 -D__EXTENSIONS__ -m64 -Wshadow -Wno-array-bounds  -Wall -O2 -g     -c -o pmnsutil.o pmnsutil.c
/opt/csw/bin/gcc -D_POSIX_PTHREAD_SEMANTICS -I../../src/include -I../../src/include/pcp -DPCP_VERSION=\"7.1.0\" -fPIC -fno-strict-aliasing -D_XPG4_2 -D__EXTENSIONS__ -m64 -Wshadow -Wno-array-bounds  -Wall -O2 -g   -o pmnsmerge -m64 -fPIC -Wall -L../../src/libpcp/src  pmnsmerge.o pmnsutil.o -lpcp -lnsl -lsocket -lresolv -ldl -lposix4
Undefined                       first referenced
 symbol                             in file
strcasestr                          ../../src/libpcp/src/libpcp.so
[...]

PR #2435, opened 2025-12-16
Solaris 10 compatibility 1/2

SeanBurford:
3 changes for https://github.com/performancecopilot/pcp/issues/2432

1. Add `strcasestr` to configure and libpcp.  I have copied the implementation from libpcp win32.c.  Solaris 10 libc does not have strcasestr.
2. Move `cfmakeraw` implementation from roomtemp into libpcp.  Solaris 10 libc does not have cfmakeraw.
3. Rename `printdelta` to `printtimedelta` to avoid a conflict with a macro in the Solaris 10 system headers.

Please check:
1. Please double check that my deletion of the strcasestr() declaration from the MINGW32 section of platform_defs.h.in is appropriate.  My assumption is that the MINGW32 build uses configure (therefore will pick up the later declaration guarded by HAVE_STRCASESTR).
2. This change moves GPL code (cfmakeraw) from the periphery into libpcp and adds an export to libpcp on Solaris.

### DIFF
--- a/src/libpcp/src/exports.in
+++ b/src/libpcp/src/exports.in
@@ -719,10 +719,11 @@ PCP_4.1 {
 
 PCP_4.2 {
     global:
-#if ! HAVE_STRCASESTR
-    strcasestr;
-#endif
+    pmUnregisterDerived;
 #if ! HAVE_CFMAKERAW
     cfmakeraw;
+#endif
+#if ! HAVE_STRCASESTR
+    strcasestr;
 #endif
 } PCP_4.1;


SeanBurford:
I performed the libpcp3 `printdelta` change because it the symbol conflict broke the Solaris 10 build.  The bug fix was just one line, so I included that too.

I've attempted a MINGW32 build, but don't have the toolchain set up for it (in particular your build system seems to have /usr/bin/mingw64-... soft links to make, configure, qmake, pkg-config), so I'm blind regarding the effect this change has on the Windows build.  CI/CD doesn't appear to have a Windows target either.

I've applied your suggestions from your previous comment and have a build of my own running as well.

SeanBurford:
The exports.in change caused a build failure.  An exports section that has an empty global section is not syntactically valid:

```
PCP_4.x {
    global:
}
```

I had a look at mk.exports and decided to (at least temporarily) add a non-conditional entry to the version to work around that issue.

PR #2436, opened 2025-12-16
pcp-ps: Added -O/--sort option to do sorting on basic of %mem and %cpu

orasagar:


kurik:
Hi @orasagar ,
I have not walk through the full code, however just compiling these code changes and running the `pcp ps` gives me the following error:
```
# pcp ps -O '%mem,%cpu'
Invalid sorting option, defaulting to %cpu
Linux  6.12.0-167.el10.x86_64  (vm-10-0-184-136)  01/05/26  x86_64    (4 CPU)
Sorting indices not set for selected flag, please remove sorting flag or choose another flag for output
Sorting indices not set for selected flag, please remove sorting flag or choose another flag for output
```

The traditional `ps` works fine with this option:
```
# ps -O '%mem,%cpu'
    PID %MEM %CPU S TTY          TIME COMMAND
1359809  0.0  0.0 S pts/0    00:00:00 /bin/bash --noprofile --rcfile /tmp/tmp.lQ1StVWzDo
2026683  0.0  0.0 R pts/0    00:00:00 ps -O %mem,%cpu
```

Am I using the `pcp ps` in a wrong way ?

sagarramdev173:
Hi @kurik ,
By default, **pcp ps** doesn't have %mem or %cpu  column that's you are seeing this "Sorting indices not set for selected flag, please remove sorting flag or choose another flag for output"

if you like to see the output sorted then you need to have column in the output to be sorted.

you can use 
pcp ps -u -O "%mem" or 
pcp ps -u -O "%cpu"

orasagar:
Hi @kurik @natoscott @kmcdonell ,

I have updated the option of -O into -S/--sort option to make it more intuitive .
Since -O option was already being used by pcp for origin .

natoscott:
> Hi @kurik @natoscott @kmcdonell ,
> 
> I have updated the option of -O into -S/--sort option to make it more intuitive . Since -O option was already being used by pcp for origin .

Hmm, -S is also being used already in PCP (--start time).  I recommend sticking with the original ps options, and PCP options like -O and -S can still be passed via "pcp -O@2am ps ..." if need be.  The pcp-ps code could also explicitly handle any PCP long options as a convenience, assuming they don't clash like the short options).

orasagar:
@natoscott I have done the code changes to be in line with the ps tool

PR #2441, opened 2025-12-27
Add macOS Tart VM build documentation

tallpsmith:
Adds documentation for building PCP on macOS using Tart VMs via Cirrus CLI.

## Rationale

- macOS cannot be containerized (no Podman support like Linux)
- Tart VMs provide lightweight, isolated builds using native macOS virtualization
- Eliminates environment differences between developers
- Matches CI environment as closely as possible

PR #2431, opened 2025-12-10
Fix local CI: Dynamic task lists and aarch64 support

tallpsmith:
## Summary

This PR fixes several issues identified during local CI testing on macOS ARM64:

- Support dynamic task lists in reproduce command (enables --until qa_sanity)
- Add Ubuntu 24.04 'any' architecture package list for aarch64 detection  
- Fix container .git detection (replace worktree pointer with directory for Makepkgs)
- Add task validation with helpful error messages

## Changes

- build/ci/ci-run.py:
  - Dynamic task discovery from platform YAML instead of hardcoded list
  - Proper validation of --until parameter
  - Container setup ensures .git directory exists for Makepkgs detection

- qa/admin/package-lists/Ubuntu+24.04+any:
  - New package list for 'any' architecture variant of Ubuntu 24.04
  - Enables aarch64 builds on ARM64 macOS

## Testing

Local testing on macOS M4 Pro ARM64 with --quick mode and --until qa_sanity parameter.

Full QA test run validation in progress to confirm end-to-end functionality.

PR #2442, opened 2026-01-02
darwin PMDA: Comprehensive metrics expansion

tallpsmith:
## Overview

Comprehensive expansion of darwin PMDA monitoring capabilities, adding 60+ new metrics across memory compression, VFS resources, network protocols, and process statistics.

### Memory (5 metrics)
| Metric | Description |
|--------|-------------|
| `mem.util.compressed` | Compressed memory size (KB) |
| `mem.compressions` | Compression operations count |
| `mem.decompressions` | Decompression operations count |
| `mem.compressor.pages` | Pages held by compressor |
| `mem.compressor.uncompressed_pages` | Uncompressed pages in compressor |

### VFS Resources (7 metrics)
| Metric | Description |
|--------|-------------|
| `vfs.files.count` | Open file descriptors |
| `vfs.files.max` | Maximum file descriptors |
| `vfs.files.free` | Available file descriptors |
| `vfs.vnodes.count` | Active vnodes |
| `vfs.vnodes.max` | Maximum vnodes |
| `kernel.all.nprocs` | Total process count |
| `kernel.all.nthreads` | Total thread count |

### Network - UDP (5 metrics)
| Metric | Description |
|--------|-------------|
| `network.udp.indatagrams` | UDP datagrams received |
| `network.udp.outdatagrams` | UDP datagrams sent |
| `network.udp.noports` | Datagrams to closed ports |
| `network.udp.inerrors` | Input errors |
| `network.udp.rcvbuferrors` | Receive buffer full errors |

### Network - ICMP (8 metrics)
| Metric | Description |
|--------|-------------|
| `network.icmp.inmsgs` | ICMP messages received |
| `network.icmp.outmsgs` | ICMP messages sent |
| `network.icmp.inerrors` | Input errors |
| `network.icmp.indestunreachs` | Destination unreachable received |
| `network.icmp.inechos` | Echo requests received |
| `network.icmp.inechoreps` | Echo replies received |
| `network.icmp.outechos` | Echo requests sent |
| `network.icmp.outechoreps` | Echo replies sent |

### Network - Socket Statistics (2 metrics)
| Metric | Description |
|--------|-------------|
| `network.sockstat.tcp.inuse` | TCP sockets in use |
| `network.sockstat.udp.inuse` | UDP sockets in use |

### Network - TCP Connection States (11 metrics)
| Metric | Description |
|--------|-------------|
| `network.tcpconn.established` | Established connections |
| `network.tcpconn.syn_sent` | SYN sent (active open) |
| `network.tcpconn.syn_recv` | SYN received (passive open) |
| `network.tcpconn.fin_wait1` | FIN_WAIT_1 state |
| `network.tcpconn.fin_wait2` | FIN_WAIT_2 state |
| `network.tcpconn.time_wait` | TIME_WAIT state |
| `network.tcpconn.close` | CLOSED state |
| `network.tcpconn.close_wait` | CLOSE_WAIT state |
| `network.tcpconn.last_ack` | LAST_ACK state |
| `network.tcpconn.listen` | LISTEN state |
| `network.tcpconn.closing` | CLOSING state |

### Network - TCP Protocol Statistics (15 metrics)
| Metric | Description |
|--------|-------------|
| `network.tcp.activeopens` | Active connection attempts |
| `network.tcp.passiveopens` | Passive connection accepts |
| `network.tcp.attemptfails` | Failed connection attempts |
| `network.tcp.estabresets` | Established connections reset |
| `network.tcp.currestab` | Currently established |
| `network.tcp.insegs` | Segments received |
| `network.tcp.outsegs` | Segments sent |
| `network.tcp.retranssegs` | Segments retransmitted |
| `network.tcp.inerrs` | Input errors |
| `network.tcp.outrsts` | RST segments sent |
| `network.tcp.incsumerrors` | Checksum errors |
| `network.tcp.rtoalgorithm` | RTO algorithm (constant) |
| `network.tcp.rtomin` | Minimum RTO (ms) |
| `network.tcp.rtomax` | Maximum RTO (ms) |
| `network.tcp.maxconn` | Max connections (constant) |

**Note:** TCP statistics require `net.inet.tcp.disable_access_to_stats=0` (documented in new `pmdadarwin(1)` man page).

### Process Metrics (3 metrics)
| Metric | Description |
|--------|-------------|
| `proc.io.read_bytes` | Bytes read from disk |
| `proc.io.write_bytes` | Bytes written to disk |
| `proc.fd.count` | Open file descriptor count |

## pmrep Monitoring Views

Six new/enhanced pmrep configurations for comprehensive macOS monitoring:

- **`:macstat`** - Enhanced overview with aggregated network bandwidth
- **`:macstat-x`** - Extended view with VFS and disk byte metrics
- **`:macstat-mem`** - Memory deep-dive (compression, paging, swap)
- **`:macstat-dsk`** - Disk I/O analysis (IOPS, throughput, latency)
- **`:macstat-tcp`** - TCP connection lifecycle and health
- **`:macstat-proto`** - Network protocol overview (UDP, ICMP, TCP summary)

These views provide out-of-the-box monitoring without custom configuration.

## Testing Infrastructure

- **Unit tests:** dbpmda-based tests for all new metrics (`scripts/darwin/test/unit/`)
- **Integration tests:** PCP tool validation (`scripts/darwin/test/integration/`)
- **CI enhancement:** macOS GitHub Actions workflow now runs darwin-specific tests post-installation

## Claude Code Artifacts (For Discussion)

This PR includes AI development tooling for the first time in the PCP codebase:

### Agents (`src/claude-code/agents/`)
- **`macos-darwin-pmda-qa`** - Automated QA testing agent that builds darwin PMDA in isolation and runs full unit/integration test suites
- **`pcp-code-reviewer`** - Code review agent that validates PCP coding standards, style consistency, and darwin PMDA architectural patterns

### Skills (`src/claude-code/skills/`)
- **`macos-qa-test`** - Skill for running darwin PMDA tests in isolated Cirrus VM environment

These tools automate repetitive QA tasks and enforce code quality standards. Maintainers should review whether these should be retained, relocated, or removed from the main codebase.

## Documentation Files

Several planning and development documents are included for transparency:
- `DARWIN-PMDA-ENHANCEMENT-PLAN.md` - Complete development plan and implementation patterns
- `PMDA_REFACTOR.md` - Code organization and refactoring notes
- Various `.plan.md` files - Phase-specific planning documents

**Note:** These documentation files can be removed before merge if desired, after maintainer review.

## Code Organization

All enhancements follow modular architecture with dedicated subsystem files (e.g., `vfs.c/h`, `udp.c/h`, `tcp.c/h`), keeping `pmda.c` focused on coordination rather than implementation.

### DIFF
--- a/dev/darwin/README.md
+++ b/dev/darwin/README.md
@@ -1,0 +1,29 @@
+# Darwin PMDA Dev tools
+
+This directory contains some build helper functions targetted at the Darwin(macOS) PMDA developer.  It helps
+iteration cycles by focusing on building/testing the `src/pmdas/darwin*` build, but 'cheats' a little by crafting
+a customized GNUMakefile to make it happen.
+
+## Prerequisites
+
+1. Run `${PROJECT_ROOT}/Makepkgs` - this ensures you have a complete and valid build tree to work off (but Darwin PMDA development still happens in src/pmdas/darwin*)
+2. Initialize, one-time, the cheat-code GNUMakefile/builddefs:
+```
+./setup-local-pcp.sh
+```
+
+## Quick Builds
+
+### Compile and run basic Darwin PMDA smoke tests
+```
+./build-quick.sh
+```
+
+### Build and Run simple PMDA tests
+```
+./dev-test.sh
+```
+
+
+# See Also
+* [macOS-specific Build stuff](../../build/mac/README.md)


kmcdonell:
@tallpsmith is the note on archive compatibility because the metadata for an existing metric changed (e.g. 32-bit to 64-bit)?
If so, it is not really a serious problem, as we already have a mechanism for rewriting archives before merging (in pmlogger_daily specifically) thanks to pmlogrewrite(1).
For some simple examples, see src/pmdas/openbsd: the *.conf files are the rewriting specs and the associated GNUmakefile rules install these files where pmlogrewrite(1) will find 'em.

tallpsmith:
> @tallpsmith is the note on archive compatibility because the metadata for an existing metric changed (e.g. 32-bit to 64-bit)? If so, it is not really a serious problem, as we already have a mechanism for rewriting archives before merging (in pmlogger_daily specifically) thanks to pmlogrewrite(1). For some simple examples, see src/pmdas/openbsd: the *.conf files are the rewriting specs and the associated GNUmakefile rules install these files where pmlogrewrite(1) will find 'em.

ah yes I had forgotten about that support from many moons ago, I'll have the plan file updated - thanks.

tallpsmith:
@kmcdonell I asked Claude to review your comment, then he(she? it?) and I had a back and forth to discuss.  Here's our combined response:

----
Claude:

Thank you for the thorough review and approval, Ken. Your feedback raises several valuable points. I'd like to address each one:

  1. Scripts Directory & Cross-Platform Potential

  "I'm wondering if the scripts/darwin testing infrastructure could be generalized for other platforms. For example, I was able to adapt run-integration-tests.sh to Ubuntu with only a few changes (mainly around the platform-specific PMDA installation paths and pmcd service start commands)."

  You raise an excellent point. The test infrastructure is largely platform-agnostic—the core logic uses standard PCP tools (pminfo, pmval, pmstat) that behave consistently across platforms. The main platform-specific code is:

  - PMDA installation path resolution (lines 83-84 in the script)
  - PMCD startup mechanism (Linux uses systemctl/service, macOS uses launchctl)
  - PMDA discovery in pmcd status

  Suggestion: Would it be valuable to generalize this into a shared test framework? We could:
  1. Move the core test functions to scripts/common/ or similar location
  2. Create platform-specific driver scripts that handle the 3-4 integration points
  3. This would benefit all platform PMDAs going forward

  As the contributor (psmith) noted in our discussion: "I do think it's possible we could consider making the unit/integration tests we've developed more general purpose... Your points about the differences between the Linux and macOS installation are still valid though, so it might actually make things trickier."

  Proposed approach: We'll add this to the enhancement plan as follow-up work for a future PR, allowing time to properly evaluate the cross-platform considerations.

  ---
  2. PCP QA System on macOS

  "I'm curious about why the existing overarching full PCP QA system is currently difficult/impossible to run on macOS, particularly given its successful implementation on *BSD and OpenIndiana."

  Fair question. There are several barriers, including some fundamental architecture naming issues:

  Architecture Naming Mismatch:
  The QA system currently relies on architecture identifiers that differ between platforms. Specifically, macOS uses arm64 while Linux uses aarch64 to describe the same 64-bit ARM architecture. This mismatch prevents the QA system from locating appropriate package lists when running on macOS ARM64 hardware.

  As psmith noted: "They currently rely on the architecture (e.g. aarch64 versus the macOS arm64) in the containers which cause problems locally. We actually started a draft PR with trying to address this change (see #2431) but it's proven a tricky problem and we haven't got back to looking in to it (focused on the Darwin PMDA)."

  PR #2431 Context:
  We started addressing this in PR #2431 ("Fix local CI: Dynamic task lists and aarch64 support"), which attempted to resolve:
  - Architecture naming discrepancies (aarch64 vs arm64)
  - Dynamic task discovery for the reproduce command
  - Package resolution for Ubuntu 24.04 on ARM64 systems
  - Git repository detection in container builds

  However, this proved more complex than initially expected and remains a work-in-progress.

  Other Barriers:
  - The QA system assumes Linux-centric paths and utilities
  - macOS deployment differs significantly (/usr/local/lib/pcp vs system defaults)
  - Some platform-specific QA tests reference Linux kernel interfaces

  Proposed path forward: The macOS-specific testing framework (scripts/darwin/test/) was created as a pragmatic solution to enable immediate progress. However, you're right that a proper macOS/Darwin port of the full QA system should be feasible given *BSD's success. We propose:
  1. Completing the work started in PR #2431 (architecture naming)
  2. Adding platform detection to QA scripts
  3. Creating Darwin-specific test groups (similar to how Linux-specific tests exist)
  4. Documenting macOS deployment paths and considerations

  Would this be a valuable long-term improvement? We can track it as follow-up work.

  ---
  3. Copyright Attribution

  "[Recommend] updating copyright assignments in PMDA source files, particularly those modified or created during this work."

  You're right. Currently, new files have generic "Red Hat 2026" copyright. Should the contributor(s) be explicitly attributed?

  Current approach: All new files (tcp.c, tcp.h, udp.c, udp.h, icmp.c, icmp.h, vfs.c, vfs.h, sockstat.c, sockstat.h, tcpconn.c, tcpconn.h) have:
   * Copyright (c) 2026 Red Hat.

  Refactored files (modified significantly but not created) like cpu.c, cpuload.c, loadavg.c, etc., have:
   * Copyright (c) 2026 Red Hat.
   * Copyright (c) 2004 Silicon Graphics, Inc.  All Rights Reserved.

  Question: What's the PCP project's preferred approach? Should we:
  1. Add individual contributor names to the 2026 line?
  2. Keep corporate attribution and handle contributor tracking via git history?
  3. Add a separate "Contributors:" comment block?

  ---
  4. Code Organization Assessment ✓

  Appreciated the confirmation that the refactoring approach (moving code between files and whitespace changes) was valid. This validates the modular architecture we've adopted for future enhancements.

  ---
  5. TCP/UDP Statistics Coverage: Granular vs. Aggregate Metrics

  "Regarding `tcp.c`, I'm wondering whether individual summed metrics warrant export. For example, tcps_rcvbadsum and tcps_rcvbadoff individually provide diagnostic value beyond their aggregate (inerrs)."

  This is the most substantive technical question, and it deserves careful consideration. Currently:

  TCP inerrs (item 176) is computed as:
  `tcps_rcvbadsum + tcps_rcvbadoff + tcps_rcvshort + tcps_rcvmemdrop`

  UDP inerrors (item 145) is computed as:
  `hdrops + badsum + badlen`

  Your point is valid: Different error types have different diagnostic implications:
  - `tcps_rcvbadsum` (bad checksums) → NIC issues, transmission line noise
  - `tcps_rcvbadoff` (bad offset) → malformed packets, potential attack patterns
  - `tcps_rcvshort` (truncated packets) → MTU mismatches, packet loss
  - `tcps_rcvmemdrop` (buffer exhaustion) → system resource pressure

  Options:

  Option A (Current): Export only the aggregate inerrs + individual byte/segment counts (insegs, outsegs)
  - ✓ Simpler PMNS, fewer metrics
  - ✗ Loses diagnostic granularity for troubleshooting

  Option B (Your suggestion): Export both aggregate AND individual error types
  - ✓ Diagnostic granularity for deep troubleshooting
  - ✓ Follows Linux PMDA precedent (need to verify)
  - ✗ Metric proliferation (adds ~4 TCP error metrics + ~3 UDP error metrics)

  Option C (Hybrid): Export individual errors but document as "expert-level" metrics
  - ✓ Diagnostic value available
  - ✓ Clear user guidance on when to use them
  - ✗ Still increases metric count

  My recommendation: I lean toward Option B for the following reasons:

  1. Diagnostic value: Network operators troubleshooting performance issues need this detail
  2. Archive compatibility: Once exported, metrics become part of the archive interface—better to export comprehensively than regret removing them later
  3. PCP philosophy: Provide what the kernel provides; let tools aggregate if needed
  4. Precedent: Need to verify what Linux PMDA does for consistency

  Question: Does Option B (export both aggregate and granular) align with PCP philosophy? Should we check Linux PMDA for consistency?

  ---
  6. TCP Connection State Validation

  "I'd suggest adding pmDebug-guarded diagnostics if TCP state values fall outside expected ranges. Historically, kernel enum values can advance without PMDA updates."

  Excellent defensive programming suggestion. This is particularly relevant because:

  - macOS kernel can be updated independently of PCP
  - TCP connection states have a fixed set (TCPS_* enums in <netinet/tcp_fsm.h>)
  - Future macOS versions might add new states

  Proposed implementation:

  In `tcpconn.c's` `refresh_tcpconn()`, after parsing the PCB list:
```
  /* Validate state is within expected range (TCPS_CLOSED=0 to TCPS_CLOSING=11) */
  if (so.so_type == SOCK_STREAM) {
      if (xt.xt_tp.t_state > TCPS_CLOSING) {
          pmNotifyErr(LOG_WARNING,
              "tcpconn: unexpected TCP state %d (expected 0-%d), treating as CLOSING",
              xt.xt_tp.t_state, TCPS_CLOSING);
          tcp->state[TCPS_CLOSING]++;  /* Fail-safe: count as CLOSING */
      } else {
          tcp->state[xt.xt_tp.t_state]++;
      }
  }
```
  Alternatively, with pmDebug for lower-priority logging:
```
  if (so.so_type == SOCK_STREAM) {
      if (xt.xt_tp.t_state > TCPS_CLOSING) {
          pmDebug(DBG_TRACE_APPL0,
              "tcpconn: unexpected TCP state %d (max %d)",
              xt.xt_tp.t_state, TCPS_CLOSING);
          tcp->state[TCPS_CLOSING]++;  /* Still count it */
      } else {
          tcp->state[xt.xt_tp.t_state]++;
      }
  }
```
  Question: Which approach is preferred—pmNotifyErr (always logged as warning) or pmDebug (only when tracing enabled)?

  ---
  7. UDP Error Reporting

  "Similar concerns for UDP—should we export the granular hdrops, badsum, badlen metrics?"

  Same reasoning as TCP above. Currently we export:
  - indatagrams, outdatagrams (counts)
  - noports (diagnostic—datagrams to closed port)
  - inerrors (aggregate: hdrops + badsum + badlen)
  - rcvbuferrors (fullsock—socket buffer exhaustion)

  The three components of inerrors represent different issues:
  - hdrops → header parsing failures
  - badsum → checksum errors (hardware/transmission issues)
  - badlen → length validation failures

  Proposed solution: Same as TCP—export both aggregate and individual metrics if you agree with Option B above. This would add 3 new metrics to the PMNS.

  ---
  Summary: Changes Needed (If Approved)

  If you agree with the approaches above, we'll update the enhancement plan and make these changes:

  1. ✓ Copyright: Clarify PCP attribution preferences (awaiting guidance)
  2. ✓ TCP/UDP metrics: Export granular error metrics in addition to aggregates (Option B)
  3. ✓ TCP state validation: Add defensive diagnostics for out-of-range states
  4. ✓ Test infrastructure generalization: Add to enhancement plan as follow-up work (Phase 5)
  5. ✓ QA system port: Add to enhancement plan as follow-up work, coordinate with PR #2431

  ---
  Questions for You

  1. Metrics granularity: Does Option B (export both aggregate and granular) align with PCP philosophy?
  2. Copyright approach: What's the project standard?
  3. State validation: pmNotifyErr (always warn) or pmDebug (trace-only)?
  4. Linux PMDA consistency: Should we verify Linux PMDA's approach to TCP/UDP error metrics before implementing?
  
  ----

Paul: From my perspective, I can get Claude to fix up the Copyright easy enough (ideally we would have something in the build that does this automatically? One day?).  But if you can review Claude's questions and options on how to handle the specific cases, I can make them happen once we've agreed.  

On the QA side, my main concern was having as fast a feedback loop as possible, and that QA cycle is just HORRIBLELY slow?  I needed something that fast that Claude could run and get feedback from to iterate on.  A 5 minute Tart VM build running the macOS-specific unit/integration tests was a fantastic win for me (us?)

thanks for looking over it though!

kmcdonell:
This is getting a bit out of control ... the "reply" to my review runs to almost 1500 words.

So I'll try to be brief.

- because the `scripts` dir is used for other stuff, suggest moving `scripts/darwin` to `dev` (new at the top level) and leave generalization for a later iteration, otherwise Proposed approach is fine
- you were referring to the CI and QA actions in github, I was referring to the `qa` directory ( and the `check` et al scripts there) in the source tree ... the former is problematic, the latter should be manageable, e.g. paths should all be abstracted to `/etc/pcp.conf` (or similar) so scripts in the `qa` directory will work across platforms, and you don't need additional qa groups as the existing -g and -x mechanisms for check can be used to cherry pick the groups that are expected to work for a specific platform, see `qa/admin/pcp-daily` for the way this is done in the QA Farm (which is different to the github QA action)
- copyrights need to belong to legal entities who are able to assign the copyright for their contributions ... so people (not Claude) or where employment contracts dictate, an employee's company ... and this only applies to significant contributions, so not cosmetic changes, not one-line fixes, but contributions that add some substantive functionality (unfortunately the law is really vague in this area and "significant" and "substantive functionality" can only be determined by a judge in a court case, so some common sense is required) ... so I'd expect to see some Copyright (c) 2026 Paul Smith here.  The "All Rights Reserved" is optional I believe and more likely to be requested by specific corporate lawyers or old-farts-force-of-habit.
- Add individual contributor names to the 2026 line? yes
- Keep corporate attribution and handle contributor tracking via git history? no, Copyright annotation needs to be in the source files
- Add a separate "Contributors:" comment block?, no (well optional, but the Copyright (c) annotation is also required)
- TCP/UDP Statistics Coverage: Granular vs. Aggregate Metrics ... go with Option B (metric proliferation is a non-issue), **but** I'd recomment grouping the metrics in the PMNS so the related metrics are below a common non-leaf node in the PMNS, e.g. network.tcp.inerrs.total, network.tcp.inerrs.badsum, network.tcp.inerrs.badoff, network.tcp.inerrs.short and network.tcp.inerrs.memdrop
- ditto for UDP Error Reporting
- TCP Connection State Validation, the pmNotifyErr() option is good (the pmDebug() one is totally wrong), **but** add one-trip guard so the log is not filled with these messages each time a tcpconn metric is fetched **and** drop the aggregation into `tcp->state[TCPS_CLOSING]` because this just pollutes that metric with an unknown number of other observations (if it is an unknown type, the one-trip warning will suffice).

and the Questions for You (me) [not already addressed above]
- Metrics granularity: Does Option B (export both aggregate and granular) align with PCP philosophy? yes, modulo the suggested PMS names, e.g. refer to disk.dev.read, disk.dev.write and disk.dev.total (3 of the oldest metrics in PCP!)
- Copyright approach: What's the project standard? there isn't one as per the discussion above
- Linux PMDA consistency: Should we verify Linux PMDA's approach to TCP/UDP error metrics before implementing? not really, software engineering in the network drivers and protocols area is a notorious refuge for anarchists, so one should expect that the instrumentation and exported metrics will **not** be the same across platforms
- [Paul] On the QA side, my main concern was having as fast a feedback loop as possible, and that QA cycle is just HORRIBLELY slow?  [Ken] again I think this relates to the github CI and QA actions ... when developing I never depend on them for testing, I'm always building and packaging locally, then overlaying with targeted `sudo make install` and then running very narrow sets of tests from the `qa` directory ... this is very fast (for me).

tallpsmith:
@kmcdonell (it's the real Paul here) I believe all your feedback is now in place, I'd probably ask if you can review the TCP/UDP changes, with the guard check as the most important (I did look, but my eyes are not as seasoned as yours).

I've taken on looking at how to move/integrate some/all of my unit/integrations tests into the existing `qa` area.  Given the exotic nature of macOS, I'm not yet sure how this works (even in the short term), there's a number of failures of existing tests running under macOS at the moment, but I might see if I can gather some more details on the current status and craft an Issue to reflect that.

tallpsmith:
> ...This PR is getting sorta long ...

@kmcdonell  to quote something @thecowan said many moons ago, I think it was "That's like saying Jack the Ripper was a bit of a lad".  Understatement.

I agree.  I would have liked @natoscott  to have had his eyes on it, or indeed anyone else with macOS interest, but if we're happy with the principles of "If you break it you fix it", I'm up for continuing with any finer touches and further future feedback from others later.

Depends if there's any release planned any time soon?  Maybe only @natoscott knows?

tallpsmith:
Also, I'd plan to squash it as a single commit merge (just for the sanity of the git history)?

natoscott:
@tallpsmith I'll take a look today.

> Also, I'd plan to squash it as a single commit merge (just for the sanity of the git history)?

IMO, its OK to have multiple independent commits introducing distinct functionality ... so either one commit if you prefer, or multiple commits with sensible boundaries (if we have lots of one-liner train-of-thought commits, ideally squash those).

tallpsmith:
@natoscott dis try and commit in logical chucks (each page) but there might some follow up bug fixes and tidy ups in there. It's just a lot of change and therefore quite a lot of commits. 

If you don't mind all of them coming across I like the history too but your call.

.claude/agents/macos-darwin-pmda-qa.md
```diff
@@ -0,0 +1,92 @@
...
+tools: Bash, Glob, Grep, Read, WebFetch, TodoWrite, WebSearch, Skill
+model: haiku
+color: red
+---
+
+You are a macOS Darwin PMDA QA Automation Agent. Your sole responsibility is to execute the isolated build and test environment for Performance Co-Pilot's darwin PMDA and report results.
+
+## Operational Constraints
+
+- You ONLY operate when code changes occur in the @src/pmdas/darwin directory
```
natoscott: what about src/pmdas/darwin_proc?

.claude/agents/macos-darwin-pmda-qa.md
```diff
@@ -0,0 +1,92 @@
...
+- You ONLY execute on macOS host systems - verify this before proceeding
+- You execute tests; you do NOT create, modify, or debug tests
+- You do NOT analyze failures beyond reporting them - pass failure details to other agents for analysis
+
+## Execution Protocol
+
+1. **Verify Prerequisites**:
+   - Confirm the host OS is macOS
+   - Confirm changes were made to @src/pmdas/darwin
```
natoscott: Again here.

.claude/agents/macos-darwin-pmda-qa.md
```diff
@@ -0,0 +1,92 @@
...
+Do NOT:
+- Attempt to diagnose the root cause
+- Suggest fixes
+- Rerun tests without instruction
+- Modify any code or configuration
+
+### On Invalid Invocation:
+If prerequisites aren't met, clearly state:
+- "Cannot execute: Host OS is not macOS" OR
+- "Cannot execute: No changes detected in @src/pmdas/darwin directory"
```
natoscott: And here.

.claude/skills/macos-qa-test/SKILL.md
```diff
@@ -0,0 +1,48 @@
...
+- The VM will be provisioned, tests run, and results reported
+
+## Success Indicators
+
+- ✅ All build and test steps pass
+- Note: "unmounting the working directory" failure at end is expected/normal
+
+## When to Use
+
+- After making changes to `src/pmdas/darwin` or `src/pmdas/darwin_proc`
```
natoscott, line 41: Ah, this skill lists both - perhaps the earlier darwin-only references were intended after all.

.github/workflows/macOS.yml
```diff
@@ -31,12 +31,32 @@ jobs:
         id: install-brew-deps
         run: |
           brew install autoconf unixodbc
+      - name: Configure PCP
+        id: configure
+        run: |
+          echo "Configuring PCP..."
+          ./configure --prefix=/usr --libexecdir=/usr/lib --sysconfdir=/etc --localstatedir=/var
```
natoscott: Is this necessary?  The following build step with Makepkgs does the configure step as well

.github/workflows/macOS.yml
```diff
@@ -31,12 +31,32 @@ jobs:
...
+          echo "Running Darwin PMDA unit tests..."
+          cd dev/darwin/test/unit
+          ./run-unit-tests.sh
+      - name: Debug on unit test failure
+        if: failure() && steps.unit-tests.conclusion == 'failure'
+        run: |
+          echo "Unit tests failed. Checking build artifacts..."
+          find . -name "pmda_darwin.dylib" -o -name "pmdadarwin"
+          echo "Checking for pmcd build..."
+          find . -name "dbpmda"
```
natoscott, line 54: "dbpmda" here doesn't match the "checking for pmcd build" comment...?

.github/workflows/macOS.yml
```diff
@@ -94,6 +114,216 @@ jobs:
...
+            echo "Checking if pmcd is responding (${ELAPSED}s elapsed)..."
+
+            # Check if pcp command works (like in .cirrus.yml)
+            if pcp 2>/dev/null; then
+              echo "✓ pmcd service is responding!"
+              break
+            fi
+
+            # Also check if pminfo can connect
+            if pminfo -h localhost -f hinv.ncpu 2>/dev/null; then
```
natoscott, line 156: Hmm, why "-h localhost" here?  The default is "-h local:" which uses the unix domain socket to pmcd, then falls back to localhost if thats not available - I'd be inclined to not use any -h option here (like most users would).

.github/workflows/macOS.yml
```diff
@@ -94,6 +114,216 @@ jobs:
...
+          echo "=== Test simple metric with explicit 127.0.0.1 ==="
+          pminfo -h 127.0.0.1 -f hinv.ncpu 2>&1 || echo "pminfo with 127.0.0.1 failed"
+
+          echo ""
+          echo "=== Test simple metric with default host ==="
+          pminfo -f hinv.ncpu 2>&1 || echo "pminfo with default host failed"
+
+          echo ""
+          echo "=== PCP environment variables ==="
+          env | grep PCP || echo "No PCP env vars set"
```
natoscott, line 321: There's overlap with QA here (above couple hundred lines) - alot of this could be deleted if we instead did "cd qa; ./check -g sanity" ... which leads us into doing more complete QA as well.  Perhaps this could be a followup PR though.

build/mac/build-installer
```diff
@@ -96,8 +96,22 @@ pkgroot=`mktemp -d /tmp/macos-pkg.XXXXXXXXX`
 cp -RP installer-files/* $pkgroot/
 cp -RP installer-resources/* $pkgroot/
 
+# WORKAROUND: Write package to /tmp first to avoid virtiofs XAR corruption
+# then copy to final location in build directory
+# without this, the pkgbuild produces a corrupted archive inside a Tart VM
+pkgout="/tmp/pcp-pkg.$$"
```
natoscott: This script has a (securely generated) $tmp variable - use that instead of hard-coded /tmp?

dev/darwin/dev/build-quick.sh
```diff
@@ -0,0 +1,65 @@
...
+echo "========================================"
+echo "Build complete!"
+echo "========================================"
+echo "Binary: $(pwd)/pmdadarwin"
+echo "DSO:    $(pwd)/pmda_darwin.dylib"
+echo
+echo "Next steps:"
+echo "  1. Run unit tests:       cd ../test/unit && ./run-unit-tests.sh"
+echo "  2. Install to system:    make install"
+echo "  3. Test with pminfo:     pminfo -f darwin.mem.physmem"
```
natoscott: "darwin.mem.physmem"?

src/pmdas/darwin/version.c
```diff
@@ -70,4 +70,6 @@ macos_version(void)
     if (CFStringGetCString(productName, name, sizeof(name), kCFStringEncodingUTF8) &&
 	CFStringGetCString(productVersion, version, sizeof(version), kCFStringEncodingUTF8))
 	pmsprintf(version_string, sizeof(version_string), "%s %s", name, version);
+    CFRelease(plist);
```
natoscott, line 73: Good catch!

CONTRIBUTING.md
```diff
@@ -167,3 +167,10 @@ keep the differences minimal in order to ease future updates of vendored code.
 
 **All bugs and CVEs of a vendored library in turn become responsibilities of the
 PCP maintainers as well** (and fixes must be pushed upstream).
+
+## Platform-Specific Development
+
+### macOS with Tart VMs
+
+macOS developers can build in isolated VMs for clean, reproducible builds that match the CI environment. See [MACOS_DEVELOPMENT.md](MACOS_DEVELOPMENT.md) for details.
```
natoscott: I'd prefer this to link somewhere platform-specific rather than at the top-level of the git repo.  Perhaps build/mac or src/pmdas/darwin would suit.

src/pmdas/darwin/DARWIN-PMDA-ENHANCEMENT-PLAN.md
```diff
@@ -0,0 +1,183 @@
+# Darwin PMDA Enhancement Plan
```
natoscott, line 1: I think this should live below src/pmdas/darwin.

PMDA_REFACTOR.md
```diff
@@ -0,0 +1,655 @@
+ # Step 4.2: Refactor pmda.c Legacy Code
```
natoscott: I'd prefer this to live below src/pmdas/darwin too.

build/mac/MACOS_DEVELOPMENT.md
```diff
@@ -0,0 +1,85 @@
+# macOS Development with Tart VMs
```
natoscott, line 1: As in other comment, either build/mac or src/pmdas/darwin for this please.

tallpsmith:
The scripts were originally in scripts/Darwin but some asked me to move them here. So maybe you two rock paper scissors it and let me know. :)

natoscott:
I noticed in another PR there were heaps in build/mac - how about we go there by default, pmdas/darwin for unit tests and qa/darwin for anything that could/should become part of regular (installed) PCP QA over time?

tallpsmith:
@natoscott  do you remember which PR that was? (I honestly can't remember any in `build/mac' but there's quite a number of things going on at the moment I'm struggling to hold it in my head myself..

.github/workflows/macOS.yml
tallpsmith: I think this may be Claude's discovery (but didn't inform me properly) that the `configure` script was out of date, possibly due to another change when I re-labled `MacOSX` -> `macOS` which cause one of the platform detection scripts to fail.  I agree this should not be necessary.

src/pmdas/darwin/DARWIN-PMDA-ENHANCEMENT-PLAN.md
tallpsmith: I was going to delete it post-completion, but it makes sense in the interim to move it there, sure.

.github/workflows/macOS.yml
tallpsmith: Remnant of some debugging, wasn't sure if there was some weird tcp issue, but it was because `pmcd` hadn't actually been started properly...  Thought I got rid of them.

tallpsmith: Actually, this is in a Debug section, and I'd still like to keep it only because this section is aiming to detect any weird networking glitchy problems.  It only runs if something isnt' working, and is just confirming whether it can connect to `pmcd` somehow..?

.github/workflows/macOS.yml
tallpsmith: good question, not sure what Claude was doing there precisely...  I think it's exploring to see if it can actually find that binary, but the comment seems of, I'll adjust that.

build/mac/build-installer
tallpsmith: I'll have to triple check this again, because Claude and I had real trouble getting the packaging/installation to create a valid PKG file inside Tart, and this is how we got it to work, let me try again with the current state to see if it's still like that (maybe a Cirrus upgrade has addressed the original issue already).

dev/darwin/dev/build-quick.sh
tallpsmith: ah, yeah, Claude somehow seems to always want to prefix metric names...

PMDA_REFACTOR.md
tallpsmith: I deleted this one.

natoscott:
> @natoscott do you remember which PR that was? (I honestly can't remember any in `build/mac' but there's quite a number of things going on at the moment I'm struggling to hold it in my head myself..

Not sure the PR but its commit ce2f98b87fb27c4e972459560977a819d8e9bb3d

tallpsmith:
@natoscott  with the QA/packaging issue broken out into #2459 , are you happy to approve this change to merge to `main` and I'll get on the QA-rework as a seperate branch/changeset.

tallpsmith:
I'd note that this one is worth the next release.

tallpsmith:
Just need to attempt to correct Nathans comment on #2441 and include it here, as this Darwin additions branch has that other PR included (amongst other Tart/Cirrus changes I think), so #2441 is now superseded.

.github/workflows/macOS.yml
tallpsmith: Resolved as being catered for in a different issue #2459

PR #2444, opened 2026-01-07
add flake.nix for reproducible builds and NixOS

randomizedcoder:
Add comprehensive Nix packaging support for PCP 7.0.5, enabling reproducible
builds via Nix flakes and preparing for inclusion in nixpkgs.

New files:
- flake.nix: Main Nix flake with package definition, dev shell, and VM test
- flake.lock: Pinned nixpkgs and flake-utils versions
- nix/vm-test.nix: NixOS VM integration test for pmcd daemon
- nix/patches/gnumakefile-nix-fixes.patch: Remove ownership flags, fix
  tmpfiles path, exclude qa from build
- nix/patches/tmpdir-portability.patch: Use ${TMPDIR:-/tmp} instead of
  hardcoded /var/tmp for sandbox compatibility
- nix/patches/configure-ar-portable.patch: Portable ar fallback (not
  applied, available for upstream)
- docs/HowTos/nix/index.rst: Comprehensive documentation explaining Nix
  packaging, technical details, and future improvements

Features enabled by default (Linux):
- Core PCP tools, libraries, and PMDAs
- Python3 and Perl language bindings
- BPF/BCC kernel tracing PMDAs
- Systemd integration
- SNMP, device mapper, and service discovery

The flake provides:
- nix build: Build PCP package with split outputs (out, man, doc)
- nix develop: Development shell with build dependencies and PCP logo
- nix build .#checks.x86_64-linux.vm-test: NixOS VM integration test

Technical fixes for Nix sandbox:
- Patch /var/tmp to ${TMPDIR:-/tmp} for sandbox compatibility
- Remove install ownership flags (handled by NixOS module)
- Fix broken symlinks with double /nix/store prefix
- Override systemd paths from pkg-config
- Export AR for configure script

Documentation ./docs/HowTos/nix/index.rst includes explanations of why Nix
packaging can be challenging (non-FHS paths, sandboxed builds, pure
environment) and how these constraints help expose portability issues in
build systems.

PR #2445, opened 2026-01-10
build: shell script portability fixes (TMPDIR, ar path, shellcheck warnings)

randomizedcoder:
### Summary

This PR improves portability of PCP's build scripts for non-FHS systems (like NixOS, Guix) and sandboxed build environments. It also fixes shellcheck warnings to improve script quality.

New Pull request in response to
https://github.com/performancecopilot/pcp/pull/2444#pullrequestreview-3647196062

#### 1. Portable `ar` fallback in configure
The configure script had a hardcoded fallback to `/usr/bin/ar`, which doesn't exist on systems that don't follow the Filesystem Hierarchy Standard. Changed to just `ar` which will be found via PATH.

```diff
-test -n "$AR" || AR="/usr/bin/ar"
+test -n "$AR" || AR="ar"
```

Please note that the configure script has a very large number of shellcheck exceptions, so this pull request did NOT attempt to resolve these. 

#### 2. Use `${TMPDIR:-/tmp}` instead of `/var/tmp`
Several build scripts hardcode `/var/tmp` for temporary files. This fails in sandboxed build environments (like Nix) where `/var/tmp` is not writable. The fix uses `${TMPDIR:-/tmp}` which:
- Uses `$TMPDIR` when set (required for sandboxed builds)
- Falls back to `/tmp` on traditional systems
- Is POSIX-compliant

**Affected files:**
- `src/pmdas/bind2/mk.rewrite`
- `src/pmdas/jbd2/mk.rewrite`
- `src/pmdas/linux/mk.rewrite`
- `src/pmdas/linux_proc/mk.rewrite`
- `src/bashrc/getargs`
- `src/libpcp/src/check-errorcodes`
- `src/libpcp3/src/check-errorcodes`
- `src/libpcp/doc/mk.cgraph`
- `src/pmlogcompress/check-optimize`

#### 3. Shellcheck warning fixes (all POSIX sh compatible)

| Fix | Issue | Description |
|-----|-------|-------------|
| Trap quoting | SC2064 | Use single quotes so `$tmp` expands at signal time, not trap definition time |
| Test portability | SC2166 | Replace `[ -a ]` with `[ ] && [ ]` (more portable) |
| Command substitution | SC2046 | Quote `$()` to prevent word splitting |
| Variable init | SC2154 | Initialize `sts` before trap to avoid false warning |

### Testing

- Verified build completes successfully
- Tested with Nix flake build and NixOS VM test
- All shellcheck warnings resolved in modified files

### DIFF
--- a/configure
+++ b/configure
@@ -5947,7 +5947,7 @@ fi
 
   test -n "$AR" && break
 done
-test -n "$AR" || AR="/usr/bin/ar"
+test -n "$AR" || AR="ar"
 
 fi
 if test ! -x "$AR"; then
--- a/src/bashrc/getargs
+++ b/src/bashrc/getargs
@@ -5,7 +5,7 @@
 #
 
 verbose=false
-if [ $# -gt 1 -a X"$1" = X-v ]
+if [ $# -gt 1 ] && [ "$1" = -v ]
 then
 verbose=true
 shift
@@ -17,9 +17,9 @@ echo "Usage: $0 source-file ..."
 exit 1
 fi
 
-tmp=/var/tmp/bashrc-getargs-$$
+tmp=${TMPDIR:-/tmp}/bashrc-getargs-$$
 status=1
-trap "rm -f $tmp.*; exit \$status" 0 1 2 3 15
+trap 'rm -f $tmp.*; exit $status' 0 1 2 3 15
 
 for file
 do
@@ -56,7 +56,7 @@ $verbose && ( echo "+ source lines"; cat $tmp.work )
 if [ ! -s $tmp.work ]
 then
     echo "Error: no pmSetShortOptions call or .short_options assignment, nothing to work with"
-elif [ `wc -l <$tmp.work | sed -e 's/ //g'` -gt 1 ]
+elif [ "$(wc -l <"$tmp".work | sed -e 's/ //g')" -gt 1 ]
 then
     cat $tmp.work
     echo "Error: more than one pmSetShortOptions call or .short_options assignment, I am confused"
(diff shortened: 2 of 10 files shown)


randomizedcoder:
This Pull Request relates to https://github.com/performancecopilot/pcp/pull/2444

kmcdonell:
Ping @randomizedcoder ... any progress here?

kmcdonell:
G'day @randomizedcoder,
As of commit bad6e5301 2. above is done all across the PCP code base (not just the ones you've spotted below src).
Unless you want to engage in a conversation to convince me about 3., I think this is now down to 1.  Would you like me to "fix" this in configure.ac?
Hopefully we can move this PR to closure, then you'll need to rebase PR #2444.

randomizedcoder:
G'day mate!

I was just started to get back to this, but it looks like you fixed it up already!  Nice work.

I'm a pretty big fan of shellcheck, and don't see a good reason not to follow it.  John Carmack of Doom game fame says lean in on code checkers and I complete agree.

How can I help?

Thanks,
Dave

randomizedcoder:
I think we can close this pull request.

I will review the nix one again after this nice fixes you just applied.

randomizedcoder:
Created a pull request for "ar"
https://github.com/performancecopilot/pcp/pull/2479

kmcdonell:
OK, @randomizedcoder, I'm going to close this one.

Just one last shot on code checkers ... I've been coding in C, sh and troff for more than 50 years.  And my experience with code checkers is mixed.
- gcc is pretty good, but just gets it wrong sometimes, which is why the PCP code is littered with "pander to gcc" comments
- coverity (serious static analysis) gets it right just often enough to warrant making the effort, but about half the "issues" there turn out to be false positives
- mandoc -T lint (troff checker) appears to have been written by people who simply do not understand the troff language
- I have no experience with shellcheck (but I will do some research and learning).

Thanks.

PR #2446, opened 2026-01-12
Allow Manual QA GitHub action run with single OS trigger option instead of ALL of them

tallpsmith:
This change allows the Manual QA invocation from the Github actions to now take a selected OS/platform option so that you can just run a single operating system QA cycle.

I found this useful to devote some problems without launching the tire set of operating systems, which is fairly heavyweight.

Defaults to `all` maintaining backwards compatibility, but provides a selector of the platforms to run if you just want to select one.


tallpsmith:
I'm not quite sure why the QA run failed, I've successfully used this exact workflow in my branch.  It definitely works in my fork, and I wonder if something to do with the existing QA workflow (in the current `main` branch) is trying to do something with the new one?  

Anyway, if you're not convinced of it's use, we can drop it.

PR #2448, opened 2026-01-13
pcp2openmetrics: resolves RHEL-138467

lmchilton:
multi-instance metrics failing when queried with
singular metrics from archive. Turning interpolation on resolves this issue. Additional code added to fix missing headers when 2+ metrics are queried together.

### DIFF
--- a/src/pcp2openmetrics/pcp2openmetrics.py
+++ b/src/pcp2openmetrics/pcp2openmetrics.py
@@ -104,7 +104,7 @@ class PCP2OPENMETRICS(object):
         self.precision = 3 # .3f
         self.precision_force = None
         self.timefmt = TIMEFMT
-        self.interpol = 0
+        self.interpol = 1
         self.count_scale = None
         self.count_scale_force = None
         self.space_scale = None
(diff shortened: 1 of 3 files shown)


src/pcp2openmetrics/pcp2openmetrics.py
```diff
@@ -337,7 +337,7 @@ def execute(self):
             sys.stdout.write("Known metric spec keywords: " + str(self.pmconfig.metricspec) + "\n")
 
         # Set delay mode, interpolation
-        if self.context.type != PM_CONTEXT_ARCHIVE:
+        if self.context.type == PM_CONTEXT_ARCHIVE:
```
wcohen: How is this going to work with live data? I don't think that inverting the sense of the test is the right thing to do to turn on the interpolation.  Isn't that going to delay things for the archive. Looking at the pmrep.py has:
```
       # Set delay mode, interpolation
        if self.context.type == PM_CONTEXT_ARCHIVE:
            self.delay = 1
            self.interpol = 1
```
And it initialize interpol in PMReporter to 1 with the following rather than 0 in pcp2openmetrics.py: 

        self.interpol = 1

wcohen: "RHEL-11406" in the comment doesn't look correct. Should that be RHEL-138467?

PR #2451, opened 2026-01-15
Fix of pmseries core dumps for invalid expressions

kurik:
Fixes a crash in pmseries caused by expressions with improper types or units. The crash occurred because the serving function (batton->info) received an invalid data structure; this change ensures the correct structure is used.

### DIFF
--- a/src/libpcp_web/src/query.c
+++ b/src/libpcp_web/src/query.c
@@ -5063,7 +5063,7 @@ series_query_funcs_report_values(void *arg)
     seriesBatonReference(baton, "series_query_funcs_report_values");
 
     /* For function-type nodes, calculate actual values */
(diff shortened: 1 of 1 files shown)


kurik:
Reproduction scenario for the core dump:
```
pmseries 'kernel.all.pswitch[count:1] / kernel.all.pswitch[count:1]'
```

Without the fix, the `pmseries` crashes (generates core dump).
With the fix it correctly informs user about invalid use:
```
pmseries: [Error] Both operands have the semantics of counter, only addition or subtraction is allowed.
```

natoscott:
@kurik could we extend the regression tests for this?  qa/1212 looks like a possible home where we could add this as a third scenario.

PR #2453, opened 2026-01-17
pmrep: Add column grouping feature

tallpsmith:
## Overview

Add column grouping capability to pmrep - allowing configuration files to define visual group headers that span multiple metric columns, similar to vmstat output format.

## Why

When monitoring many related metrics (e.g., memory.free, memory.buff, memory.cache), they're difficult to mentally group without visual separation. Column grouping makes output more readable and easier to scan.

## What

**Example Output** (with group separators):
```
 % pmrep :macstat-grouped -t1 -s3          
           load |             memory            |   paging  |     io    |      cpu     
            1min|   free   wired  active   inact| pi/k  po/k| bi/k  bo/k|  us   sy   id
06:21:47    0.89     166     957    3401    3366   N/A   N/A   N/A   N/A  N/A  N/A  N/A
06:21:48    0.89     166     958    3400    3367     0     0     0     1    2    2   96
06:21:49    0.89     169     958    3396    3369     0     0     0     0    1    2   96
```

**Configuration Syntax**:

See `macstat-grouped.conf` included in the PR for further example.

```ini
[vmstat-grouped]
header = yes
groupalign = center          # Optional: left|center|right
groupsep = |                 # Optional: separator between groups

# Group definitions
group.procs = running, blocked
group.procs.prefix = kernel.all
group.procs.label = procs

group.memory = swap.used, free, bufmem, allcache
group.memory.prefix = mem.util
group.memory.label = memory

# ... metric definitions follow ...
```

## Implementation Status

- ✅ Core parsing and rendering complete
- ✅ 166+ unit tests passing
- ✅ Manual VM testing successful
- ⏳ QA integration tests to follow

## Design Notes

- **Backwards compatible**: Existing configs unchanged
- **Auto-detection**: Group headers enable automatically if `group.*` entries exist
- **Prefix resolution**: Leaf metric names get prefix prepended; FQDNs used as-is
- **Phase 1 scope**: Stdout output; CSV support planned for Phase 2


tallpsmith:
@myllynen - this is still draft, I just have to fix up some linting errors, and add in support for `--no-group-headers` CLI option to override any view configuration. But would love your eyes on this to see what you think.

I've refactored the code pulling out into different responsible code module/blocks, just because the `pmrep` file was getting _looooong_ and Claude was choking on it :-D

tallpsmith:
`--no-group-headers` option added:

```
admin@Manageds-Virtual-Machine pmrep % pmrep --no-group-headers -s3 :macstat-grouped
            1min|   free   wired  active   inact| pi/k  po/k| bi/k  bo/k|  us   sy   id
01:06:51   10.48     361     800    3528    3488   N/A   N/A   N/A   N/A  N/A  N/A  N/A
01:06:52   10.48     361     801    3526    3490     0     0     0     0    1    2   96
01:06:53   10.48     361     801    3525    3492     0     0     0     0    2    2   96

admin@Manageds-Virtual-Machine pmrep % pmrep -s3 :macstat-grouped 
           load |             memory            |   paging  |     io    |      cpu     
            1min|   free   wired  active   inact| pi/k  po/k| bi/k  bo/k|  us   sy   id
01:06:58    9.80     258     807    3585    3527   N/A   N/A   N/A   N/A  N/A  N/A  N/A
01:06:59    9.80     258     808    3583    3528     0     0     0     0    1    1   97
01:07:00    9.02     260     807    3582    3529     0     0     0     0    2    2   97

```

natoscott:
@tallpsmith FWIW, this is looking increasingly like pcp-dstat ... would it be worth exploring that on macOS rather than adding more pmrep options?  🤷

tallpsmith:
> @tallpsmith FWIW, this is looking increasingly like pcp-dstat ... would it be worth exploring that on macOS rather than adding more pmrep options? 🤷

@natoscott  Ok, well, I never knew that existed!  Doesn' this violate the `pm*` standard tooling convention?  I see a number of tools in the `pcp` tree now I had no idea about (I thought that directory was related to a core library thing.  I best get exploring...

Having said that, there may still be legitimate niceties to group metrics in 'pmrep' for whatever reason, but it may certainly reduce the likelihood of needing a more 'adhoc' tool like `pmrep` now ?  

Still, this just opens another macOS-specific bug to fix:

Screenshot 2026-01-20 at 11 39 46 am

(sorry, copy/paste seems to have choked in my VM and can't seem to fix it ).  I'll open a bug.

I can see we probably won't need the `:macstat` `pmrep` config if I can get `pcp dstat` to work on macOS.

natoscott:
That's coming from src/pcp/dstat/plugins/cpu:

```
usr = 100 * (rate(kernel.percpu.cpu.user) + rate(kernel.percpu.cpu.nice) + rate(kernel.percpu.cpu.irq.soft) + rate(kernel.percpu.cpu.irq.hard))
```

This is a derived metrics expression, as in pmRegisterDerived(3).  Its likely the IRQ metrics that are problematic here, and I think it might be solvable with the "defined" keyword - @kmcdonell will know for sure though.

natoscott:
>  Doesn't this violate the pm* standard tooling convention?

Yes, so historically these tools are those that have a matching non-PCP implementation, and we had several issues to solve as a result.  We wanted to match the original tool arguments as best we can (which may conflict with usual PCP options).  We wanted a way to pass those standard PCP command line options (like -h/-a/-t, etc) into the PCP versions, which can be done via "pcp --host www.example.com --interval 5sec dstat" (because the pcp(1) front end handles the PCP arguments separately and magically passes that info to the tool).

Every now and then I toy with the idea of making all the tools accessible this way (e.g. "pcp info", etc) ... which could be done by a bunch of symlinks initially ... but I never get around to following up and exploring that further.  I think that could be a better user experience though, if all tools were accessible this same way.

kmcdonell:
@natoscott you're correct, these metric names need to be "wrapped" in some defined() magic.  I've just push commit e9d51aac70 that does the wrapping for the "irq", "wait" and "steal" CPU metrics.
It remains open to decide if pcp-dstat should say nothing and report 0 for "metrics" that have no underlying instrumentation, e.g. `wai`, `hiq`, `siq` and `stl` in the example below:
```
kenj@vm06:~$ dstat --cpu-adv
------total-cpu-usage------
usr sys idl wai hiq siq stl
  1   2  96   0   0   0   0
  0   2  96   0   0   0   0
```
And of course this is the tip of the iceberg, don't even think about `dstat --vm` for a non-Linux system, and I've not looked at any of the other `dstat` derived metric definitions.

natoscott:
Awesome!

> And of course this is the tip of the iceberg, don't even think about `dstat --vm` for a non-Linux system, and I've not looked at any of the other `dstat` derived metric definitions.

Replaying Linux archives on macOS - and vice-versa - is another wrinkle we need to keep in mind here.

kmcdonell:
@natoscott Replaying will "just work" because the metadata will match the PMNS for the PMAPI context.

If it works on Linux for a local pmcd, an archive created there will replay just fine on macOS or *BSD.  And vice versa.

Or am I missing something obvious?

natoscott:
> Or am I missing something obvious?

I was just thinking in relation to your "tip of the iceberg" and "dstat --vm for a non-Linux system" comments ... some of these may be tough to solve via derived metrics (memory metrics in particular, can be totally different between platforms).

tallpsmith:
@natoscott  I (personally) think this is nice to have for the next release, but if we want to hold off on a review until the next one that's also cool by me.

natoscott:
@tallpsmith @myllynen pmrep is Marko's baby so let's wait to hear his thoughts.  No rush on this one though I guess, looks like 7.1.1 material to me if he's keen.

myllynen:
Pardon me for the huge delay here, at some point I completely forgot about this until now.

From the user perspective the feature looks nice and can probably be helpful with larger metricsets. The required configuration options look reasonable for the purpose and require no changes for existing definitions.

However ...

I see new tests that look like not being part of the standard PCP QA framework? Since this PR @kmcdonell published https://github.com/performancecopilot/pcp/blob/main/qa/COOKBOOK.md, could that help our AI friends to create test cases that are part of the QA framework or what might be the best approach here? Changing the string to represent truncated metrics breaks at least QA 1062, not sure what kind of impact the rest of changes have there.

My main concern however is that the current PR is recreating/rewriting parts of the pmrep internals to support this feature. So who would maintain this new code going forward? One could argue that modernizing the pmrep code, dropping all the Python 2 era stuff, etc is long overdue but would it make sense for the whole PCP have more holistic approach here?

So instead of rewriting or modernizing one corner of pmrep code in isolation would it be better to use AI to gradually modernize PCP Python code for Python 3.6+ first (some such changes have been added recently), including clean up and refresh of pmconfig/pmrep code, and then add this new functionality? The changes for this feature may then have much lesser impact when there's clear separation of modernization vs new functionality. And perhaps this overall modernization could consider the old idea of drop-in plugins for customized pmrep output.

But the above idea is of course a daydream scenario assuming someone would have spare cycles for such a coordinated Python modernization. I guess it's evident by now that these days I have barely any time for PCP related activities and given other priorities everyone has I'd assume this won't be anyone's top priority going forward.

I'm not sure what would be the best way forward here, especially considering everyone's time constraints.

PR #2457, opened 2026-01-20
Extended test 1212 with invalid semantic case

kurik:
Added `invalid semantic` case to the pmseries test 1212
This is a follow up of https://github.com/performancecopilot/pcp/pull/2451


kmcdonell (approved):
LGTM

PR #2469, opened 2026-01-23
Drop the dependency on python-six

hrnciar:
Since python 2 is EOL, there is no need to keep python-six around.


kurik:
I did some preliminary testing on Fedora-43 and RHEL-10. All looks good from my perspective.

PR #2475, opened 2026-01-29
pmlogger_daily: add disk size limit option and enforce archive purging

orasagar:
Introduce -d/--disk option for setting a maximum total disk size for PCP archives. Update documentation and script logic to parse size values with K/M/G suffixes, enforce purging of oldest files when the limit is exceeded, and improve size reporting. This helps manage storage usage more effectively.

### DIFF
--- a/man/man1/pmlogger_daily.1
+++ b/man/man1/pmlogger_daily.1
@@ -20,6 +20,7 @@
 .B $PCP_BINADM_DIR/pmlogger_daily
 [\f3\-DEfKMNnoPpQrRVzZ?\f1]
 [\f3\-c\f1 \f2control\f1]
+[\f3\-d\f1 \f2fssize\f1]
 [\f3\-k\f1 \f2time\f1]
 [\f3\-l\f1 \f2logfile\f1]
 [\f3\-m\f1 \f2addresses\f1]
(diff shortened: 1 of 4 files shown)


kmcdonell:
@orasagar sorry I botched the build, but this is now fixed.  Could you please rebase or pull latest commits from main, so we rerun the checks here and get a cleaner bill of health?

orasagar:
@kmcdonell I have successfully re-based my branch on top of the current main branch

orasagar:
Hi @kmcdonell ,

Thank you for taking the time to review my work. I appreciate your valuable feedback and am currently addressing your comments. I will be back with the required fixes and updates shortly.

kmcdonell:
@orasagar hung in there ... I'm part way through a big review and I'll send you a patch set with proposed changes, later today or tomorrow.

kmcdonell (changes requested):
Thanks, this is looking much better.

I still have a long list on "nits", but most of them I've already
addressed and I've uploaded a patch file for your consideration.

Firstly logutil.sh (and some of this also applies to pmlogger_daily.sh,
especially the "style" parts).

- In PCP, our scripts use sh not bash for portability, so constructs
  like
  if (( foo * bar > 12345 )) need to be rewritten as
  if [ \`expr $foo \* $bar\` -gt 12346 ] and ditto for
  foo=((bar * 1024 * 1024)) which becomes
  foo=\`expr $bar \* 1024 \* 1024\`

- And as you can gather from the above, we consistently (except where
  nesting forces otherwise) use \`cmd\` in preference to $(cmd) ... based
  on my 50+ years of shell programming, I simply don't accept the
  arguments proposed by shellcheck and others to prefer $(cmd) over
  \`cmd\`.

- Another style nit, we (or at least I) use
  if <expr>
  then
    ...
  rather than
  if <expr>; then
  ...
  Ditto for while ...; do
  Since your changes are going into a mass of existing shell scripts,
  I'd expect the style to remain consistent.

- QA ... this cannot proceed with out tests added to the qa dir.
  So far I've added
  1863 - exercise _convert_to_kb()
  1864 - exercise new pmlogger_daily features

- The "units" are a bit of a problem ... KB (and MB and GB) are the SI
  sanctioned abbreviations for the 10^3, 10^6 and 10^9 multipliers
  (not the base 1024 ones) ... I've simplified this to drop all the
  ?b and ?B variants, and use Kbytes in error|warning|info messages.
  This also matches the man page!  And I think a unit should be mandatory
  everywhere, not just in the environment variable setting.

- If the semantic tests were re-ordered in _convert_to_kb() then
  the error messages would not complain about units unless the number
  is correct.

- Some suggested wordsmithing of error message text in _convert_to_kb()
  for consistency and readability.

- I'm not sure what the overflow tests are checking ... where would the
  limitation of a 32-bit integer mess with this?  Even on a 32-bit
  platform I believe expr(1) does 64-bit arithmetic.  If you drop this,
  qa/1863 will need updating and qa/1863.out will need to be remade.

- I'm not sure allowing whitespace between <num> and <unit> in
  _convert_to_kb() is a good idea ... extra whitespace is only inviting
  user errors IMHO [this is not changed in the patch, I leave this to
  you to decide] ... if you keep it, you definitely need the quoting
  I added around $1 in the initial assignment in this function ... if
  you drop it you'll need to fix qa/1863 and remake the qa/1863.out
  file.

- I prefer SPACELIMIT (and $PCP_SPACELIMIT) to SIZE_LIMIT (and
  $PCP_SIZE_LIMIT) ... "size" is an overloaded term and this is more
  consistent with the similar control variables, like CALLBACK

- We need a "unlimited" value for -d or $PCP_SPACELIMIT to reinstate
  the default in the case where we have multiple control files with
  different settings in each.

- what is this trying to do?
  \# Enforce explicit unit, do NOT accept empty unit as KB!
  I think it would be better to demand a unit everywhere ... see note
  above about "units".

Moving on to in pmlogger_daily.sh

- I've re-wordsmithed the -d description in the Usage.

- Moved the -d block of arg processing to be in alphabetical order (makes it
  easier to find and reduces the risk of "doubling up" some option). Ditto
  for the Usage text and the man page.

- Removed some superflous blank lines, fixed the indentation ... for reference
  all my source file editing is done in vi with :set sw=4 ts=8 ai

- I don't believe we need a $PCP_PURGE_CALLBACK, and the # purge callback
  initialization block does not appear to do anything useful ... I've culled
  this, but please put it back if you can explain why and what it is doing.

- Conventionally (in PCPland) local shell functions begin with one _, not
  two.

- The _do_purge() logic is not right.  You need to treat archives as atomic
  units, not individual files and handle archives spread over multiple
  directories for a single pmlogger instance (search for wobbly in the
  pmlogger.control(5) man page to see how this happens).
  I've reworked all of this.

There is probably more, but I decided to concentrate on fixing the code and
QA at this point, so the review notes petered out.

Suggest you apply the patches, review the changes and adjust as you see fit
then re-commit and I'll review any differences between what I have and what
you commit.

[review.patches.txt](https://github.com/user-attachments/files/25118711/review.patches.txt)

orasagar:
@kmcdonell, thank you for your thorough and thoughtful review. I sincerely appreciate you taking the time to provide such detailed feedback. Additionally, I have noted your comments about expr, and this will be added to my learning journey moving forward.

kmcdonell:
@orasagar OK, I want to move this along.  I'm going to merge this PR, then I'll push the missing qa files in a separate commit.

orasagar:
@kmcdonell sorry I was out this weekend so didn't get chance to get my mail.
thanks for the merge

PR #2476, opened 2026-01-29
Upstream htop code sync

natoscott:


randomizedcoder:
@natoscott 

It's interesting that PCP is vendoring in various dependencies.  Ideally, once the Nix pull request is merged ( https://github.com/performancecopilot/pcp/pull/2444 ), then it would be possible to use the nixpkgs version of htop, for example, and so PCP could benefit automatically from the changes in nixpkgs

The nixpkg htop seems to be tracking 3.4.1
https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/ht/htop/package.nix#L22C14-L22C19

Build via Nix will essentially mean if it builds once with Nix on any operating system, then it will work on all the other operating systems

natoscott:
@randomizedcoder I don't know anything about Nix, but I think the situation is more subtle here.

While htop is vendored, its not a traditional dependency like a library - there is substantial amounts of PCP code in the 'upstream' htop which allows the pcp-htop binary to be build from the same code base as htop itself.

This results in two different binaries - pcp-htop uses the PMAPI (hence offers UI for many thousands of metrics not supported by regular htop, distributed operation, etc), and htop uses syscalls directly to query fixed kernel metrics.

PR #2480, opened 2026-02-03
pcp2opentelemetry: resolves rhel-130463

lmchilton:
Few minor updates to make pcp2opentelemetry output OTLP compatible. Fixed json dumps output, timestamp type, and deleted extra instance field and header lines.

Resolves RHEL-130463


wcohen:
The patch looks sane.  The patch has been tested with the various pcp2telemetry tests (1131 and 1977)? The one thing that is missing is additions to the qa testsuite that ensures that pcp2opentelemetry is providing valid OTLP output and doesn't generate output listed as problems in RHEL-130463.  Another thing that the tests should check is that more than one sample works correctly (something other than -s1).  qa test 1131 exercises pcp2openmetrics with the -s2 option. Given some of the previous issues with pcp2openmetrics it would be good to add some testing to ensure that multiple samples works properly for pcp2opentelemetry also.

lmchilton:
To-Do: update archive sample to make it compatible for merge

qa/1131
```diff
@@ -31,7 +31,7 @@ signal=$PCP_BINADM_DIR/pmsignal
 trap "_cleanup; exit \$status" 0 1 2 3 15
 
 A="$here/archives/rep"
-A1="$here/archives/20180606"
+A1="$here/archives/20260211"
```
wcohen: Why the need for the new archive 20260211?  It looks like archives/20180606 has
the desired hinv.ncpu and disk.partitions.write metrics for 1131. The
new 20260211 archive has a lot of stuff.  The 20260211.0.zst is 13MB
and there is another 400KB in 20260211.meta.xz. Would it make sense
just to use the existing archives/20180606 and update the 1131.out
with updated output.

src/pcp2opentelemetry/pcp2opentelemetry.py
wcohen, line 1: The changes to pcp2opentelemetry.py look reasonable.  All the pcp2opentelemetry tests work with:

`./check -g pcp2opentelemetry`

wcohen:
The updated patch eliminated the additional archive and things look to be fine in the revised patch.

Issue #2481, opened 2026-02-06
The pmproxy process should not be running as a unconfined_t


wcohen:
Recent updates to Fedora rawhide bohdi update process checks the labels of running processes from an RPM.  These checks do not allow the 
[pcp-7.1.0-1.fc44](https://bodhi.fedoraproject.org/updates/FEDORA-2026-07cfd5f7f5) to be pushed.   Looking through the tests the basic problem is that when the things like pmlogger are not running unconfined_t.  In the output of the testing see:

```
::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
::   Test
::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::

:: [ 04:29:29 ] :: [  BEGIN   ] :: Running 'ps -efZ | awk '$1~/.*unconfined.*/{print $9}' |             sort -u |             tee /var/tmp/EhqK80C9wejx/unconfined.processes'
awk
bash
/bin/bash
cat
dbus-broker
ps
sort
tee
/usr/bin/dbus-broker-launch
/usr/libexec/pcp/bin/pmproxy
/usr/lib/systemd/systemd
:: [ 04:29:29 ] :: [   PASS   ] :: Command 'ps -efZ | awk '$1~/.*unconfined.*/{print $9}' |             sort -u |             tee /var/tmp/EhqK80C9wejx/unconfined.processes' (Expected 0, got 0)
:: [ 04:29:29 ] :: [  BEGIN   ] :: Running 'belongsto=$(rpm -qf /usr/bin/awk)'
:: [ 04:29:29 ] :: [   PASS   ] :: Command 'belongsto=$(rpm -qf /usr/bin/awk)' (Expected 0, got 0)
:: [ 04:29:29 ] :: [   LOG    ] :: awk belongs to gawk-5.3.2-3.fc44.x86_64 package
...

:: [ 04:29:29 ] :: [  BEGIN   ] :: Running 'belongsto=$(rpm -qf /usr/libexec/pcp/bin/pmproxy)'
:: [ 04:29:29 ] :: [   PASS   ] :: Command 'belongsto=$(rpm -qf /usr/libexec/pcp/bin/pmproxy)' (Expected 0, got 0)
:: [ 04:29:29 ] :: [   LOG    ] :: /usr/libexec/pcp/bin/pmproxy belongs to pcp-7.1.0-1.fc44.x86_64 package
[...]

PR #2482, opened 2026-02-06
Fix selinux issue after pmproxy startup scrip relocation

kurik:
Fixes issue #2481, where the pmproxy startup script lost its SELinux context after relocation from /etc/pcp/pmproxy/rc to /usr/libexec/pcp/services/pmproxy.


kurik:
I tested this fix on Fedora-Rawhide and the development version of RHEL-10.2 with the latest upstream PCP (commit https://github.com/performancecopilot/pcp/commit/a836765cbf6d2c3f5b6fc56e8dd22ffe936c79d4). The patch fixes the issue.
@natoscott , @wcohen can  you please review the patch and merge (if OK) ?

natoscott:
@kurik good catch!  LGTM.

PR #2488, opened 2026-02-10
Fix AVC denial for rocestat pmda

wcohen:
When testing out the new rocestat pmda on RHEL 10 we found that rocestat pmda would trigger a AVC denial (RHEL-132402).  The small patch to src/selinux/pcp.te allows the pmda to function when SEL is enforced.


kmcdonell:
Beyond my selinux knowledge I'm afraid.

PR #2489, opened 2026-02-10
Fix PCP selinux policy for rhel-8 after rc scripts relocation

kurik:
This commit fixes selinux policy for rhel-8 after rc (startup) scripts relocation to /usr/libexec/pcp/services/.


kmcdonell:
@kurik Unfortunately this change does not work for the (related I think) failures I'm seeing on vm14 x86_64 CentOS Stream10.

kurik:
@kmcdonell I am closing this, as agreed via email.

PR #2492, opened 2026-02-11
QA #1321 fix to address failure on s390x

lmchilton:
PR #2448 caused QA test failure #1321 on s390x machines


natoscott:
This doesn't make sense to me in the context of the test - see the comment at line 64, it seems that loop is supposed to be happening at the same time that pmlogger is running (in parallel).  This change means we do the two stages serially, one after the other, and the pmlogger is put in the background for no reason now (we sleep and then kill it?  hmm, bit sus).

PR #2493, opened 2026-02-11
selinux: Update nvidia pmda policy

wcohen:
When running the nvidia pmda found a couple more AVC denials that needed to be addressed.

RHEL-133519


kmcdonell:
LGTM

PR #2496, opened 2026-02-13
Add new valkey PMDA for monitoring Valkey servers

natoscott:
This PMDA collects performance metrics from Valkey servers via the INFO command. Valkey is an open source, in-memory data store forked from Redis.

The PMDA provides metrics in the following categories:
- Server: uptime, process ID, configuration information
- Clients: connected clients, blocked clients, tracking and pub/sub
- Memory: usage, fragmentation ratio, dataset and overhead memory
- Stats: command and connection counts, keyspace hits/misses, evictions
- CPU: CPU time consumed by server and child processes
- Persistence: RDB snapshot and AOF status information

Configuration is via /etc/pcp/valkey/valkey.conf with host and port settings.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sfeifer:
Hey Nathan. There are a few things on my [branch](https://github.com/sfeifer/pcp/commit/3ea529003830bc4a38c761e89bf75769e73cd92d) we may want to add.

I added to build/rpm/pcp.spec.in to get Makepkgs builds working. I think we’d also need to add to build/rpm/redhat.spec too. Are there other files under build/ that would need additions to properly package the valkey pmda?

For the Install script, I think I needed to add dso_opt=true for it to work correctly after installing the rpms from a Makepkgs build. Trying out your branch, though, that does not seem to be the case. Not quite sure what happened, but is this something we want to add to the Install script?

The pmid’s and metric structure is different between this valkey pmda and the old redis pmda. The redis pmda does not have metric clusters (other than the main redis cluster). Do we want to now have metric clusters in the valkey pmda, or keep the same metric structure as the redis one?

Also on the metrics, the redis pmda has 71 metrics while this valkey pmda only has 36. Is this shortened set of metrics intentional?

I have a bit of a different Makefile structure that works for me when running Makepkgs builds locally. It might be worth looking into simplifying the pmda valkey makefile since a bunch of the libvalkey files are in the libpcp_web makefile (could reuse those). I’ll defer to your experience here :).

natoscott:
> Hey Nathan.

HI, thanks for the notes & review.

> There are a few things on my [branch](https://github.com/sfeifer/pcp/commit/3ea529003830bc4a38c761e89bf75769e73cd92d) we may want to add.
> 
> I added to build/rpm/pcp.spec.in to get Makepkgs builds working. I think we’d also need to add to build/rpm/redhat.spec too. Are there other files under build/ that would need additions to properly package the valkey pmda?
> 

Since we use a bundled libvalkey, and its all C code, there are no external dependencies pmdavalkey can be installed with the other core PCP agents in the pcp rpm/deb packages.  No need for spec file changes as a result.

> For the Install script, I think I needed to add dso_opt=true for it to work correctly after installing the rpms from a Makepkgs build. Trying out your branch, though, that does not seem to be the case. Not quite sure what happened, but is this something we want to add to the Install script?
> 

This agent communicates using sockets, so I went with the no-DSO approach here (don't want to introduce potential for blocking pmcd).

> The pmid’s and metric structure is different between this valkey pmda and the old redis pmda. The redis pmda does not have metric clusters (other than the main redis cluster). Do we want to now have metric clusters in the valkey pmda, or keep the same metric structure as the redis one?
> 

Clusters are good.

> Also on the metrics, the redis pmda has 71 metrics while this valkey pmda only has 36. Is this shortened set of metrics intentional?
> 

Looks like there's plenty that can be added still - 'valkey-cli info' - shows plenty more possibilities.  I'll do an audit and add more a follow up commit, thanks!

> I have a bit of a different Makefile structure that works for me when running Makepkgs builds locally. It might be worth looking into simplifying the pmda valkey makefile since a bunch of the libvalkey files are in the libpcp_web makefile (could reuse those). I’ll defer to your experience here :).

AFAICT, libpcp_web doesn't export and libvalkey symbols (which is a good thing) so we cannot directly use the libpcp_web so.  There is currently no dependency between pmdavalkey and libpcp_web, at the source or binary level, so I think I prefer to keep that the way it is now and have both use the vendored code independently.

PR #2498, opened 2026-02-18
pcp2arrow: add pmFetchArchive optimization for raw archive mode

natoscott:
## Summary

Optimise pcp2arrow archive processing using pmFetchArchive API for --raw/-r mode, observed 64x speedup on a very large (high sampling frequency, many hours, 22 volumes) benchmark archive while maintaining backwards compatibility.

## Performance

Large archive (21,340 metrics, 1000 samples): 5m29s → 5.5s (64x faster)

## Implementation

- execute_raw(): Dispatches to pmFetchArchive path for archives with --raw flag
- _iter_archive_raw(): Iterator yielding (result, timestamp_ns) tuples
- _process_raw_result(): Handles dynamic instance discovery
- Skip pmSetContextOptions for raw archive mode (no interpolation needed)

## Behaviour

Non-raw (default): Interpolated, rate-converted metrics using pmFetchGroup  
Raw mode (-r): Direct archive reading at recorded frequency, no interpolation, 64x faster


kmcdonell:
@natoscott Caveat: pcp2arrow ignorance on display here ...
Are the archives pcp2arrow chews on ones with the **same** metrics in each pmResult?
And are the timestamps on the archive given to the Apache Arrow ecosystem required to be regular?
Are instances not expected in the metricspec from the command line?
If any condition is not true, then the differences in semantics between pmFetch() and pmFetchArchive() my bite the unsuspecting punter.

natoscott:
| Are the archives pcp2arrow chews on ones with the same metrics in each pmResult?

Not necessarily, no.  We're building up a potentially sparse matrix here, its OK to have missing values all over the place (except for timestamps).

| And are the timestamps on the archive given to the Apache Arrow ecosystem required to be regular?

No.

| Are instances not expected in the metricspec from the command line?

I think that can happen (depends on pmconfig.py semantics, not 100% sure), but think its OK as metricspecs are handled differently to the regular fetch path here (we accept eveything from the raw result initially, and then cull things we don't want if a metricspec was provided).

| If any condition is not true, then the differences in semantics between pmFetch() and pmFetchArchive() my bite the unsuspecting punter.

Yes, absolutely, this is an unusual case where the data structure we're moving to (sparse matrix) can handle all of these quirks.  (hence @myllynen why I'm not convinced we should push any of this into common py API code)

PR #2499, opened 2026-02-19
pmdabcc: removed completely now after retirement last release

natoscott:


natoscott:
>     * Comment block at head of src/pmdas/bpf/Upgrade probably needs some post cut-n-paste care to speak about bcc -> bpf, rather than notready and OpenMetrics.

Fixed, thanks.

>     * Also I was surprised to see no changes below qa/admin/package-lists and for qa/admin/other-packages/manifest (not just python bcc or bpfcc but all the other things dragged in for QA of the bcc PMDA)

Should all be resolved now via second commit on that branch.

>     * Shouldn't ‎src/pmdas/bcc/modules/fs/__init__.python be deleted? shows as "Whitespace-only changes" ... ditto for src/pmdas/bcc/modules/__init__.python
> 

It is/was git-rm deleted, not sure why its showing up like that there.

>     * for retired qa tests, I presume the plan is to use qa/really-retire on 'em in a subsequent round of changes, or you could do it now? (so these ones will be like qa/1152 that is already really-retired)

I've done that now, and also dropped common.bcc which I missed before - all in the second commit if you could cast an eye over it once more.

qa/admin/other-packages/unavailable
kmcdonell, line 1: Strictly speaking, you don't need to make changes here ... this is a record of fact wrt the package availability for the distros.  Mentioning python3-bcc or bpfcc-tools here will have no effect if they have been culled from the manfest.

qa/group
kmcdonell, line 1: Could probably add a note here about PMDA being retired and today's date.

kmcdonell (approved):
Couple of small nits in per-file comments as I passed through.
But if this compiles, then ship it!

natoscott:
Thanks Ken, tweaked and merged.

PR #2505, opened 2026-02-26
pmcd: fix endianness state extraction issue

lmchilton:
On Big-Endian systems (s390x), the flags used by pmda fetch functions to signal a PMDA change fail to be sent and extracted by PMCD.

Raw memory access via memcpy and pointer aliasing correctly targets the most significant byte, causing the status flags to be misplaced or zeroed.

The proposed changes replace byte-level manipulation with numeric assignment to the tv_sec field, ensuring architecture-neutral behavoir.

### DIFF
--- a/qa/1321
+++ b/qa/1321
@@ -72,7 +72,7 @@ done
 
 # When new dynamic metrics are detected, the _next_ fetch will fetch them.
 # So wait for a bit so pmlogger has a chance to fetch the new metrics.
-sleep 2
+sleep 3
 
 # now kill pmlogger and check the new metrics appeared dynamically
 $sudo $PCP_BINADM_DIR/pmsignal -s TERM $pmlogger_pid >/dev/null 2>&1
--- a/src/libpcp_pmda/src/callback.c
+++ b/src/libpcp_pmda/src/callback.c
@@ -476,18 +476,18 @@ pmdaInstance(pmInDom indom, int inst, char *name, pmInResult **result, pmdaExt *
  * The first byte of the (unused) timestamp field has been
  * co-opted as a flags byte - now indicating state changes
  * that have happened within a PMDA and that need to later
- * be propogated through to any connected clients.
+ * be propagated through to any connected clients.
  */
 
 static void
 __pmdaEncodeStatus(pmdaResult *result, unsigned char byte)
 {
-    unsigned char	*flags;
+    uint64_t    *flags;
 
     memset(&result->timestamp, 0, sizeof(result->timestamp));
     if (byte) {
-	flags = (unsigned char *)&result->timestamp;
-	*flags |= byte;
+        flags = (uint64_t *)&result->timestamp;
+        *flags = byte;
     }
 }
 
--- a/src/pmcd/src/dofetch.c
+++ b/src/pmcd/src/dofetch.c
@@ -356,9 +356,8 @@ SendFetch(DomPmidList *dpList, AgentInfo *aPtr, ClientInfo *cPtr, int ctxnum)
 static int
 ExtractState(int i, void *timestamp)
 {
-    unsigned char	byte;
+    uint64_t flag = *((uint64_t *) timestamp);
 
-    memcpy(&byte, timestamp, sizeof(unsigned char));
     /*
      * integrity checks on the state change ...
      * - only 7 bits defined in pmapi.h (so max value is 0x3f)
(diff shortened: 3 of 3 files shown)


qa/1321
```diff
@@ -72,7 +72,7 @@ done
 
 # When new dynamic metrics are detected, the _next_ fetch will fetch them.
 # So wait for a bit so pmlogger has a chance to fetch the new metrics.
-sleep 2
+sleep 3
```
wcohen, line 75: This change to the qa/1321 looks unneccesary.  Are there timing issues on the s390x that require this?

src/libpcp_pmda/src/callback.c
```diff
@@ -482,12 +482,9 @@ pmdaInstance(pmInDom indom, int inst, char *name, pmInResult **result, pmdaExt *
 static void
 __pmdaEncodeStatus(pmdaResult *result, unsigned char byte)
 {
-    unsigned char	*flags;
-
     memset(&result->timestamp, 0, sizeof(result->timestamp));
     if (byte) {
-	flags = (unsigned char *)&result->timestamp;
-	*flags |= byte;
+        result->timestamp.tv_sec |= (long)byte;
```
wcohen: This change doesn't match up with the behavior listed in the comment for __pmdaEncodeStatus() as the byte will now be stored in the LSB of timestamp.tv_sec field and requires a change to ExtractState() in dofetch.c:

```
/*
 * The first byte of the (unused) timestamp field has been
 * co-opted as a flags byte - now indicating state changes
 * that have happened within a PMDA and that need to later
 * be propogated through to any connected clients.
 */
```
Is should be possible to rework __pmdaEncodeStatus() to always use the first byte of the timestamp field regardless of the endianess of the machine.  This would make this patch more concise. Maybe also correct the mispelled "propagated" in the comment in the updated patch.

wcohen (changes requested):
Looks like the idea is solid.  Would like to the patch updated to eliminate the change to ExtractState().

qa/1321
kmcdonell: I'm with @wcohen here, I'd like to see if there is a reason for this change.

src/libpcp_pmda/src/callback.c
kmcdonell, line 1: The comment does not match the code.  You've co-opted the first 64-bits (not the first byte) of the (unused) timestamp field to encode the flags.
I also think it would help those that follow if this comment block was expanded to include words along the lines of
```
This encoding of a PMDA's change of state is passed back to pmcd where it is decoded
in ExtractState() in dofetch.c.
```


I think the code would be clearer if __pmdaEncodeStatus() was changed so the second argument was `__uint64_t flags` instead of `unsigned char byte`, then change the name of the pointer variable from `flags` to `ptr` or something similar.   In fact `ptr` is not needed 

Note, we have historically preferred to use the PCP-defined `__uint64_t` over the `uint64_t` for portability reasons ... I suspect those arguments are now largely historical, but for code consistency I'd vote to use the double-underscore prefixed variants for the type.

And in this same source file, change the type of `flags` in the caller pmdaFetch() to also be `__uint64_t`.

So the body of  this routine reduces to something like
```
*((__uint64_t *)&result->timestamp) = flags;

```

src/pmcd/src/dofetch.c
kmcdonell, line 1: I don't follow this.  Why is it not simply the converse of the __pmdaEncodeStatus logic, so something like:
```
__uint64_t flags  = *((__uint64_t *)timestamp);
```
and then use `flags` instead of `byte` in the rest of the code in this routine.

Also add a comment cross reference at the head of this routine, similar to the other place, like
```
For PMDAs using libpcp_pmda, the encoding of PMDA state changes is performed in
__pmdaEncodeStatus() in that library's callback.c.
```

The comment about 7 bits and the use of 0x3f in the code that follows is all a bit messy and brittle, but I'll clean that up once you've merged this PR.

src/pmcd/src/dofetch.c
```diff
@@ -356,9 +356,11 @@ SendFetch(DomPmidList *dpList, AgentInfo *aPtr, ClientInfo *cPtr, int ctxnum)
 static int
 ExtractState(int i, void *timestamp)
 {
-    unsigned char	byte;
+    uint64_t full_value;
+
+    memcpy(&full_value, timestamp, sizeof(uint64_t));
+    unsigned char byte = (unsigned char)full_value;
```
wcohen: Some old C compilers may not like the declaration of `byte` after the `memcpy()` statement.  Given that this is just a fetch of a value it might be better to eliminate the  `memcpy() `and simplify the code to:

`    uint64_t byte = *((uint64_t *) timestamp);`

The above also closer to the line used `__pmdaEncodeStatus()` to store the information.  Calling the variable `byte` is a bit of a misnomer.  Change the references of `byte` to `flags` (and `new_flags`) in `ExtractState()` to be more descriptive.

natoscott:
@wcohen @lmchilton you seem to have missed the feedback from @kmcdonell on this change?

lmchilton:
@natoscott I used Ken's feedback in the first revision of the PR. I tried using int32 but it continued to fail; int64 worked though. Did I miss anything else?

natoscott:
| I tried using int32 but it continued to fail

this information is new (?) and would be an important addition to this discussion.  :)

natoscott:
And I wonder if there's still a problem here then, because as Ken mentioned on some systems tv_sec is 32 bits not 64 bits (casting to time_t* might be worth a try?).

lmchilton:
@natoscott yes - I was wondering if there were any unforesceen effects using int64, so I requested a second review from Ken. @wcohen and I were in sync on this PR, but we probably should have waited for his review before merging. I will try casting to time_t* though, thanks.

wcohen:
I thought about the issue of using uint64_t in that place and worried about something outside being overwritten.  There are two different uses of ExtractState.  One passes in __pmResult * and the other passes pmdaResult.

```
changes |= ExtractState(i, &results[j]->timestamp);
changes |= ExtractState(i, &rp->timestamp);

```
The "timestamp" fields are two different types. For the first line it coulde be could be pmResult or pmResult_v2 and for the second line it could be __pmTimestamp:

```
typedef struct pmResult {
    struct timespec	timestamp;	/* time stamped by collector */
    int			numpmid;	/* number of PMIDs */
    pmValueSet		*vset[1];	/* set of value sets, one per PMID */
} pmResult;

typedef struct pmResult_v2 {
    struct timeval	timestamp;	/* time stamped by collector */
    int			numpmid;	/* number of PMIDs */
    pmValueSet		*vset[1];	/* set of value sets, one per PMID */
} pmResult_v2;

typedef struct {
    __int64_t	sec;
    __int32_t	nsec;
} __pmTimestamp;

```
From man pages struct timespec and timeval are:

```
       struct timespec {
           time_t     tv_sec;   /* Seconds */
           /* ... */  tv_nsec;  /* Nanoseconds [0, 999'999'999] */
       };


       struct timeval {
           time_t       tv_sec;   /* Seconds */
           suseconds_t  tv_usec;  /* Microseconds */
       };

```
In all the cases there should be 64 bits available to store the flags even on 32-bit machines as the struct timespec/timeval would have two 32-bit fields.

natoscott:
| In all the cases there should be 64 bits available to store the flags

No question there is plenty of space to receive the data.

The issue is what is happening at the other end of the connection where the flags are being set in the first place - on a 32 bit system, the flags write will happen in the first 32 bits, I think, not the first 64 bits.  Its possibly further complicated by both DSO and process PMDAs being on the other end of the "connection" (there is no connection for DSO PMDAs, its all happening in-memory in pmcd and maybe with different result types).  It needs very careful auditing.

lmchilton:
@natoscott @wcohen casting to time_t * worked.

PR #2509, opened 2026-03-02
pcp-system-tools: restore backward compatibility with older pmapi versions

orasagar:
Commit 0a37ed0ec8e93a30b3ee349b78c23ba708caef45 introduced support for pmapi version 4, but upgrading to the latest pmapi is not currently desired. This change updates pcp-iostat, mpstat, pidstat, and ps to fall back to tv_usec when tv_nsec is unavailable during timestamp delta calculation, ensuring compatibility with older pmapi versions and collectors that only expose tv_usec.

### DIFF
--- a/src/pcp/iostat/pcp-iostat.py
+++ b/src/pcp/iostat/pcp-iostat.py
@@ -56,9 +56,18 @@ class IostatReport(pmcc.MetricGroupPrinter):
     Hcount = 0
     def timeStampDelta(self, group):
         s = group.timestamp.tv_sec - group.prevTimestamp.tv_sec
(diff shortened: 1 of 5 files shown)


kmcdonell:
I'm going to "speak" with @orasagar via email to understand the use case here ... _a priori_ I don't think this is the sort of thing we want to carry forward in the PCP code base.

natoscott:
Agreed, I don't think this category of changes should be merged upstream.

orasagar:
@natoscott after having discussion with @kmcdonell i do also agree that it should not be upstream, closing this out

PR #2514, opened 2026-03-03
qa/1987:pcp-ps sort/format overage

orasagar:
Add new pcp-ps test cases for explicit/user/ALL columns sorted by %cpu and %mem

-optimized code in pcp-ps for dynamic sorting
- fixed sorting colum indexes in -u case

### DIFF
--- a/src/pcp/ps/pcp-ps.py
+++ b/src/pcp/ps/pcp-ps.py
@@ -375,7 +375,7 @@ class DynamicProcessReporter:
 
     def _is_last_and_args(self, key):
         return (key == "args") and \
-        self.processStatOptions.colum_list.index(key) == len(self.processStatOptions.colum_list) - 1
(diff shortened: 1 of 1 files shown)


src/pcp/ps/pcp-ps.py
```diff
@@ -855,6 +817,10 @@ def handle_o():
             try:
                 if optarg.upper() == "ALL":
                     self.filterstate = optarg.upper()
+                    self.colum_list = [
```
natoscott: @orasagar looks good, just a small typo on "colum_list" (should be "column_list")

orasagar: @natoscott  Fixed the issue

natoscott:
@orasagar QA test 1987 is failing on all CI hosts with this change - could you take a look?  thanks!

orasagar:
@natoscott it will be resolve by https://github.com/performancecopilot/pcp/pull/2529

PR #2517, opened 2026-03-04
rc scripts: pmcd, pmlogger and pmproxy allow local only access by default

natoscott:
Resolves Debian bug #1128921


kmcdonell (approved):
LGTM, thanks @natoscott

johnmaguire:
Thanks for this! :-)

mvollmer:
@natoscott, I understand pmcd, but does it make sense to also restrict pmproxy to listening on localhost?

PR #2528, opened 2026-03-13
pmrep.py: add units to CSV header output

orasagar:
Previously, pmrep displayed units in stdout but omitted them when exporting data with the CSV option. This change updates the CSV header to include units, aligning CSV exports with the stdout format.


natoscott:
QA test updates/additions please.

natoscott (changes requested):
Needs some QA - updates to existing and/or new tests please.  Possibly a man page update too?

src/pmrep/pmrep.1
```diff
@@ -813,6 +816,10 @@ Requires
 .TP
 .I csv
 Print metrics in CSV format (subject to formatting options).
+When unit information is enabled (the default), a second header row
+listing unit strings is emitted after the column names; use
```
myllynen: This and the previous description doesn't match the QA output, there is no second line but instead the unit info is after the metric name.

qa/1169.out
```diff
@@ -1021,34 +1021,34 @@ Time,"disk.dev.read-sdb","disk.dev.write-sdb"
...
-Time,"disk.dev.read-sda","disk.dev.read-sdc","disk.dev.write-sda","disk.dev.write-sdc","mem.util.used"
+Time,"disk.dev.read-sda (count)","disk.dev.read-sdc (count)","disk.dev.write-sda (count)","disk.dev.write-sdc (count)","mem.util.used (Kbyte)"
 1970-01-01 00:00:13,1.000,1.000,1.000,1.000,13
 1970-01-01 00:00:14,1.000,1.000,1.000,1.000,14
 1970-01-01 00:00:17
-Time,"disk.dev.read-sdb"
+Time,"disk.dev.read-sdb (count)"
 1970-01-01 00:00:18,1.000
-Time,"disk.dev.read-sdb","disk.dev.write-sdb"
+Time,"disk.dev.read-sdb (count)","disk.dev.write-sdb (count)"
```
myllynen: Do we have any QA cases that test the output without unit info? Perhaps a couple of such cases would be still good to have.

src/pmrep/pmrep.py
```diff
@@ -1182,6 +1209,15 @@ def remove_delimiter(self, value):
                 value = value.replace(self.delimiter, " ")
         return value
 
+    def sanitize_header_field(self, value):
```
myllynen: I think this might be better as a nested function or at least having csv in its name, not sure if the name could be slightly more descriptive as well.

src/pmrep/pmrep.py
```diff
@@ -932,19 +943,35 @@ def write_header_csv(self, results=()):
...
+                if first:
+                    self.writer.write("\"" + first + "\"")
+                else:
+                    self.writer.write("")
+                for entry in entries[1:]:
+                    self.writer.write(self.delimiter + "\"" + entry + "\"")
+            self.writer.write("\n")
+
+
+
```
myllynen: Otherwise looks good but let's avoid these unncessary empty lines.

myllynen (changes requested):
I think few minor changes would be needed, please see the comments for details.

src/pmrep/pmrep.py
orasagar: extra lines have been removed

src/pmrep/pmrep.py
orasagar: Function name have been made more descriptive

qa/1169.out
orasagar: added new test case for no unit , although new 1929 have been added as well

src/pmrep/pmrep.1
orasagar: It has been left as it since we want to match the csv behavior as default behavior where units should come in next row, so code has been fixed for it .

qa/1169.out
```diff
@@ -938,6 +938,47 @@ QA output created by 1169
 == csv test
 1970-01-01 00:00:00
 Time,"disk.dev.read-sda","disk.dev.read-sdb","disk.dev.write-sda","disk.dev.write-sdb"
+,"count/s","count/s","count/s","count/s"
```
myllynen: I don't think would work, I also checked with two AIs and both agreed only one header line is generally understood by csv-consuming utilities. I think the earlier option was much better one.

src/pmrep/pmrep.py
```diff
@@ -1182,6 +1206,15 @@ def remove_delimiter(self, value):
                 value = value.replace(self.delimiter, " ")
         return value
 
+    def sanitize_CSV_header_field(self, value):
```
myllynen: We use lower-case csv in other function names so best to do so here as well.

myllynen (changes requested):

src/pmrep/pmrep.py
orasagar: function has been renamed

qa/1169.out
orasagar: I have checked by exporting test data into csv 

image
image

orasagar: it seems to be working fine @myllynen

kurik:
Even the [RFC-4180](https://www.rfc-editor.org/rfc/rfc4180.html) states that `there is no formal specification in existence`, it still describes the `header` as the **first line**.

Do we really want to implement this feature ? At least it breaks tools like `sqlite` or PostgreSQL`s imports from CSV files.

myllynen:
I think the original version which kept everything on one line would still work by default, however I remember seeing some tools failing to interpret anything but the first line as the header so it would feel prudent to avoid a multi-line header.

orasagar:
I have updated the code and test cases.

natoscott:
@orasagar I'm concerned about the approach where the column headers are changed, and no longer simple metric/instance names.  I think it will break existing uses of csv-exporting in data science libraries like pandas etc, where the column name become accessible in code.

My recommendation is to have a new flag that produces a new comment section at the start of the file - fixed number of lines containing the metadata for each metric, maybe like...
```
#semantics: rate,instant,instant,...
#units: count/sec,none,none,...
#type: double,float,float,...
timestamp,kernel.all.pswitch,kernel.all.load["1 minute"],kernel.all.load["5 minute"],...
[...data...]
```

@myllynen mentioned maybe the -x option could be used, but there's some unrelated issue there.

orasagar:
@natoscott but user has an option to use -u (no unit) option where it will not have units information in the header.
It will not break the existing usage of csv export option in any library.

The intent of this PR was to make CSV output as similar to default STDOUT.
Having said that I do agree with your point to display the units with metric name make it default for parser to parse it.

I think by default we can give the unit information in the next row like (similar to STDOUT) which I initially tried to do .and rest this semantic and type information in case of extended header information.

natoscott:
> It will not break the existing usage of csv export option in any library.

It will for people using this feature today, relying on the current header form, then upgrading - the change in default will break their setup.  The -u (no units) would need to be the inverse (add units) to not break things.

orasagar:
@natoscott I can implement the inverse behavior, wherein specifying the -u option will include unit information in the CSV output. Under this approach, units would be presented in the row immediately following the header, consistent with how they are displayed in the default STDOUT format.

myllynen:
So the considerations are whether to display unit info by default or by request and how that info should look like. As mentioned above, changing the default might break existing use cases so probably best to avoid that.

The other question would be how to provide that information, should be like this as is proposed:

```
Time,"disk.dev.read-sdb(count)","disk.dev.read-sdd(count)","disk.dev.write-sdb(count)"
```

or as part of the extended header, something like:

```
...
# interval: 1
# duration: 00:12:34
# semantics: rate,instant,instant,...
# units: count/sec,none,none,...
# type: double,float,float,...
# 
```

The former might work more easily with some tools while the latter could convey more information. Or perhaps these could be implemented both? Without the extended header the change would be as proposed now and additionally the extended header could provide this extra info. Or perhaps we could start with the CSV-only change and create a follow-up issue/RFE for the extended header update if deemed helpful.

natoscott:
```
# interval: 1
# duration: 00:12:34
```

This information is redundant - can be calculated trivially from the timestamp column - so I recommend not adding it.

orasagar:
Hi @natoscott / @myllynen ,


Here is my takeaway from the above conversation.
Will leave the default csv output as it is because change it might break the usage of csv output with pandas or such libraries.

Will implement -M(--include-units) option which will include the units 

```
...
Time,"disk.dev.read-sdb(count)","disk.dev.read-sdd(count)","disk.dev.write-sdb(count)"
```

and for extended information will use existing -k(--extended-csv    write extended CSV ) option 
which will have 

```
...
# semantics: rate,instant,instant,...
# units: count/sec,none,none,...
# type: double,float,float,...
# 

```
In this PR will just implement -M option for now , for this extended header thing will raise a different PR.

myllynen:
The -M option is reserved for other ideas, see [src/pmrep/TODO](https://github.com/performancecopilot/pcp/blob/main/src/pmrep/TODO).

I think we could re-use -U for this with some man page changes and possibly or perhaps ideally adding also a long option for this, like --csv-unit-info or so.

I was planning to do some extended header updates but I'll hold on with those so that you don't have to rebase although that sounds unlikely given the scope you outlined.

myllynen:
We're getting very close, two relatively minor issues I see:

1) The new QA 1929 is not testing CSV output in the first two cases even though the descriptions say so, I'd suggest to update the commands and output there to match the test descriptions.

2) I don't think we could remove or alter any existing command-line options so if possible the --csv-unit-info should be an additional one and --no-unit-info would stay as-is. Hopefully this could be done in a relatively straightforward manner.

Thanks.

qa/1929.out
```diff
@@ -0,0 +1,117 @@
...
+00:00:12        12        12         12         12       N/A        12
+00:00:13       N/A       N/A        N/A        N/A       N/A        13
+00:00:14       N/A       N/A        N/A        N/A       N/A        14
+00:00:15       N/A       N/A        N/A        N/A        15       N/A
+00:00:16       N/A       N/A        N/A        N/A        16       N/A
+00:00:17        17       N/A        N/A        N/A       N/A       N/A
+00:00:18        18       N/A         18        N/A       N/A       N/A
+00:00:19        19       N/A         19        N/A       N/A       N/A
+=== pmrep csv output with units header (-U)
+Time,"disk.dev.read-sda","disk.dev.read-sdb","disk.dev.read-sdc","disk.dev.read-sdd","disk.dev.read-sde","disk.dev.write-sda","disk.dev.write-sdb","disk.dev.write-sdc","disk.dev.write-sdd","disk.dev.write-sde","mem.util.free","mem.util.used"
```
myllynen: This looks not matching the test case description? So if -U should work in this case why are units included here?

myllynen (changes requested):

qa/1929.out
orasagar: Fixed it my bad

myllynen (approved):
Thank you for your perseverance. It looks good now although one tiny part of the QA is perhaps a bit unexpected but I'd say let's merge this and once I'm fixing the extended header part I'll run all the QA then.

PR #2557, opened 2026-04-08
libpcp_web: fix libuv teardown; pmfind: defer context release and shutdown order

kurik:
Add `pmWebTimerLoopFinalize()` to `uv_close` the global instrumentation timer after `pmWebTimerClose` and export it in PCP_WEB_1.24. Order `pmWebGroupClose` so `struct webgroups` is freed only from `uv_close` callbacks (timer then async) and clear `module->privdata` early to avoid double teardown.

In `pmfind`, defer `source_release` to the next tick to avoid use-after-free when `pmWebGroupDestroy` runs from `on_done`; delete from contexts dict before destroy; track defer_pending and use extra `uv_run`/`uv_stop` as needed. Call `pmWebGroupClose` and `pmWebTimerLoopFinalize` before `uv_loop_close` with draining `uv_run` passes.

This is a followup of the https://github.com/performancecopilot/pcp/pull/2555 which fixes unsafe multi-threaded use of timers during request handling.
Later changes fix libuv shutdown rules (do not free memory that still holds closing handles) and pmfind-specific ordering (destroy vs. deref vs. dict, and draining the loop before `uv_loop_close`). Those are separate failure modes; fixing one does not imply the other is fixed.

### DIFF
--- a/src/libpcp_web/src/timer.c
+++ b/src/libpcp_web/src/timer.c
@@ -86,6 +86,20 @@ pmWebTimerClose(void)
     uv_mutex_unlock(&server.callback_mutex);
 }
 
+/*
+ * uv_close(2) the global instrumentation timer after pmWebTimerClose().
+ * Call before uv_loop_close(); run the loop once to process the close cb.
+ */
+void
+pmWebTimerLoopFinalize(void)
+{
+    if (server.events == NULL)
+	return;
+    if (uv_is_closing((uv_handle_t *)&server.timer))
+	return;
+    uv_close((uv_handle_t *)&server.timer, NULL);
+}
+
 static void
 timer_worker(uv_timer_t *arg)
 {
--- a/src/pmfind/source.c
+++ b/src/pmfind/source.c
@@ -11,6 +11,7 @@
  * or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
  * for more details.
  */
+#include <stdlib.h>
 #include <uv.h>
 #include "dict.h"
 #include "pmapi.h"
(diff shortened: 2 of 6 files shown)


kurik:
Hmmm... Ubuntu 18.04 is using old libuv which does not support `uv_handle_get_loop()`  yet :-(
Will need to find a way around it....

src/libpcp_web/src/timer.c
```diff
@@ -86,6 +86,20 @@ pmWebTimerClose(void)
     uv_mutex_unlock(&server.callback_mutex);
 }
 
+/*
+ * uv_close(2) the global instrumentation timer after pmWebTimerClose().
+ * Call before uv_loop_close(); run the loop once to process the close cb.
+ */
+void
+pmWebTimerLoopFinalize(void)
```
natoscott, line 94: Just in terms of naming, "pmWebTimerDestroy" might be more consistent with other parts of libpcp_web?

src/pmfind/source.c
```diff
@@ -281,12 +351,32 @@ source_discovery(int count, char **urls)
...
+     * uv_close callbacks on this loop (Approach A — minimal shutdown).
+     */
+    pmWebGroupClose(&settings.module);
+    uv_run(loop, UV_RUN_DEFAULT);
+
+    pmWebTimerLoopFinalize();
+    uv_run(loop, UV_RUN_DEFAULT);
+
+    uv_close((uv_handle_t *)&timing, NULL);
+    uv_run(loop, UV_RUN_DEFAULT);
```
natoscott, line 373: Oh wow, so many explicit calls to uv_run ... are they all really necessary?

kurik:
Thanks @natoscott  for the review.
I agree that this patch still needs some work. For some reason, which I still need to investigate, it does not work on older Debian and Ubuntu releases as expected. Unfortunately, I do not have those older distros readily available for testing, so this might take some time.

The issue itself seems to be orthogonal to the fix delivered in #2555; this patch just revealed it.

I will keep working on this and ping you for another review once it is ready.

kurik:
I started the root cause analysis from scratch for this issue and find out I was wrong with the proposal. As such I am closing this PR as it is superseded by #2559 .

PR #2558, opened 2026-04-08
pmdasmart: Add nvme_error_log.total metric

pauljevans:
Add smart.nvme_error_log.total metric that reports the total number of NVMe error log entries per disk, returning zero when no errors exist. Available for both device name and WWID instances.

### DIFF
--- a/src/pmdas/smart/smart_stats.c
+++ b/src/pmdas/smart/smart_stats.c
@@ -1433,6 +1433,7 @@ nvme_error_log_refresh(void)
 		int found_nvme_log = 0;
 		int count = 0;
 		int total;
+		dev->nvme_error_log_total = 0;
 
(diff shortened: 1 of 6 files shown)


src/pmdas/smart/smart_stats.c
```diff
@@ -1446,9 +1447,11 @@ nvme_error_log_refresh(void)
 				continue;
 			}
 
-			if (found_nvme_log && strstr(buffer, "No Errors Logged"))
+			if (found_nvme_log && strstr(buffer, "No Errors Logged")) {
 				// No error entries head to next drive
+				dev->nvme_error_log_total = 0;
```
natoscott: The logic here is a bit questionable - when we break here, this zeroing of nvme_error_log_total may be overwritten unintentionally with 'count' outside the loop, no?

pauljevans: Indeed, the statement is redundant also as the value is initialised on line 1436. I will update the commit to fix this. Thanks

PR #2574, opened 2026-04-26
Selinux fixup for pmcd on F43

kmcdonell:


wcohen (commented):
One thing I find odd is that there is very similar rule for pcp_pmproxy_t from git commit 6c1cbd04507 (2023-10-04) by Zdenek Pytela.  However, it has an ifdef around it in pcp.te:

```
ifdef(`kernel_io_uring_use',`
	kernel_io_uring_use(pcp_pmproxy_t)
')
```

Is a similar ifdef needed for the pcp_pmcd_t addition to allow this to run on older distributions?

natoscott:
> Is a similar ifdef needed for the pcp_pmcd_t addition to allow this to run on older distributions?

Probably not - pmproxy uses libuv which uses async I/O (epoll, etc) which somewhere along the path must interact using io_uring it seems.  pmcd is much simpler and just uses blocking I/O (select, etc) for all PDU transfers.

kmcdonell:
@natoscott
The problem was observed with the shping PMDA in qa/597 (I have the wrong qa test in the commit message, bugger).  And in this test, the shping PMDA does indeed run nslookup.  And qa/597 already had a _notrun for F39 because of a similar situation (I have no explanation as to why this "issue" was not noticed for F40, F41 and F42).

This is part of (I suspect) a much bigger lurking problem.  I am guessing (from a position of Selinux ignorance, and to be frank, distrust) that pmcd launches all PMDAs with the pmcd_t context and this means they could do pretty much anything to trigger any random AVC.

So far, whack-a-mole has been used to expand the pcp.te policy for the pmcd_t context each time a new PMDA or a change in Selinux base policies causes an AVC breach.

DSO PMDAs means we cannot change the pmcd_t context for PMDAs, so we're stuck.

In this particular case, I think the right thing (given the relative low value of the shping PMDA in general, and running nslookup in particular) is to _notrun qa/597 if Selinux is installed.

If others agree, I'll close this PR.

If not, I think @wcohen 's observation is correct ... if kernel_io_uring_use is guarded elsewhere it needs to be guarded everywhere.

wcohen:
To head off future problems if someone has a PMDA that ends up anon inodes it might be better to include this policy change. 

I took a look at [selinux-policy git repo](https://github.com/fedora-selinux/selinux-policy.git) to get a better idea of the need for `kernel_io_uring_use`.  The particular selinux-policy commit that added kernel_io_uring_use interface was
git commit 65b9e0bdceb7e6adbe308f9a591b103cba6986ef:

```
Author: Ondrej Mosnacek <omosnace@redhat.com>  2022-07-07 10:33:54
Committer: Zdeněk Pytela <zdenek@pytela.net>  2023-10-03 16:01:53
Parent: 39896a0f307a7c580b18575f17226eb0222efb35 (Allow targetd write to the syslog pid sock_file)
Child:  c412df7396671cf9f83cf2a23816ce7bdbfdbcb2 (Allow udev_t to search all directories with a filesystem type)
Branches: rawhide, remotes/origin/c10s, remotes/origin/f40, remotes/origin/f41, remotes/origin/f42, remotes/origin/f43, remotes/origin/f44, remotes/origin/rawhide
Follows: v38.29
Precedes: v40.1

    Implement proper anon_inode support
    
    Drop the existing allow-all workaround that simply allows all operations
    on anon inodes and replace it with a better approach.
    
    First, create individual types for each anon inode type (currently
    userfaultfd and io_uring) and add transition rules for all domains so
    that any anon inode they try to create gets the correct label.
    
    Then create interfaces to allow common userfaultfd/io_uring usage to
    individual domains.
    
    Finally, grant the necessary permissions for known anon inode users in
    the system policy (currently only libvirt) and to unconfined domains.
    
    Note that the transition rules added in this commit require the policy
    compiler to support self keyword in type transitions, which was
    introduced in libsepol/checkpolicy version 3.4.
    
    Signed-off-by: Ondrej Mosnacek <omosnace@redhat.com>
```
Zdeněk Pytela the committer of this change to selinux-policy is the same person that made the pcp selinux.te git commit 6c1cbd0450 fix that has the `ifdef()`.  I suspect the `ifdef` is in there as it looks like RHEL9 selinux-policy doesn't have `kernel_io_uring_use`.

wcohen (approved):
Looks good to me.

PR #2583, opened 2026-05-04
Making PCP a bit friendlier to AI agents

wcohen:
The [AgentReady](https://github.com/ambient-code/agentready) tooling reviews a git repository and suggest changes to make it easier for AI agents to navigate and improve the overall processes for development for the code in the git repository.  AgentReady scores the results.  The initial assessment of PCP was towards the bottom of the Bronze (42.5).  The patches in this merge request provides:

- A symbolic link for tests to qa to make it easier for AI to determine where the tests are (standard project layout)
- Headers in the README.md to address (concise documentation)
- github issue templates and pull request checklist (Issue & Pull Request Templates)

These changes moves PCP's accessment to the middle of the Bronze (51.8).  There are some other areas for improvement flagged by AgentReady that are not yet addressed:

- type annotations of the python code as PCP has now moved to python 3 as requirement.  This is going to take some work as there is less than 5% of the python code that currently has annotations
- pre-commit hooks & ci/cd linting to automate those check
- additional linting configuration ("Code Smell Elimination") to catch python and markdown issues.


kurik:
Would it make sense, as part of this PR, to rename CLAUDE.md to AGENTS.md ?
Both files serve the same purpose, however CLAUDE.md is used just by the Claude agent, while AGENTS.md is consumed by most of the AI tools.

natoscott:
@kurik yes it would IMO - Claude doesn't read that though, so what I see other projects doing (e.g. vllm) is to create a CLAUDE.md that contains:
```
@AGENTS.md
```

wcohen:
@kurik and @natoscott, the AgentReady CLAUDEmdAssessor class does accept a `@` reference or a symbolic link to the AGENTS.md file, so the suggested changes would both pass the tests.  I don't know which would be "better."  However, having the PCP codebase widening the acceptable AI agents with that type of change would be be an improvement.

kurik:
> the AgentReady CLAUDEmdAssessor class does accept a @ reference or a symbolic link to the AGENTS.md file, so the suggested changes would both pass the tests. I don't know which would be "better." 

I do not have any strong opinion, not even experience, here. My personal preference would be to have only the `AGENTS.md` file and no `CLAUDE.md`. However as @natoscott mentioned above, other project are referencing the `AGENTS.md` via `@` operator in the `CLAUDE.md` file. So, perhaps we can follow this pattern as well.

.github/ISSUE_TEMPLATE/pcp_issue_report.yml
```diff
@@ -0,0 +1,60 @@
+name: Performance Co-Pilot Issue Report
+description: Create a report for a PCP related issues on Linux
+title: "[Issue] "
```
natoscott: Does adding "[Issue]" to every non-RFE issue really help in any way?  I'd prefer to leave the title string as was originally provided if not.

wcohen: Yes, "[Issue]" is too generic.  Looking at how the [papi issues](https://github.com/icl-utk-edu/papi/issues) which this was based on there are two templates described it .github/ISSUE_TEMPLATE but there is a third "Blank issue" choice when filing a new issue.  Maybe it would be better to this called a bug template for PCP.

wcohen: I did some fixup to make the issue template a more descriptive bug template.  I also collected up the patches and put them on an updated checkout of pcp git repo branch.

tallpsmith:
> > the AgentReady CLAUDEmdAssessor class does accept a @ reference or a symbolic link to the AGENTS.md file, so the suggested changes would both pass the tests. I don't know which would be "better."
> 
> I do not have any strong opinion, not even experience, here. My personal preference would be to have only the `AGENTS.md` file and no `CLAUDE.md`. However as @natoscott mentioned above, other project are referencing the `AGENTS.md` via `@` operator in the `CLAUDE.md` file. So, perhaps we can follow this pattern as well.

The `@AGENTS.md` pattern seems to be the best option.  Annoying that Claude Code doesn't natively support it, because the extra `CLAUDE.md` files are clutter, but I'd personally be screwed without them now, so thanks for taking the time to migrate but maintain Claude support. I agree for an open-source project that being agnostic and aligned on `AGENTS.md` makes sense though.

wcohen:
@tallpsmith Any suggested changes to the patches? Or is this good to be merged into upstream?

sfeifer:
@wcohen I tested out the new AGENTS.md and CLAUDE.md structure locally with claude and it is working.

PR #2585, opened 2026-05-06
pmdaamdgpu: Prevent out-of-bound references in amdgpu_fetch function

### LINKED ISSUE #2584: potential off by one in `amdgpu_fetch`
https://github.com/performancecopilot/pcp/blob/7338e8aeb84de30dafd5f4ba8e93533d5eae9838/src/pmdas/amdgpu/amdgpu.c#L356-L358

it appears to me that this `pmidlist` array has `numpmid` elements, so it should go from 0 to numpmid-1, while it uses `<=` here. many other for loops in that code are using `<` instead. is this  a fencepost error?

wcohen:
The valid range of indices for the pmidlist array is 0 to numpmid-1. The loop in amdgpu_fetch was incorrectly using a "i <= numpmid" rather than "i < numpmid" to terminate the loop.

Resolves: https://github.com/performancecopilot/pcp/issues/2584

### DIFF
--- a/src/pmdas/amdgpu/amdgpu.c
+++ b/src/pmdas/amdgpu/amdgpu.c
@@ -353,7 +353,7 @@ static int amdgpu_fetch(int numpmid, pmID pmidlist[], pmdaResult **resp,
 {
   uint32_t i = 0;
 
-  for (i = 0; i <= numpmid; i++) {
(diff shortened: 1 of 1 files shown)


natoscott:
@wcohen LGTM

PR #2587, opened 2026-05-07
qa: unset LC_ALL in the check script to fix locale-dependent sort order

kurik:
When LC_ALL is set in the caller's environment (e.g. LC_ALL=en_US.utf8),
  it overrides any per-command LC_COLLATE=POSIX setting used inside
  individual QA tests, causing locale-sensitive sort ordering instead of
  the expected POSIX byte-value ordering.  This manifests as spurious
  output mismatches in tests such as 821/1663 where metric names are
  enumerated via "LC_COLLATE=POSIX sort".
  Unsetting LC_ALL at the start of the check script removes the inherited
  override while still allowing individual tests to apply their own
  per-command locale settings (e.g. "LC_COLLATE=en_US.utf8 sort") as
  intended by their authors.

### DIFF
--- a/qa/check
+++ b/qa/check
@@ -21,6 +21,9 @@ __aborted=true
 myname=`basename $0`
 __scriptname=$myname  #  a synonym
 
+# Ensure LC_ALL is not set to allow locale-dependent sort order
+unset LC_ALL
+
 # lock (and PID) file to prevent concurrent QA execution
 #
 CHECKLOCK=/tmp/PCP-QA-LOCK


kurik:
@kmcdonell  May I ask you please for a review, to confirm this idea?
It hit me when I saw many tests to fail on my downstream test-bed , just because by an accident the LC_ALL was propagated from my workstation to the test-bed global environment.

kmcdonell (approved):
Well I'll be darned ... I'm 76 years old, reviewing from Tashkent Uzbekistan, and I'm still learning ... it had never occurred to me that LC_ALL trumps all the finer-grain LC_FOO locale environment variables.

Fix looks 100% OK to me, I'll approve and merge.

Thanks @kurik

PR #2602, opened 2026-05-28
libpcp_web: accept NaN/Inf in series_pmAtomValue_conv_str on float/double

kurik:
The first-character digit check rejected pmAtomStr_r() output like "nan" from 0.0/0.0 on s390x (positive quiet NaN), breaking pmseries division tests 1886 and 1906.

For some reason the FPU on `s390x` returns NaN with sign bit clear which is different from most of the other architectures which return negative NaN (sign bit set). Both ways are allowed by IEEE 754 which states only that Nan should be returned (no sign defined).

Edit: I can reproduce this on `ppc64le` arch as well, so it is not specific for `s390x` FPUs.


natoscott:
@kurik LGTM.

PR #2603, opened 2026-05-28
qa: cover ubuntu2604 differences in some tests

kurik:
qa/1429 - fix mkdir error message:
ubuntu2604 uses uutils coreutils which returns different error message when `mkdir` command fails for a permission reason.
    
fix _filesize():
On Fedora/RHEL (and older distros), `stat` prints 'Size' while on Ubuntu26.04 `stat` prints 'size'. The fix covers this difference and fixes various tests using the `stat` command to get file size.

### DIFF
--- a/qa/1280
+++ b/qa/1280
@@ -72,11 +72,7 @@ _doit
 
 echo
 echo "=== volume 2 truncated ==="
-size=`stat 20180416.2 | sed -n -e '/Size:/{
-s/.*Size: *//
-s/ .*//
-p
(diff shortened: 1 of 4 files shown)


kmcdonell (commented):
@kurik LGTM
The "nitpick" from @coderabbitai probably should be guarded in _filesize() not the qa test itself, but that's a bigger change to the structure of _filesize() which is outside the scope of this PR ... I'll put that on my TODO list.

PR #2614, opened 2026-06-09
zeroconf: drop dependency on pcp-pmda-bpftrace and bpftrace

natoscott:
The link to bpftrace brings in an additional 200MiB+ (Fedora) and 400MiB+ (RHEL) dependencies via clang (and gcc-toolset on RHEL) - which is simply too much to justify as a default dep.

Resolves Red Hat issue RHEL-180384

### DIFF
--- a/build/rpm/pcp.spec.in
+++ b/build/rpm/pcp.spec.in
@@ -2144,9 +2144,6 @@ Requires: pcp-pmda-dm = @package_version@
 %if "@pmda_bpf@" == "true"
 Requires: pcp-pmda-bpf = @package_version@
 %endif
-%if "@pmda_bpftrace@" == "true"
-Requires: pcp-pmda-bpftrace = @package_version@
-%endif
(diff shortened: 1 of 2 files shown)


kmcdonell (approved):
@natoscott So all of this stuff was pulled in and none of it is apparently used in the zeroconf versions of pmlogger or pmie or pmcd config?

Other than that observation, LGTM.

natoscott:
> @natoscott So all of this stuff was pulled in and none of it is apparently used in the zeroconf versions of pmlogger or pmie or pmcd config?
> 
> Other than that observation, LGTM.

That's right - it was installed so as to make at available when needed (on-demand). It's very difficult to get certain enterprise customers to update anything during an incident (or even afterward) that wasn't in the original system images.

However, the unexpectedly large on-disk footprint and the need to install a full-blown compiler (or two!) has been identified as problematic.

PR #2621, opened 2026-06-15
pmdadb2: new PMDA for IBM Db2 performance metrics

natoscott:
Adds a PMDA for collecting performance data from IBM Db2 databases using the ibm_db Python package and the Db2 MON_GET_* monitoring table functions.

Nine metric clusters (0-8) covering ~85 metrics across eight namespaces:

  db2.database.*   (27) database-wide aggregates: transactions, row counts,
                        sort/lock/deadlock/escalation/timeout statistics,
                        package cache efficiency, direct I/O, connections
  db2.bufferpool.* (14) per buffer pool: logical/physical/async reads and
                        writes for data, index and column-organised pages,
                        pool I/O times, unread prefetch pages, no-victim events
  db2.table.*      (10) per user table: row DML counts, lock waits/escalations/
                        deadlocks, overflow accesses, full table scans
  db2.workload.*    (8) per workload: row counts, CPU time, transactions,
                        lock waits
  db2.tablespace.* (14) per tablespace: buffer pool and async I/O for data
                        and index pages, direct read/write counts and times
  db2.connection.* (10) per active connection: CPU time, row counts, commits,
                        rollbacks, lock waits, deadlocks, escalations
  db2.latch.*       (2) per latch type: extended wait time and count
                        (MON_GET_LATCH, filtered to contended latches only)
  db2.log.*        (10) transaction log I/O: writes, reads, physical I/O
                        counts, buffer-full events, current used/available KB
  db2.pkg_cache.*   (6) package cache summary aggregate: cached statement
                        count, total executions, CPU time, rows read, sorts

Seven instance domains: buffer pools, tables (schema.table), workloads, tablespaces, connections, latches, and databases.

Multiple databases on a single Db2 instance are supported.  The [databases] section of the configuration file holds a space-separated list of database names.  One connection per database is maintained.  The database name is the instance for the db2.database.*, db2.log.* and db2.pkg_cache.* metrics; all per-object metrics use dbname::component instance naming.  A labels callback decomposes compound instance names into separate database= and object-type= label dimensions for use with time series queries.

All metrics and instance domains carry both one-line and long help text, including the source MON_GET_* column name.

Configuration is at $PCP_SYSCONF_DIR/db2/db2.conf (/etc/pcp/db2/) only.

Design notes:
- ibm_db is not packaged in major Linux distributions, so the PMDA is built unconditionally whenever Python is available (like pmdanetcheck/pmdanrabbitmq) rather than gating on a configure-time library check.  A graceful ImportError with a useful diagnostic is emitted at runtime if ibm_db is absent, and the Install script performs the same check before registering with pmcd.
- Each PCP cluster maps to exactly one MON_GET_* SQL call per configured database, fetching all columns for all instances in a single round-trip. Only clusters containing requested metrics are queried per fetch PDU.
- Reconnection after a Db2 instance restart is handled by probing with VALUES(1) in db2_cursor() rather than relying on ibm_db.active() alone. db2_exec() clears the connection handle on any query failure to ensure the next call triggers a full reconnect.

QA test 2002 exercises Install/Remove and validates key metrics against the first configured database, _notrun if prerequisites are absent.


natoscott:
Three coderabbit review comments were assessed but not actioned:

**GNUmakefile: `install -t` flag usage**

The `$(INSTALL)` macro in PCP's build system is not standard POSIX `install(1)` — it is PCP's own internal macro defined in `src/include/buildrules`. The form `$(INSTALL) -m mode -t dest_path src_file` installs a single file to a specific named destination path. This is the identical pattern used by every other PMDA in the tree (e.g. `src/pmdas/mssql/GNUmakefile`, `src/pmdas/postgresql/GNUmakefile`). No change needed.

**`db2_cursor`: unbounded reconnect loop**

The reconnect loop in `db2_cursor` is intentionally unbounded, matching the identical pattern in `pmdamssql`'s `mssql_cursor`. If a database remains down, the PMDA will keep retrying rather than giving up and permanently marking metrics unavailable. This is the correct behaviour for a long-running monitoring daemon — it should resume collecting as soon as the database recovers, without requiring a pmcd restart.

**`Remove`: missing `domain=42`**

`pmdaRemove` locates the PMDA registration by the `$iam` name, not by domain number. No other Python PMDA Remove script sets `domain=` — see `src/pmdas/mssql/Remove`, `src/pmdas/postgresql/Remove`, `src/pmdas/rabbitmq/Remove` etc. The domain variable is only required in `Install` (where it is set). No change needed.

src/pmdas/db2/GNUmakefile
```diff
@@ -0,0 +1,66 @@
...
+include $(BUILDRULES)
+
+ifeq "$(HAVE_PYTHON)" "true"
+build-me: check_domain
+install_pcp install: default
+	$(INSTALL) -m 755 -d $(PMDAADMDIR)
+	$(INSTALL) -m 755 -d $(PMDATMPDIR)
+	$(INSTALL) -m 755 -t $(PMDATMPDIR) Install Remove $(PYSCRIPT) $(PMDAADMDIR)
+	$(INSTALL) -m 755 -d $(PMDACONFIG)
+	$(INSTALL) -m 600 -t $(PMDACONFIG)/$(CONF) $(CONF)
```
kurik: IMO this (line 47) should be
```$(INSTALL) -m 600 -t $(PMDATMPDIR)/$(CONF) $(CONF) $(PMDACONFIG)/$(CONF)```
otherwise the `db2.conf` is missing after installation.

natoscott: @kurik yep, thanks - fixed now (slightly differently - since this is a new PMDA we do not need /var symlinks for an "existing" config file, and the PMDA is coded to only use the /etc location).

kurik:
@natoscott What version of DB2 you were using as the backend for this PMDA?

When I tried to make this working, using DB2 community edition v12.1, I realized the set of SQL commands and SQL functions used for the monitoring is different comparing to what the PMDA use.

natoscott:
@kurik  Its been built from first-principles using the docs (I don't have a spare 20-50Gb required to run the container on my laptop).  Can you try the latest updates from main which have several fixes & let me know what you're seeing in the logs?

Most accessors should be correct now, though I still have not confirmed STMTS_COMPLETED, TOT_LOG_USED_KB, TOT_LOG_AVAILABLE_KB, and NUM_EXECUTIONS.

Thanks!

PR #2624, opened 2026-06-15
Makepkgs: add --ci mode to reduce output volume, preserving error context

natoscott:
--verbose floods container CI platforms with enough output to trigger internal container runtime timeouts (libpod exit-file races etc.). Dropping --verbose silently loses the error context on build failure.

Add --ci mode as a middle ground: all output still goes to Logs/pcp in full, but only phase headers and error/failure lines are streamed to stdout in real time via a single _tee_log() helper.  On failure, build_failure() emits a filtered error summary and the last 100 lines of the log, which is far more useful than the previous 10-line tail.

Switch all CI platform configs and the macOS workflow to --ci.

### DIFF
--- a/Makepkgs
+++ b/Makepkgs
@@ -19,6 +19,7 @@
 topdir=`pwd`
 signed=false
 verbose=false
+cibuild=false
 srconly=false
 cfgonly=false
 dorpm=unknown
@@ -195,6 +196,10 @@ do
 	;;
     -verbose|--verbose)
 	verbose=true
+	shift
+	;;
+    -ci|--ci)
+	cibuild=true
 	shift
 	;;
     --check)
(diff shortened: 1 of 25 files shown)


Makepkgs
kmcdonell, line 1: The RE does introduce some false matches ... we have filenames that contain components like "failed" and "error", so shell for lines and install.sh lines seem to be the most likely false match cases ... but it is probably not worth worrying about these because smacking the false matches will only increase the risk of missing a real error/warning.

kmcdonell (approved):
LGTM, modulo false match comment below.

Makepkgs
natoscott: Yep, that matches my thinking as well.

Issue #2538, opened 2026-03-19
[RFE] Option to export timestamps with pcp2openmetrics in live mode too


natoscott:
There was a request on the mailing list relating to have pcp2openmetrics(1) export timestamps in live mode.  Currently thats only supported for archive mode, but a decent case was made for live mode also.

Since support is already there for archive mode (where this defaults to enabled), it seems like it'd be pretty straightforward to add a command line option allowing this in live mode too.

In the PMWEBAPI(1) /metrics endpoint, there is a "times=true/false" parameter which switches this on/off there - perhaps a mirror command line option in pcp2openmetrics would be the way to go?  (defaulting to on for archives and off for live mode for back-compat?)

natoscott:
OTOH maybe it should simply become the default for live mode too?

natoscott:
`I misunderstood the semantics of these timestamps in OpenMetrics format - they are not sample times. 
`

Reading further, I think we were correct all along but that this is something that changed in v2.0 - the v1.0 spec contains "An example with a Metric with no labels, and a MetricPoint with a timestamp and a created" which shows that our original mechanism for exposing sample timestamps was correct all along (even though we did not expose 'created' times):  https://prometheus.io/docs/specs/om/open_metrics_spec/#counter-1

In v2 passing sample timestamps this way is now explicitly not recommended, but it is not disallowed, and we have no other sensible way of providing meaningful timestamps (especially from archives).

Issue #1131, opened 2020-10-30
pmseries query to redis via pmproxy triggers assert in redisProcessCallBacks()


goodwinos:
```
$ pmseries -Dseries -p 44323 kernel.all.pswitch
...
pmseries: [Debug] Connection established
    metric.name
==
    "kernel.all.pswitch"
Baton [query/0x56315c415480] references: 0 -> 1 (@ series_query_services)
Baton [query/0x56315c415480] references: 1 -> 0 (@ series_query_end_phase[seriesPassBaton])
Baton [query/0x56315c415480] references: 0 -> 1 (@ series_query_maps)
Baton [query/0x56315c415480] references: 1 -> 0 (@ series_query_end_phase[seriesPassBaton])
Baton [query/0x56315c415480] references: 0 -> 1 (@ series_query_eval)
Baton [query/0x56315c415480] references: 1 -> 2 (@ series_prepare_expr[direct])
Redis [slot=16316] SMEMBERS pcp:series:metric.name:b23b0b8c0277ba9db64019494f6e51f549a6ee83
Redis [slot=16316] localhost:44323
Baton [query/0x56315c415480] references: 2 -> 1 (@ series_query_end_phase[seriesPassBaton])
CALLBACK redisProcessCallBacks: flags=0x0202 replies=0x56315c414ea0
|REPLY type=array elements=1
[0]||REPLY type=string len=20 SID d7832c4fba33bcc980b1a1b614e0508043288480
metric pcp:series:metric.name:b23b0b8c0277ba9db64019494f6e51f549a6ee83
    d7832c4fba33bcc980b1a1b614e0508043288480
Union of large(1) and small(0) series
Baton [query/0x56315c415480] references: 1 -> 0 (@ series_query_end_phase[seriesPassBaton])
Baton [query/0x56315c415480] references: 0 -> 1 (@ series_query_expr)
Baton [query/0x56315c415480] references: 1 -> 0 (@ series_query_end_phase[seriesPassBaton])
Baton [query/0x56315c415480] references: 0 -> 1 (@ series_query_mapping)
Redis [slot=08367] HGETALL pcp:map:metric.name
Redis [slot=08367] localhost:44323
CALLBACK redisProcessCallBacks: flags=0x0202 replies=0x56315c414ea0
CALLBACK redisProcessCallBacks: flags=0x0202 replies=0x56315c414ea0
CALLBACK redisProcessCallBacks: flags=0x0202 replies=0x56315c414ea0
[...]

PR #2637, opened 2026-06-25
pcp2openmetrics: add live timestamp mode

lmchilton:
update pcp2openmetrics tool to allow for live timestamp mode option. This update follows openmetric / prometheus exposing timestamp guidelines. Updated man page & QA #1131

Resolves issue #2538

### DIFF
--- a/src/pcp2openmetrics/pcp2openmetrics.1
+++ b/src/pcp2openmetrics/pcp2openmetrics.1
@@ -418,6 +418,9 @@ More than one
 .B \-K
 option may be used.
 .TP
+\fB\-l\fR, \fB\-\-times\fR
+Live timestamp mode. When configured metric values are accompanied by their respective timestamp
+.TP
 \fB\-L\fR, \fB\-\-local\-PMDA\fR
 Use a local context to collect metrics from DSO PMDAs on the local host
 without PMCD.
--- a/src/pcp2openmetrics/pcp2openmetrics.py
+++ b/src/pcp2openmetrics/pcp2openmetrics.py
@@ -55,7 +55,7 @@ class PCP2OPENMETRICS(object):
         # Configuration directives
         self.keys = ('source', 'output', 'derived', 'globals',
                      'samples', 'interval', 'precision', 'daemonize',
-                     'timefmt', 'everything',
+                     'timefmt', 'everything', 'times',
                      'count_scale', 'space_scale', 'time_scale', 'version',
                      'count_scale_force', 'space_scale_force', 'time_scale_force',
                      'precision_force', 'limit_filter', 'limit_filter_force',
@@ -111,6 +111,7 @@ class PCP2OPENMETRICS(object):
         self.space_scale_force = None
         self.time_scale = None
         self.time_scale_force = None
+        self.times = 0
 
         # Not in pcp2openmetrics.conf, won't overwrite
         self.outfile = None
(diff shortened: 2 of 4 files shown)


src/pcp2openmetrics/pcp2openmetrics.py
```diff
@@ -495,13 +504,23 @@ def openmetrics_labels(inst, name, desc, labels):
...
                     openmetrics_name_end = openmetrics_name(metric)
 
                 if self.context.type == PM_CONTEXT_ARCHIVE:
                     body += '%s%s %s %s\n' % (openmetrics_name_end, openmetrics_labels(inst, name, desc, labels), value, ts)
                 else:
                     body += '%s%s %s\n' % (openmetrics_name_end, openmetrics_labels(inst, name, desc, labels), value)
+                    if self.time_stamp and openmetrics_type(desc) == "counter":
+                        counter_key = (metric, inst)
+                        creation_ts = self.counter_timestamps.get(counter_key, ts)
+                        body += '%s%s %s\n' % (openmetrics_name_ts, openmetrics_labels(inst, name, desc, labels), creation_ts)
```
natoscott: I think its incorrect to expose timestamps for counter metrics *only* here - we should treat live mode the same as archive (a few lines above) in this regard, and export the timestamp for the sample (as per the v1 OM spec, and as we always have).

The "_created" lines are indeed intended for counters only, re-reading the spec(s).  They could be applicable to both archive and live PCP values, however, looking at our code - and in particular the fetchgroup code behind the python API - these are not going to be handled correctly (there is no knowledge of "mark" records in src/libpcp/src/fetchgroup.c which is the code being used by pmconfig.py and in turn by pcp2openmetrics.py).  We should ignore the _created timestamp, its not ever going to be correctly handled here.

Finally, the PMWEBAPI(3) uses the parameter name "times" to enable these timestamps - could we mirror that here?  i.e. instead of --time_stamp lets use --times?  (and this should remain the default in archive mode).

lmchilton: @natoscott thank you for the clarification! I updated everything

src/pcp2openmetrics/pcp2openmetrics.1
```diff
@@ -418,6 +418,9 @@ More than one
 .B \-K
 option may be used.
 .TP
+\fB\-l\fR, \fB\-\-times\fR
+Live timestamp mode. When configured counters are accompanied by their respective timestamp
```
natoscott: counters -> metric values

src/pcp2openmetrics/pcp2openmetrics.py
```diff
@@ -221,6 +223,8 @@ def option(self, opt, optarg, _index):
         """ Perform setup for individual command line option """
         if opt == 'daemonize':
             self.daemonize = 1
+        elif opt == 'l':
+            self.time_stamp = 1
```
natoscott: I agree with the rabbit - this would be clearer as self.times and not the previous option name.

src/pcp2openmetrics/pcp2openmetrics.py
```diff
@@ -498,7 +502,7 @@ def openmetrics_labels(inst, name, desc, labels):
                 else:
                     openmetrics_name_end = openmetrics_name(metric)
 
-                if self.context.type == PM_CONTEXT_ARCHIVE:
+                if self.context.type == PM_CONTEXT_ARCHIVE or self.time_stamp:
```
natoscott: It would be cleaner to set self.times to true when in archive mode earlier (outside the loop, so we don't need an "or" here) - there's a spot in execute() that might suit ...
```
        if self.context.type != PM_CONTEXT_ARCHIVE:
            self.delay = 1
            self.interpol = 1
        else:
            self.times = 1
```

natoscott (changes requested):

PR #2642, opened 2026-07-02
build: PCP_IMPORT_DIR needs to mirror PCP_RUN_DIR setup

natoscott:
Resolves issues Ken encountered with Debian builds.


natoscott:
Merged awhile back.

PR #2643, opened 2026-07-02
Miscellaneous work items

kmcdonell:
+ teach the darwin PMDA about extra units
+ some code tweaks for the darwin PMDA and help text
+ new -1 option for pminfo (pmDesc as tab-separated fields on a single line per metric)
+ assorted QA fixups, many for OpenBSD
+ fix some disk.{md,dm}.*_bytes metrics ... type should be U64 on 64-bit kernels


qa/group
kmcdonell, line 1: @coderabbitai this "group" file provides an inventory of all QA tests.  In this case the change to the lines for tests 1676 and 1677 removes the :reserved keyword and this means the files 1676, 1676.out (or similar), 1677 and 1677.out (or similar) must also exist in the qa directory ... this was not the case here, leading to build and packaging failures.  Similarly if a new line is added into "group" with a numeric first field and no :reserved or :retired tag, the corresponding test (from the first field) and the associated .out file should exist.
Failures here are usually the result of a missing git add for the new test script and validated output file, and is a recurring failure pattern.

PR #2648, opened 2026-07-07
local.conf and local.root PMDA sub-package dropins

natoscott:
RPM packaging changes to allow PMDAs with dependencies to be added/removed from the local-context-only setup through spec file %post / %preun scripting.


natoscott:
I don't think we should go down this path anymore, closing.

PR #2649, opened 2026-07-08
Split up pcp-system-tools pcp-atop/htop/dstat packaging

natoscott:
Now that pcp-atop, pcp-htop and pcp-dstat can operate completely independently to functionality from the pcp package (esp. pmcd now being optional for them), move these into separate sub-packages and allow operation either with/without pcp.


natoscott:
Merged after rebase.

Issue #2644, opened 2026-07-03
Harden PCP systemd services


orasagar:
### Issue Description

 PCP systemd units currently get high `systemd-analyze security` exposure scores.

  Observed:

  ```text
  pmcd.service                       9.7 UNSAFE
  pmie.service                       9.7 UNSAFE
  pmlogger.service                   9.7 UNSAFE
  pmproxy.service                    9.7 UNSAFE
  pmlogger_check.service             9.1 UNSAFE
  pmlogger_daily.service             9.1 UNSAFE
  pmlogger_daily_report.service      9.1 UNSAFE
  pmlogger_farm.service              9.1 UNSAFE
  pmlogger_farm_check.service        9.6 UNSAFE
  pmie_check.service                 9.1 UNSAFE
  pmie_daily.service                 9.1 UNSAFE
  pmie_farm.service                  9.1 UNSAFE
  pmie_farm_check.service            9.6 UNSAFE
  pmfind.service                     9.6 UNSAFE
```
  The units appear to lack common systemd hardening options such as NoNewPrivileges=, ProtectSystem=, ProtectHome=, PrivateTmp=, RestrictNamespaces=, RestrictSUIDSGID=,
  SystemCallArchitectures=, and a narrowed CapabilityBoundingSet=.

  Please review the PCP service files and add least-privilege hardening where possible, while preserving required host-metric collection behavior. For paths that need writes, use
  explicit allow-lists such as /var/lib/pcp, /var/log/pcp, and /run/pcp.

### Operating System

Oracle linux

### Architecture

x86_64

kmcdonell:
OK, this is all new to me ... but with not much effort I was able to reduce the pmlogger.service exposure score to `2.0 OK`.
I think the same settings could be used for everything except pmcd, pmproxy and maybe pmfind.
Do we need to aim lower than 2.0, or is that enough to get PCP's head below the parapet?

kmcdonell:
I'm not sure how useful any of this is in practice ... on bozo, changing UMask to 077 in the pmlogger.service file made NO DIFFERENCE to the mode of the created PCP archive files.

kmcdonell:
OK, I am now going to propose we DO NOTHING here.

Whatever the mechanisms this systemd service hardening is using,  they absolutely SUCK in terms of performance.

[...]

PR #2655, opened 2026-07-13
Harden PCP systemd units with minimal, service-only changes

kurik:
Add least-privilege directives to unit files only; omit SystemCallFilter, ProtectHostname on pmcd, and PrivateTmp on farm services to avoid performance impact and preserve existing behavior.

## Patch description

  This change hardens PCP systemd service units for issue #2644 
  (https://github.com/performancecopilot/pcp/issues/2644) by adding common least-privilege directives
  (ProtectHome, ProtectClock, ProtectKernel*, RestrictNamespaces, UMask=0027, etc.) to the main PCP services
  and their timer/check/farm companions.

  The patch is intentionally minimal in scope: it touches only systemd unit template files (*.service.in) and
  packaging metadata. There are no changes to application code, libraries, PMDAs, or rc scripts.
  SystemCallFilter is deliberately not used, based on upstream discussion that syscall filtering caused large
  QA slowdowns (e.g. qa/966 going from ~17s to ~89s).

###  Omitted hardening options

  Two options that would otherwise be straightforward to add were left out after test-suite analysis:

  **`ProtectHostname=yes` on `pmcd.service`**

  In normal production use, this is unlikely to matter much: pmcd reads the hostname at startup and most
  deployments do not change the system hostname at runtime.

  It does break hostname-following behavior that PCP relies on when the hostname changes. pmcd detects changes
  via `gethostname()` and signals clients with `PMCD_HOSTNAME_CHANGE`; pmlogger and pmie then restart to pick up
  the new name. With `ProtectHostname=yes`, pmcd sees a frozen hostname namespace and never propagates that
  change.

  Test impact: qa/1441 failed — after a hostname change, the primary pmlogger never restarted and still
  referenced the old hostname.

  Status: Omitted from pmcd.service with an inline comment. This is a reasonable hardening candidate later, but
  would need either a different pmcd hostname-detection approach or test rework.

   **`PrivateTmp=yes` on` pmie_farm.service` and `pmlogger_farm.service`**

  For typical deployments, farm instances write logs under `/var/log/pcp/.`.. from control files, so `PrivateTmp`
  may have little practical effect.

  It does break non-primary farm instances that use log paths outside the service’s private `/tmp` — including
  user-configured locations and QA tests that deliberately use `/tmp/<test-pid>/....`

  Test impact: qa/1060 failed — the non-primary pmie instance could not create its nested log tree under
  /tmp/1060-*.

  Status: Omitted from both farm service units with an inline comment. Low-hanging fruit for a future hardening
  pass, but farm log paths are user-configurable, so proper validation would need more than a quick unit-file
  change.

###   What was added

  Hardening was applied across pmcd, pmproxy, pmlogger, pmie, pmfind, atop, geolocate, pmseries, and their
  check/daily/farm companion units, with service-specific exceptions documented in comments (e.g.
  `NoNewPrivileges` omitted on pmcd because it drops to the pcp user via setuid; `NotifyAccess=main` on pmproxy).

###  Rationale for deferring `ProtectHostname` and `PrivateTmp`

  Both options are attractive from a `systemd-analyze security` perspective and would likely be fine in many
  production setups. They were not included in this iteration because:

  1. They have clear functional interactions with hostname following and farm log paths.
  2. Failures showed up only in targeted QA tests, not in obvious production breakage — but that is not
     enough confidence without broader validation.
  3. Properly addressing them would require additional testing and possible QA/test rework, which is out of
     scope for this minimal, time-constrained patch.

  The goal here is a small, low-risk hardening step that improves exposure scores without touching runtime code
  or risking the performance regressions seen with heavier sandboxing. `ProtectHostname` and `PrivateTmp` remain
  documented candidates for a follow-up once there is time to validate behavior and adjust tests as needed.


kmcdonell:
@kurik are you proposing this for the 7.1.6 release? 
I'm nervous about this change late in the release cycle, so NOT :cowboy_hat_face:  or :clown_face: this time.

Do you have empirical data to show that there's no significant performance impact here (the syscall assertion in issue #2644 is speculation, I don't have any data (yet) to prove this is the root cause of the slowdown I observed in my parallel efforts down this path).

Of course, I have no visibility into any firm customer-pressure to address this quickly which would be a non-technical counter argument. :smile:

kurik:
> @kurik are you proposing this for the 7.1.6 release? I'm nervous about this change late in the release cycle, so NOT 🤠 or 🤡 this time.

No, I am not proposing this for  7.1.6 release. It still might need some work as there are some differences on Debian (as I see in the CI).

> Do you have empirical data to show that there's no significant performance impact here (the syscall assertion in issue #2644 is speculation, I don't have any data (yet) to prove this is the root cause of the slowdown I observed in my parallel efforts down this path).

I do not have the data anymore, but I can easily generate it. I was using `check -TT` to get timestamps and then I compared the duration for each test. The result was +/- 1s per test, while the overall duration was 2s faster when run with the hardening. So it is IMO within natural fluctuation. I will keep the data from the next run when I investigate the Debian failures.

> Of course, I have no visibility into any firm customer-pressure to address this quickly which would be a non-technical counter argument. 😄

There is no pressure on my side.

kmcdonell:
Excellent @kurik, let's try and nail this down as early as possible in the next release cycle.  So I don't forget, tasks I can see are:

1. Compare my service changes with yours and focus on the differences
2. I think I have QA fixes for most of the PrivateTmp fallout ... this is only needed for QA AFAICT and we can work around it there, so I suggest that one can go back in the mix
3. Some controlled timing experiments to see what the overhead really is.
4. Determine if the final exposure score for all the services is acceptable.

kurik:
Hi @kmcdonell ,
finally I have some minimal set of systemd hardening options, which work even on Ubuntu :-) ( I typically do not test on Ubuntu, so this takes me more time... ).

As part of this exercise I tried to compare performance with and without this hardening. At first, I run all tests in the `local` group without the hardening and considered this as a baseline. Then I did two runs of the same  test set with the hardening enabled. Here is the outcome:

### Average performance impact
* Mean per-test delta: **+0.026 s** (effectively zero at 1 s resolution)

### Largest single spike (not reproducible):
* Test 1108: 1 s → 52 s in hardening run 2 only (+51 s). Run 1 was 1 s. This is IMO some pmlogextract/cron noise, not a hardening effect.

### Largest reproducible slowdown (both hardened runs):
*  Test 1845: 1 s → ~10.5 s (+9.5 s) — pmlogger_daily for pmproxy-pushed archives; the only pmproxy-related case with a consistent delta
* Test 1662: 19 s → 22.5 s (+3.5 s) — pmproxy valgrind leak test
* Six tests total reproducibly slower: +21.5 s combined

### Reproducible speedups (both hardened runs faster):
* Tests 061 (−24.5 s), 1614 (−18.5 s, failed→passed), 1626 (−11.5 s) — net −54.5 s on those three.

### Conclusion:
The data does not support a measurable suite-wide performance cost. The only plausibly hardening-related reproducible case is 1845.

### One more comment:
I haven't yet investigated why the success rate is higher with hardening enabled compared to the baseline run. I know some tests are unstable, but this still warrants a closer look when I have some spare time:
| Run | Passed | Failed | Not run |
|-----|--------|--------|---------|
| No hardening | 1,492 | 14 | 193 |
| Hardening 1 | 1,497 | 11 | 191 |
| Hardening 2 | 1,501 | 7 | 191 |

kurik:
> Compare my service changes with yours and focus on the differences

That would be great. What I learned during this exercise is the fact there might be some side effects one (at least me :-) ) do not expect. Specifically I mean `RemoveIPC` when the removal is based on UID/GID, so if you have two services (i.e. `pmcd` and `pmie`) where only one service uses IPC (the `pmcd`) then restart of the second service (`pmie`) removes all IPC artifacts owned by the same UID/GID (shared memory, semaphores, ...) just because it is run as the same UID/GID as the first service.

As such, I definitely agree with you to merge these changes early in the release cycle, so there is as much time as possible to catch possible issues.

> I think I have QA fixes for most of the PrivateTmp fallout ... this is only needed for QA AFAICT and we can work around it there, so I suggest that one can go back in the mix

This is great. I tried to come up with a minimal set of hardening that wouldn't even affect the test suite. However, it seems that settings like `PrivateTmp` don't affect individual services; they only affect specific tests.

> Some controlled timing experiments to see what the overhead really is.
> Determine if the final exposure score for all the services is acceptable.

I am not sure how much we can have some controlled environment when nowadays is "everything" virtual, even the hardware :-)
I am typically using [TestingFarm](https://docs.testing-farm.io/Testing%20Farm/0.1/test-request.html) for my test environment, which offers semi-transparent access to AWS (for x86_64) or[ IBM Cloud](https://cloud.ibm.com/docs/vpc?topic=vpc-creating-virtual-servers&interface=ui) (s390x) etc. My point of view would be rather to have more test runs with a tested feature disabled and then with the feature enabled and do some statistic on top of this (averages, spikes, the most significant deviations, etc.) However as you have access to bare metal (bozo machines?) we can do both.

orasagar:
@kurik great work on this.
I have tested it and it seems to be promising ,just tested it for systemd score not performance mesaurement

PR #2657, opened 2026-07-17
selinux: add pcp-import policy module for local context recording tools

natoscott:
## Summary

- New SELinux policy module `pcp-import` for PCP tools operating in `PM_CONTEXT_LOCAL` mode without pmcd (pcp-atop, sysstat sadc with `-O pcp`)
- Single reusable `pcp_import_t` domain — future import tools just add their binary path to the `.fc` file
- Augments existing `sysstat_t` domain via `optional_policy` when sysstat is installed (no sysstat packaging changes needed)
- Works independently of `pcp-selinux` (which covers pmcd/pmlogger/pmproxy/pmie)
- New `pcp-selinux-import` RPM subpackage, weak-required by `pcp-libs` when SELinux is active
- Rename `PCP_IMPORT_DIR` → `PCP_IMPORTRUN_DIR` (runtime status, `/run/pmimport`)
- New `PCP_IMPORTLOG_DIR` (`/var/log`, base directory for import tool archives)
- pcp-atop uses `$PCP_IMPORTLOG_DIR/atop` for daily archives

🤖 Generated with [Claude Code](https://claude.com/claude-code)


natoscott:
@coderabbitai The `%{_rundir}/pmimport` in `redhat.spec` line 2498 is correct as-is. This is the standard RPM macro for `/run` and matches the resolved value of `@pcp_importrun_dir@` in `pcp.spec.in`. The `redhat.spec` doesn't use autoconf substitution — it uses RPM macros directly, which is the established pattern throughout that file. The suggestion to use `%{_localstatedir}/run/pmimport` would actually be worse (resolves to `/var/run/pmimport` which is a legacy compat symlink on modern systems).

All CI checks are now passing. The `kernel_read_psi` dummy interface, boolean dependency version guards, and AVC fixes for DSO PMDA temp files and socket access have all been folded into the single commit.

PR #2665, opened 2026-07-25
pmdalmsensors: query only the requested chip during metric fetch

Ali-Nasrolahi:
# PMDA lmsensors: Reduce redundant sensor probing during metric fetches

## Patch Description

This patch optimizes the `lmsensors` PMDA by avoiding redundant full-system sensor scans when serving metric fetch
requests.  Currently, every metric fetch executes `sensors -j` which probes **all** sensor chips, even though only a
single metric value is ultimately returned. As a result, a PCP fetch involving multiple metrics repeatedly re-reads the
same hardware.  This patch records the originating lm_sensors chip for each metric during PMDA initialization and uses
that information to query only the chip associated with the requested metric.

In other words, instead of executing `sensors -j`, we run `sensors -j <chip>` which reduces the amount of work that is
needed.

Additionally, the cost with current implementation grows with:

- the number of sensor chips present,
- the latency of individual sensor backends,
- and the number of metrics requested by PCP clients.

## Related Issue

While investigating why the PMDA was occasionally timing out, I found that one sensor backend on my system requires
roughly half a second to respond. Because the PMDA executes a full `sensors -j` invocation for every metric, the latency
combined across all requested metrics, which then exceeded the PMCD request timeout and causing all metrics in the
request to fail. Since only one metric value is needed for each callback, querying only the corresponding sensor chip
avoids repeatedly waiting on unrelated hardware.

For the context, my environment is Fedora 44 with PCP 7.1.5:
```bash
❯ rpm -q pcp pcp-pmda-lmsensors
pcp-7.1.5-3.fc44.x86_64
pcp-pmda-lmsensors-7.1.5-3.fc44.x86_64
```

Warm Regards,
Ali :)

### DIFF
--- a/src/pmdas/lmsensors/pmdalmsensors.python
+++ b/src/pmdas/lmsensors/pmdalmsensors.python
@@ -32,9 +32,10 @@ import cpmapi as c_api
 
 sensorvalues = {}   # lookup sensorname -> sensorvalue (temp/fan)
 sensornames = {}    # lookup sensornumber -> sensorname
+sensorchips = {}    # lookup sensorname -> original lm_sensors chip name
 basename = "lmsensors"
 
-def lmsensors_get():
+def lmsensors_get(chip=None):
     ''' read sensor data '''
 
     if args.inject:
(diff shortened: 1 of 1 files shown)


natoscott:
@christianhorn PTAL.  also, does qa/1480 need updates along with these changes?

christianhorn:
LGTM, qa/1480 should not be affected, and no new tests making sense coming to mind.

Good spot, the fix is very welcome, @Ali-Nasrolahi ! I did a quick comparison, the win in syscalls and execution time, of executing the more specific 'sensors -j <chip>' is clear. This is also great for example on Asahi Macs where one uses a lmsensors-metric to regularly query the current power consumption. I'm doing these queries to have consumption shown in my window manager, so these cycles really add up over time..

christianhorn (approved):

Issue #2654, opened 2026-07-13
[RFE] Reimplement pmsearch(1) to be standalone


natoscott:
### Description of Request for Enhancement

Currently we use RediSearch full text search module in the backend of pmseries and the /search endpoint in the PMWEBAPI (as used by grafana-pcp).

However, it's got numerous problems:
- we no longer use Redis, and this module does not work with Valkey.  Its was also only packaged on Fedora.
- the Valkey equivalent supports most of the functionality we used, but it has a complex build system and complex dependencies like gRPC (difficult to package in Linux distros - its usually vendored, which creates additional problems).  Its not packaged on Fedora/RHEL.
- using Valkey/Redis here is complete overkill for our needs

We are indexing tiny amounts of text compared to what these enterprise-grade systems are designed for.  We can simplify things greatly and achieve 99% effectiveness with 1% of the code I suspect, and also build an always-available solution as opposed to the current messy situation, where its rarely available.

Bear in mind we index only the following info:  metric names, oneline and long help text, for both indoms and metrics.

The search module complexity comes from multiple places - stemming and tokenization are language specific.  Redis/ValkeySearch use the Snowball library to solve the stemming problem, not sure what they do for tokenisation (but that's far simpler).  Enterprise grade solutions typically default to the BM25 document ranking algorithm - https://www.geeksforgeeks.org/nlp/what-is-bm25-best-matching-25-algorithm/ - however, for us its total overkill and a simpler solution like TF.IDF would work just as well - https://www.geeksforgeeks.org/machine-learning/understanding-tf-idf-term-frequency-inverse-document-frequency/

If we can arrange for an inverted index to be built from all help text files available at build time (perhaps newhelp could be extended to do this), we'd be most of the way there.  This would need a simple tokeniser and stemmer - I think we could build our own here, the help text is very simple (scanning through  pminfo -Tt, just lower-casing everything and chopping plurals would cover the vast majority of word normalisation cases in our metrics/text corpus).

All uses (pmsearch, PMWEBAPI and grafana-pcp) are hidden by a libpcp_web API already, so we could simply replace the implementation behind that hopefully.

This topic has come up in the context of improving AI/MCP use of PCP.

### Component

pmsearch pmproxy grafana-pcp

sfeifer:
@natoscott Thanks for putting this together. Some topics for discussion:

[...]

PR #2675, opened 2026-07-31
Refactor pmsearch to use sqlite instead of RediSearch

sfeifer:
As described in #2654 , we want to reimplement pmsearch to no longer depend on RediSearch and instead depend on sqlite, a much more common package.


src/newhelp/GNUmakefile
```diff
@@ -18,8 +18,9 @@ include	$(TOPDIR)/src/include/builddefs
 include	$(TOPDIR)/src/libpcp/src/GNUlibrarydefs
 
 TARGETS = newhelp$(EXECSUFFIX) chkhelp$(EXECSUFFIX)
-CFILES = newhelp.c chkhelp.c
+CFILES = newhelp.c chkhelp.c search_sqlite.c
 
+LCFLAGS += $(SQLITE3CFLAGS)
```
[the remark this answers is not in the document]
sfeifer: The GNUmakefile is correct as-is. If sqlite-devel isn't installed, the build fails as expected when missing a required dependency.

qa/admin/other-packages/manifest
```diff
@@ -1018,6 +1018,17 @@ S_pkg?		/usr/lib/*/libsqlite3.so	[database/sqlite-3 (QA optional)]
...
+dpkg?		/usr/include/sqlite3.h		[libsqlite3-dev (build optional)]
+rpm?		/usr/include/sqlite3.h		[sqlite-devel (build optional)]
+emerge?		/usr/include/sqlite3.h		[dev-db/sqlite (build optional)]
+pkgin?		/usr/pkg/include/sqlite3.h	[sqlite3 (build optional)]
+pkg_add?	/usr/local/include/sqlite3.h	[sqlite3 (build optional)]
+F_pkg?		/usr/local/include/sqlite3.h	[sqlite3 (build optional)]
+S_pkg?		/usr/include/sqlite3.h		[database/sqlite-3 (build optional)]
+slackpkg?	/usr/include/sqlite3.h		[sqlite (build optional)]
+pacman?		/usr/include/sqlite3.h		[core/sqlite (build optional)]
+brew?		/usr/local/include/sqlite3.h	[sqlite (build optional)]
```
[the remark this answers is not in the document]
sfeifer: pkg-config resolves paths at configure time regardless of what the manifest lists.

sfeifer:
@natoscott Ready for review :). There are a few small changes left (merge conflict for qa test number and coderabbit feedback), but you should be able to build and test the branch as is.

natoscott:
@sfeifer I ran "./new -r" to reserve test 1861 for you.

I'm flatout like a lizard drinking this week though - might get some time on Friday for a look, else next week - hopefully Ken & others have more cycles before then to get you some feedback.

man/man1/pmsearch_index.1
```diff
@@ -0,0 +1,120 @@
...
+.\"
+.\" This program is distributed in the hope that it will be useful, but
+.\" WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+.\" or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+.\" for more details.
+.\"
+.\"
+.TH PMSEARCH_INDEX 1 "PCP" "Performance Co-Pilot"
+.SH NAME
+\f3pmsearch_index\f1 \- rebuild the PCP full-text search index
```
natoscott: @sfeifer could this be named pmsearch_daily?  In support of all our other dailys...
```
$ rpm -ql pcp|grep _daily
/usr/lib/systemd/system/pmie_daily.service
/usr/lib/systemd/system/pmie_daily.timer
/usr/lib/systemd/system/pmlogger_daily.service
/usr/lib/systemd/system/pmlogger_daily.timer
/usr/lib/systemd/system/pmproxy_daily.service
/usr/lib/systemd/system/pmproxy_daily.timer
/usr/libexec/pcp/bin/pmie_daily
/usr/libexec/pcp/bin/pmlogger_daily
```

man/man1/pmsearch_index.1
```diff
@@ -0,0 +1,120 @@
...
+It queries the running
+.BR pmcd (1)
+(or a PCP archive when
+.B \-a
+is specified)
+for all metric names, help text, and instance domain members,
+then builds a SQLite FTS5 search index using
+.BR newhelp (1).
+The resulting index file is written to
+.I $PCP_VAR_DIR/lib/pcp.search
```
natoscott: And this full path has "pcp" in it twice - could we name it pmsearch.index or search.index or pmsearch.sqlite ... ?

natoscott (changes requested):

qa/1861
```diff
@@ -1,69 +1,106 @@
...
+cp $tmp.share/lib/pmsearch.index $tmp.var/lib/pmsearch.index
+echo "Nightly index initialized from base"
+
+export PCP_VAR_DIR=$tmp.var
+export PCP_SHARE_DIR=$tmp.share
+
+# --- Step 3: Verify Python PMDA metrics are NOT in base index ---
+echo
+echo "=== Verify Python PMDA metrics absent from base index ==="
+pmsearch -C "qa_python"
```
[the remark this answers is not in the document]
sfeifer: If pmsearch outputs an error with a $tmp path, we'd want the test to fail so we catch the problem — not silently filter it away.

man/man1/pmsearch_index.1
sfeifer: Changed to pmsearch.index

man/man1/pmsearch_index.1
sfeifer: Good catch. It has now been updated.

natoscott:
@sfeifer the build fails on systems with no sqlite...
```
checking for sqlite3 >= 3.9.0... no
checking for sqlite3_open in -lsqlite3... no
configure: error: SQLite >= 3.9.0 with FTS5 support is required
```

this should not be a hard error, just warn and keep going.

natoscott:
@sfeifer its working pretty well! (haven't gone through the code, just from using it).  I suspect the indom help text is not being indexed though?  Try "NUMA" as search term - this should hit multiple indom help text entries in the Linux kernel PMDA, but all I see are metric names and metric help text?  If there's a priority/ranking option - rank indom help text highly when indexing, any hit there is super important and very likely relevant (there's also far fewer of these docs than the metric help text, so any kind of boost we can give indom help would be good).

sfeifer:
@natoscott Thanks for taking a look!

> @sfeifer the build fails on systems with no sqlite...
> 
> ```
> checking for sqlite3 >= 3.9.0... no
> checking for sqlite3_open in -lsqlite3... no
> configure: error: SQLite >= 3.9.0 with FTS5 support is required
> ```
> 
> this should not be a hard error, just warn and keep going.

Should this be a warning instead of an error for rpm builds too, or just when building from source on systems without sqlite installed?

> @sfeifer its working pretty well! (haven't gone through the code, just from using it). I suspect the indom help text is not being indexed though? Try "NUMA" as search term - this should hit multiple indom help text entries in the Linux kernel PMDA, but all I see are metric names and metric help text? If there's a priority/ranking option - rank indom help text highly when indexing, any hit there is super important and very likely relevant (there's also far fewer of these docs than the metric help text, so any kind of boost we can give indom help would be good).

I'm still working through the kinks locally, but I added to the SQL query in search_query_table in search.c to multiple the score by 1.5 for indom help text. This was just my first attempt and I'm thinking of increasing that multiple to as high as I can without causing totally bad results.

PR #2676, opened 2026-07-31
pcp2openmetrics & pcp2opentelemetry revisions

lmchilton:
openscanhub revisions for pcp2openmetrics & pcp2opentelemetry 
Updated error in format for unit assignment in pcp2opentelemetry Updated http authentication

### DIFF
--- a/src/pcp2openmetrics/pcp2openmetrics.py
+++ b/src/pcp2openmetrics/pcp2openmetrics.py
@@ -512,6 +512,10 @@ class PCP2OPENMETRICS(object):
         if self.url:
             auth = None
             if self.http_user and self.http_pass:
+                if not self.url.lower().startswith("https://"):
+                    raise ValueError(
+                        "HTTP Basic Authentication requires an HTTPS connection to prevent leaking credentials."
+                    )
                 auth = requests.auth.HTTPBasicAuth(self.http_user, self.http_pass)
             try:
                 timeout = self.http_timeout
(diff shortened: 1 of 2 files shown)


src/pcp2openmetrics/pcp2openmetrics.py
```diff
@@ -512,6 +512,10 @@ def openmetrics_labels(inst, name, desc, labels):
         if self.url:
             auth = None
             if self.http_user and self.http_pass:
+                if not self.url.lower().startswith("https://"):
+                    raise ValueError(
```
natoscott: This could break existing setups - should this just be a warning not an error?

src/pcp2opentelemetry/pcp2opentelemetry.py
```diff
@@ -531,14 +531,16 @@ def UCUM_count_prefix(scale):
                 if units.dimSpace != 1:
                     ucum_string += "%d" % units.dimSpace
                 first_unit = 0
-            elif units.dimTime:
+
+            if units.dimTime:
```
natoscott, line 535: Do these changes impact on tests?  If not, why not?  :)

lmchilton: @natoscott Looks like its not picked up in the QA test because right now I only have the metric "hinv.ncpu" being tested which is unitless. Should I add a metric or two that has a unit into the QA test?

natoscott: That'd be great.  The sample PMDA has heaps of options and is always available in QA, and the metrics are the same for every platform.

PR #2683, opened 2026-08-06
Refactor _service use in scripts

kmcdonell:
1. We had this sort of use pattern (with many, many variants):
     ```if ! _service ... | _some_filter; then _exit 1; fi```
     which is 100% broken because the _exit 1 only happens if _some_filter has non-zero exit, and the return value from _service is ignored.
    This should be something like:
    ```if ! _service >$tmp.tmp 2>&1; then cat $tmp.tmp; _exit 1; fi```
    ```_some_filter <$tmp.tmp```
2. rework qa/check-group as a by product of checking scripts after these changes
3. fixes to qa/group as a result of running qa/check-group


kmcdonell:
OK, I'll split this into 2 PRs so we get @coderabbitai reviews.

PR #2691, opened 2026-08-21
pcp-tools: correct None handling and CPU utilization

orasagar:
Handle None values safely in mpstat printers instead of calling replace() on None. Fix per-CPU utilization scaling so aggregate CPU values are divided by the total CPU count only when reporting the aggregate.

Update interrupt usage tests to use the new computation context and explicit interrupt report type.fix(mpstat): correct None handling and CPU utilization

Handle None values safely in mpstat printers instead of calling replace() on None. Fix per-CPU utilization scaling so aggregate CPU values are divided by the total CPU count only when reporting the aggregate.

Update interrupt usage tests to use the new computation context and explicit interrupt report type.


natoscott:
@orasagar this seems to have regressed tests 883 and 1782 - PTAL, thanks!

orasagar:
@natoscott raised [PR 2699](https://github.com/performancecopilot/pcp/pull/2699) to fix the issue

PR #2692, opened 2026-08-21
pcp-ps: optimize initial reporting and archive samples

orasagar:
Print non-CPU reports from the first sample, while retaining the
previous-sample wait for %CPU output. Compute TIME from current user
and system CPU time, exit live one-shot reports immediately, and add
an archive warm-up fetch so -s N prints N reports.

### DIFF
--- a/qa/1987.out
+++ b/qa/1987.out
@@ -3,6 +3,262 @@ QA output created by 1987
 pcp-ps output : Display default output
 Linux  5.4.17-2136.313.6.el8uek.x86_64  (localhost.localdomain)  12/14/22  x86_64    (4 CPU)
 Timestamp        PID		TIME		CMD
+06:20:30         1       	00:00:01	systemd             
+06:20:30         2       	00:00:00	kthreadd            
+06:20:30         3       	00:00:00	rcu_gp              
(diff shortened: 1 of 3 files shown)


orasagar:
This fixes two `pcp ps` sample-handling problems.

For normal live `pcp ps`, we no longer wait for a previous sample before printing. The default report only needs current process values, so it can print straight away. On the OL10 test VM, runtime dropped from 2.19s to 0.15s.

`pcp ps -u` and custom reports containing `%cpu` still wait for a previous sample. They need two samples to calculate a CPU rate, and they complete in about 1.19s.

`TIME` now comes from the current `utime + stime` values. Previously it mixed current system time with user time from the previous sample.

Archive reports need one warm-up record for CPU-rate output. We now fetch that record internally, so `-s N` prints N reports:

```text
pcp -a archive -s 4 ps -u
```

natoscott:
@orasagar please update as needed based on the several coderabbit review comments.

orasagar:
@natoscott I have resolved the comments

natoscott:
@orasagar qa/1987 seems to be failing with these changes.

orasagar:
@natoscott update the 1987 test case file

PR #2698, opened 2026-08-26
pmproxy: release the X509 reference in the CERT_REQD check

### LINKED ISSUE #2693: pmproxy: X509 reference leak in on_headers_complete() CERT_REQD check (81a9efe96db6)
Commit [`81a9efe96db6`](https://github.com/performancecopilot/pcp/commit/81a9efe96db6f7764a1a1646b64eb85de5cdaf77)
("pmproxy: enforce -Q (CERT_REQD) for REST API connections") calls
`SSL_get_peer_certificate()` for its NULL test and discards the return value:

```c
    /* client certificate required for all servlets */
    if (__pmServerHasFeature(PM_SERVER_FEATURE_CERT_REQD)) {
#ifdef HAVE_OPENSSL
	if (!client->stream.secure ||
	    !client->secure.ssl ||
	    SSL_get_peer_certificate(client->secure.ssl) == NULL) {
	    client->u.http.parser.status_code = HTTP_STATUS_FORBIDDEN;
	}
#else
```

`SSL_get_peer_certificate()` returns a certificate whose reference count has been
incremented — the caller owns that reference and must release it with `X509_free()`.
On OpenSSL 3.x the name is a compatibility macro for `SSL_get1_peer_certificate()`,
which has the same owning semantics:

```
/usr/include/openssl/ssl.h:
#   define SSL_get_peer_certificate SSL_get1_peer_certificate
```

The non-owning accessor is `SSL_get0_peer_certificate()`.

### Impact

Because of the short-circuit order, the leak happens on the **success** path. If the
connection is not TLS, or there is no SSL object, the call is never reached. If the call
returns NULL the request is rejected and there is nothing to free. The reference is
leaked precisely when a client certificate *is* present — i.e. when the request is
allowed through.

`on_headers_complete()` runs once per request, so a long-lived `pmproxy` started with
`-Q` accumulates one leaked `X509` per authenticated REST request. Typical dashboard
polling produces thousands of requests per hour, so this grows steadily over the
lifetime of the daemon.

This requires `-Q` (`PM_SERVER_FEATURE_CERT_REQD`), which is not enabled by the shipped
`pmproxy.conf`, so it only affects deployments that have opted into client-certificate
enforcement.

### Confirmation that nothing else releases it

In the 6.3.7 tree this commit introduces the only call to `SSL_get_peer_certificate()`
anywhere under `src/`, and there is no `X509_free()` call anywhere in the tree:

```
$ grep -rn "SSL_get_peer_certificate\|SSL_get1_peer_certificate" src/   # only the new hunk
$ grep -rn "X509_free" src/                                            # no matches
```

### Suggested fix

Keep the accept/reject logic exactly as-is and release the reference. The existing
short-circuit order matters — `SSL_get_peer_certificate()` must not be called when
`client->secure.ssl` is NULL:

```c
    /* client certificate required for all servlets */
    if (__pmServerHasFeature(PM_SERVER_FEATURE_CERT_REQD)) {
#ifdef HAVE_OPENSSL
	if (!client->stream.secure || !client->secure.ssl) {
	    client->u.http.parser.status_code = HTTP_STATUS_FORBIDDEN;
	} else {
	    X509	*cert = SSL_get_peer_certificate(client->secure.ssl);

	    if (cert == NULL)
		client->u.http.parser.status_code = HTTP_STATUS_FORBIDDEN;
	    else
		X509_free(cert);
	}
#else
	/* no TLS support compiled in, reject all connections */
	client->u.http.parser.status_code = HTTP_STATUS_FORBIDDEN;
#endif
    }
```

On OpenSSL 3.0+ only, `SSL_get0_peer_certificate()` would avoid taking a reference at
all, but it is unavailable on the older OpenSSL versions pcp still supports, so the
`X509_free()` form above is the portable fix.

### Versions

Found while backporting `81a9efe96db6` to pcp 6.3.7 and 6.0.1 for AlmaLinux 9.6 / 9.2
(OpenSSL 3.2.2). Present on `main` and in distro backports of this commit, including
Red Hat's `pcp-6.3.7-pmproxy-rest-certreqd.patch`.

wtatarski:
`SSL_get_peer_certificate()` returns a certificate whose reference count has been
incremented, so the caller owns that reference and must release it with `X509_free()`.
On OpenSSL 3.x the name is a compatibility macro for `SSL_get1_peer_certificate()`,
which has the same owning semantics.

81a9efe96db6 ("pmproxy: enforce -Q (CERT_REQD) for REST API connections") called it
inline in the NULL test and discarded the result:

```c
	if (!client->stream.secure ||
	    !client->secure.ssl ||
	    SSL_get_peer_certificate(client->secure.ssl) == NULL) {
	    client->u.http.parser.status_code = HTTP_STATUS_FORBIDDEN;
	}
```

Because of the short-circuit order the reference is only taken when the connection is
TLS and an SSL object exists, and it is leaked precisely when a certificate **is**
present — that is, on the path where the request is allowed through. The rejection paths
do not leak. `on_headers_complete()` runs once per request, so a `pmproxy` started with
`-Q` accumulates one `X509` per authenticated REST request for the lifetime of the
daemon.

This only affects deployments that have opted into client-certificate enforcement; `-Q`
is not enabled by the shipped `pmproxy.conf`. It is a reliability defect rather than a
security issue, since reaching it requires a client certificate that already passed
validation.

The fix captures the reference and frees it, leaving the accept/reject logic unchanged.
`SSL_get0_peer_certificate()` would avoid taking a reference at all, but it is not
available on the older OpenSSL versions pcp still supports, so `X509_free()` is the
portable form.

Found while backporting 81a9efe96db6 to pcp 6.3.7 and 6.0.1 (OpenSSL 3.2.2). Verified
against current `main` (55365f2e): the block is unchanged there, this commit introduces
the only call to `SSL_get_peer_certificate()` under `src/`, and there is no other
`X509_free()` in the tree — so nothing else was releasing it.

## Related Issues

Fixes #2693

### DIFF
--- a/src/pmproxy/src/http.c
+++ b/src/pmproxy/src/http.c
@@ -1132,10 +1132,15 @@ on_headers_complete(http_parser *request)
     /* client certificate required for all servlets */
     if (__pmServerHasFeature(PM_SERVER_FEATURE_CERT_REQD)) {
 #ifdef HAVE_OPENSSL
-	if (!client->stream.secure ||
-	    !client->secure.ssl ||
-	    SSL_get_peer_certificate(client->secure.ssl) == NULL) {
+	if (!client->stream.secure || !client->secure.ssl) {
 	    client->u.http.parser.status_code = HTTP_STATUS_FORBIDDEN;
+	} else {
+	    X509	*cert = SSL_get_peer_certificate(client->secure.ssl);
+
+	    if (cert == NULL)
+		client->u.http.parser.status_code = HTTP_STATUS_FORBIDDEN;
+	    else
+		X509_free(cert);
 	}
 #else
 	/* no TLS support compiled in, reject all connections */


natoscott:
@wtatarski thanks!

PR #2703, opened 2026-08-31
macos: fix libpcp install_name for qmake-built Qt tools

lukas7811:
pmchart, pmdumptext, pmtime, pmquery, pmgadgets and pmview are linked directly by qmake's generated Makefile via the QTMAKE build-me recipe, which bypasses the $(PROGTARGET)/$(LIBTARGET) rules in buildrules that invoke LINKER_MAKERULE after linking. On macOS that rule is what rewrites a binary's bare "libpcp.N.dylib" dependency into an absolute $(PCP_LIB_DIR) path, so without it these Qt tools keep the bare name and dyld fails to load them outside the build tree.

Add QT_LINKER_FIXUP, the same fixup parameterized on an explicit binary path (build-me has no mtime-based prerequisites, so $@ isn't available and the recipe can legitimately run more than once per build; the fixup skips paths already rooted at $(PCP_LIB_DIR) so repeat runs are a no-op instead of nesting the prefix), and call it after $(LNMAKE) in each affected GNUmakefile.

## Related Issues
Fixes #

## Checklist
- [ ] ** Description **
_Why_ this PR exists. Reference all relevant information, including _background_, _issues_, _test failures_, etc
- [ ] ** Commmits **
_Commits_ are self contained and only do one thing
_Commits_ have a header of the form: `module: short description`
_Commits_ have a body (whenever relevant) containing a detailed description of the addressed problem and its solution
- [ ] Documentation updated
- [ ] Tests added/updated

### DIFF
--- a/src/include/builddefs.in
+++ b/src/include/builddefs.in
@@ -449,8 +449,27 @@ LINKER_MAKERULE = \
     for lib in `otool -L $@ | ${AWK} '/libpcp/ {print $$1}'`; do \
         install_name_tool -change $$lib $(PCP_LIB_DIR)/$$lib $@; \
     done
+
+# Same fixup as LINKER_MAKERULE above, but for qmake/QTMAKE-built binaries.
+# These are linked directly by qmake's generated Makefile, bypassing the
+# $(PROGTARGET)/$(LIBTARGET) rules in buildrules that invoke
+# LINKER_MAKERULE via $@, so without this the binary keeps a bare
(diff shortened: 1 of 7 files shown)


lukas7811:
Hi guys, on one project, I needed to build PCP on Mac and I learned that there are problems with build in general based on the README and more in the pmdumptext build. Claude analyzed and fixed that for me, so be careful to check the proposed changes. So far I can confirm that it can be started without a crash at least.

PR #2714, opened 2026-09-15
Add TLS/HTTPS encryption to the PCP push model

sfeifer:
The push model ships PCP archive data to pmproxy over HTTP POST (/logger/{label,meta,index,volume}, default port 44322) from two senders - pmlogpush and pmlogger remote-push - through the shared HTTP client in libpcp_web.  That transport was plaintext only; this adds https:// support so the push path can be encrypted end-to-end, reusing PCP's existing OpenSSL machinery and /etc/pcp/tls.conf.

libpcp/libpcp3:
- New __pmSecureClientConnect(fd, hostname): raw TLS setup + handshake only (no PDU/SASL), sharing the SSL-setup block with the existing PDU client path.  Non-OpenSSL stub returns -EOPNOTSUPP.  Exported via exports.in and declared in libpcp.h.

libpcp_web/http_client.c:
- https scheme support: F_SECURE flag, TLS handshake after connect (no silent fallback - https fails cleanly if TLS is unavailable or the handshake fails), scheme-aware keep-alive matching and redirects.
- Send each POST's header and body as a single write.  Over TLS two writes become two records, and pmproxy drains only one record per network read, so a coalesced header+body could leave the body record unread and hang the exchange.  One contiguous write keeps each POST to a single TLS record.

Senders:
- pmlogpush: new --secure/-S option (-S claimed in overrides() so it is not consumed as the standard start-time option); build_conn() emits https:// when set.
- pmlogger: accept an https:// remote-push target.
- logpush.class: documented https:// opt-in.

QA:
- New test qa/2110 exercises secure push, plaintext auto-detect, and clean failure (no cleartext fallback) of --secure against a non-TLS pmproxy.


sfeifer:
@natoscott @kurik Ready for review. There were a few decisions I made when adding this that I wanted to highlight in case you (or others) may have some thoughts on them.

1. Server authentication strength
  - I chose to have the HTTPS push always authenticate the server (peer chain + hostname verification).
  - This is different from current PCP norms as no other PCP TLS client does hostname verification, and client-side chain verification is opt-in (SSL_VERIFY_NONE by default).
  - This means the push model is stricter than the rest of PCP. The pmproxy cert must chain to a CA trusted in tls.conf and match the connect hostname, otherwise --secure fails.
  - It's applied per-connection in the HTTPS helper only, so the PDU-based
    clients (pmcd/pmproxy) are unaffected.

2. Reuse pcp’s existing openssl machinery
- I chose to reuse PCP's existing openssl machinery and tls.conf rather than developing a new TLS setup somewhere in the http client code.

3. New __pmSecureClientConnect
- Created a new __pmSecureClientConnect function to only handle the raw TLS setup + handshake, factoring the shared SSL-setup block (secure_client_setup) out of the PDU path
- The other already developed option, __pmSecureClientHandshake, also drives PCP PDU/SASL choreography that a plain HTTPS server doesn’t speak

4. Fix the multi-volume hang client-side vs. in pmproxy
- I chose having the client write each post as one contiguous __pmSend (header + body = 1 TLS record) since I did not want to make any changes in pmproxy. The root cause of this is the server drains one record per read event, so a split write left the body buffered and unread.

5. Never silently fallback to plaintext
- https:// on a non-openssl build, failed handshake/verification, and redirect that would downgrade https -> http all error out rather than falling back to cleartext

kurik:
> @natoscott @kurik Ready for review. There were a few decisions I made when adding this that I wanted to highlight in case you (or others) may have some thoughts on them.

>    1. Server authentication strength
>         Hostname verification

The code performs hostname validation with `SSL_set1_host()`, which means it will not work with IP addresses (like my home network without DNS :-) ).

I would also prefer if there can be a different mechanism (like "per-pmlogger Bearer token" or "Basic auth" which pmproxy already has implemented) to authenticate. The reason is because to use authentication via certificates limits (or complicates) the use of self-signed certificates, which is IMO what many users use for communication between services on internal networks.

If I understand the implementation correctly, there might also be an issue with the already implemented `--certreqd` option, when `pmproxy` will require the same trusted client certificate for `pmlogger` <-> `pmproxy` communication as well as for `pmproxy` <-> `grafana` communication, which users might need to distinguish.

>     2. Reuse pcp’s existing openssl machinery
> 
>     * I chose to reuse PCP's existing openssl machinery and tls.conf rather than developing a new TLS setup somewhere in the http client code.

I am unsure whether this is a good decision. I.e. in Federated/Centralized setup you might want to use different certificate/key towards `pmcd` and different towards `pmproxy`. This is not possible with the current implementation as `pmlogger` uses the same cert/key with regards the endpoint it talks to.

>     4. Fix the multi-volume hang client-side vs. in pmproxy
> 
>     * I chose having the client write each post as one contiguous __pmSend (header + body = 1 TLS record) since I did not want to make any changes in pmproxy. The root cause of this is the server drains one record per read event, so a split write left the body buffered and unread.

I would prefer to fix the way how `pmproxy` reads the stream instead of implementing a workaround.

